PR #16, opened 2020-10-06
Add files necessary for amazon-freertos

AniruddhaKanhere:
Description
-----------
- Since this repository will be submoduled in aws/amazon-freertos, some files need to be added to allow the projects in amazon-freertos to work.
  These files ideally should be moved out of this repository at a later point in time.

- Also, this PR removes the 'test' directory from the "complexity" check since the 'test' directory is not actual source code.

Test Steps
-----------

Related Issue
-----------


CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+afr_module(INTERNAL)
```
RichardBarry, line 1: Is this just a bulk standard cmakelists.txt file?  If it is in the root directory within the source of truth repo it can't have anything Amazon specific in the file.

AniruddhaKanhere:
Closing this PR since it is not required anymore as the projects in amazon-freertos have been modified.

PR #17, opened 2020-10-08
Create manifest.yml

AniruddhaKanhere:
Description
-----------
This PR adds `manifest.yml` file to the TCP repository.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,7 +1,7 @@
 name : "FreeRTOS-Plus-TCP"
 version: "V2.3.0"
 description: |-
-  "FreeRTOS TCP/IP stack working on top of FreeRTOS-Kernel to implement IP protocol. Suitable for microcontrollers."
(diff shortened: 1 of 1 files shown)


manifest.yml
```diff
@@ -0,0 +1,14 @@
+name : "FreeRTOS-Plus-TCP"
+version: "V2.3.0"
+description: |-
+  "FreeRTOS TCP/IP stack working on top of FreeRTOS-Kernel to implement IP protocol. Suitable for microcontrollers."
```
gitwallit, line 4: 1) small grammar - "the" FreeRTOS-Kernel (2) does it not also implement the TCP protocol?

AniruddhaKanhere: Thanks for your review.
1. I corrected the grammatical mistake. Thanks.
2. It has been called TCP/IP stack in all our documentation (and technically TCP or IP cannot work without one another). Thus, I modified the statement to call it "TCP/IP" instead of "IP" stack.

gitwallit (approved):

cobusve (approved):

gkwicker (approved):

PR #41, opened 2020-10-20
buffer fixes to the NetworkInterface for the LCP54018

n9wxu:
Buffer fixes for the LPC54018 ethernet

Description
-----------
fixed the zerocopy buffer allocation.  Both in alignment and in initializing the first entry to the descriptor pointer.

### DIFF
--- a/FreeRTOS_DNS.c
+++ b/FreeRTOS_DNS.c
@@ -1098,7 +1098,7 @@ for testing purposes, by the module test_freertos_tcp.c
 */
 	uint32_t ulDNSHandlePacket( const NetworkBufferDescriptor_t *pxNetworkBuffer )
 	{
-	DNSMessage_t *pxDNSMessageHeader;
+	uint8_t *pucPayLoadBuffer;
 	size_t uxPayloadSize;
 
 		/* Only proceed if the payload length indicated in the header
@@ -1109,11 +1109,10 @@ for testing purposes, by the module test_freertos_tcp.c
 
 			if( uxPayloadSize >= sizeof( DNSMessage_t ) )
 			{
-				pxDNSMessageHeader =
-					ipCAST_PTR_TO_TYPE_PTR( DNSMessage_t, pxNetworkBuffer->pucEthernetBuffer );
+				pucPayLoadBuffer = &( pxNetworkBuffer->pucEthernetBuffer[ sizeof( UDPPacket_t ) ] );
 
 				/* The parameter pdFALSE indicates that the reply was not expected. */
-				( void ) prvParseDNSReply( ( uint8_t * ) pxDNSMessageHeader,
+				( void ) prvParseDNSReply( pucPayLoadBuffer,
 										   uxPayloadSize,
 										   pdFALSE );
 			}
--- a/FreeRTOS_DNS.c
+++ b/FreeRTOS_DNS.c
@@ -42,8 +42,6 @@
 #include "NetworkBufferManagement.h"
 #include "NetworkInterface.h"
 
-#include "FreeRTOSIPConfigDefaults.h"
-
 /* Exclude the entire file if DNS is not enabled. */
 #if ( ipconfigUSE_DNS != 0 )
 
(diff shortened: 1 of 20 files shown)


lundinc2 (approved):

AniruddhaKanhere (approved):

htibosch:
@lundinc2 , @n9wxu , thank you for the new network driver!

Although the PR is merged and closed, I do have some remarks.

The function `xSendEventStructToIPTask()` returns either pdPASS or pdFAIL. Although these values are the same as pdTRUE and pdFALSE, it is custom to compare the result of `xSendEventStructToIPTask()` with pdPASS or pdFAIL.

More serious: I do not understand why [Joseph](https://github.com/n9wxu) has changed the logic. As it is now: a packet that is transferred to the IP-task will be released, whereas packets that could not be delivered will cause memory/buffer leaks.
Note: the IP-task will release all network buffers that it receives with the `eNetworkRxEvent` message.

So unless I am missing some logic, I think that the code should be like this:
~~~c
    if( eConsiderFrameForProcessing( pxBufferDescriptor->pucEthernetBuffer ) == eProcessBuffer )
    {
        xRxEvent.eEventType = eNetworkRxEvent;
        xRxEvent.pvData = ( void * ) pxBufferDescriptor;

        if( xSendEventStructToIPTask( &xRxEvent, 0 ) == pdFAIL )
        {
            /* Sending the packet failed, release the network buffer. */
            vReleaseNetworkBuffer( pxBufferDescriptor );
            iptraceETHERNET_RX_EVENT_LOST();
        }
        else
        {
            /* Now the IP-task owns the network buffer, and will release it. */
            iptraceNETWORK_INTERFACE_RECEIVE();
        }
    }
~~~

The comment was not correct:

    /* put this back if using bufferallocation_2.c */

`vReleaseNetworkBuffer()` does 2 things:

- Call `vPortFree()` to release `pucEthernetBuffer` ( indeed bufferallocation_2.c only )
- Add the network buffer descriptor to the list of available descriptors ( both bufferallocation_1 and bufferallocation_2.c ).


I added `iptraceNETWORK_INTERFACE_RECEIVE()`, as it is common in all network drivers.

I would reorganise `rx_task()` as follows, using a while in stead of an if:

    while( pdTRUE )
    {
        ulTaskNotifyTake( pdTRUE, portMAX_DELAY );

        while( pdTRUE )
        {
            status = ENET_GetRxFrameSize( EXAMPLE_ENET_BASE, &g_handle, &length, 0 );
            if( ( status != kStatus_Success ) || ( length == 0U ) )
            {
                break;
            }
            pxBufferDescriptor = pxGetNetworkBufferWithDescriptor( length, 0 );

            if( pxBufferDescriptor != NULL )
            {
                /* Send the bufffer to the IP-task. */
                prvSendDescriptor( pxBufferDescriptor );
            }
            else
            {
                /* I think that ENET_ReadFrame() must still be called.
                 * If not, it might keep on triggering task.
                 * Maybe call ENET_GetRxFrameSize() with NULL as a parameter and
                 * change the code to allow for NULL?
                 * The packet must be dropped.
                 */
            }
        } /* while( pdTRUE ) */
        /* Check the status of the PHY regularly. When packet are sent while the
         * PHY status is low, it may hold up the task for a long time, because of
         * time-outs.
         */
    } /* while( pdTRUE ) */

n9wxu:
This is very helpful.  Working on this interface was a learning experience so I am happy to have your feedback.  I will correct and submit a new PR shortly.

n9wxu:
One question concerning vReleaseNetworkBuffer
I cannot find that in BufferAllocation_1
However: vReleaseNetworkBufferAndDescriptor is present in BufferAllocation_1.  Does that the actual function required?

n9wxu:
At the end of the recommended receive loop, you have a comment on checking the PHY and the caution is to prevent TX timeouts.  It seems to be that checking the link state before TX is a better choice.  It looks like I could adjust the notifyTake in the receiver to timeout every so often and poll the link state but I don't see an API to push the link state into the TCP stack.  Instead there is a xGetPhyLinkStatus to allow the TCP stack to pull the state.  I think checking the link before TX is the better choice instead of checking the link state in the receiver and then signaling the transmitter.

htibosch:
> However: vReleaseNetworkBufferAndDescriptor is present in
> BufferAllocation_1. Does that the actual function required?

Oops on my side: I was mistaken between `vReleaseNetworkBuffer()` (BufferAllocation_2 only), and `vReleaseNetworkBufferAndDescriptor()`, which exists in both BufferAllocation_1 and 2. I corrected my sample code here above.

PR #42, opened 2020-10-21
Uncrustify v3

htibosch:
Description
-----------
In an earlier commit, Uncrustify was applied on the sources. However, that version of Uncrustify was probably too old ( <= 0.70 ), and it produced incorrect indentations. Also the uncrustify.cfg would still use tabs.
Now we used [uncrustify.cfg](https://github.com/aws/amazon-freertos/blob/master/tools/uncrustify.cfg) from AWS/FreeRTOS.

This time, Uncrustify-0.71.0_f was applied recursively on all files matching `*.{c,h}`.


test/unit-test/ConfigFiles/portmacro.h
```diff
@@ -1,164 +0,0 @@
...
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
- * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
- * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
- * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- *
- * http://www.FreeRTOS.org
- * http://aws.amazon.com/freertos
- *
- * 1 tab == 4 spaces!
```
AniruddhaKanhere, line 25: This file should not be removed for unit tests to function correctly...

AniruddhaKanhere:
Hello @htibosch, do we need this PR anymore since we already have the other uncrustify PR [#44](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/44) merged.

test/unit-test/ConfigFiles/portmacro.h
htibosch: There seems to be a bug in some versions of Uncrustify: it gets stuck on the above file portmacro.h. After pressing control-C, the output of Uncrustify is empty, and the file disappears.
Version 0.66 / 0.67 doesn't have this bug, so we're safe.

htibosch:
As the uncrustifying has been done in PR #44, this PR can be closed.
Hein

PR #44, opened 2020-10-22
Uncrustify

AniruddhaKanhere:
Uncrustify the sources. [.ch].

Description
-----------

Test Steps
-----------

Related Issue
-----------


htibosch:
That has been a lot of work, thank you for this.

I checked everything by replaying it: I took the current master, and applied uncrustify on about 150 source and header files.
Then I compared the results. Only in about 4 files I saw a difference in the number of newlines, but that is not a blocker to me.

I recommend getting on with this PR!

Hein

PR #48, opened 2020-10-23
removed extra prints

n9wxu:
removed extra print statements in the LPC54018 NetworkInterface

### DIFF
--- a/portable/NetworkInterface/LPC54018/NetworkInterface.c
+++ b/portable/NetworkInterface/LPC54018/NetworkInterface.c
@@ -154,7 +154,6 @@ status_t status;
 	{
 		ulTaskNotifyTake( pdTRUE, portMAX_DELAY );
 
-		PRINTF( "RX Task Notify - " );
 		BaseType_t receiving = pdTRUE;
 
 		while( receiving == pdTRUE )
@@ -174,15 +173,6 @@ status_t status;
 							status = ENET_ReadFrame( ENET, &g_handle, pxBufferDescriptor->pucEthernetBuffer, length, 0 );
 							pxBufferDescriptor->xDataLength = length;
 
-							PRINTF( "Receiving : " );
-
-							for( int x = 0; x < pxBufferDescriptor->xDataLength; x++ )
-							{
-								PRINTF( " %02x", pxBufferDescriptor->pucEthernetBuffer[ x ] );
-							}
-
-							PRINTF( " - " );
-
 							if( eConsiderFrameForProcessing( pxBufferDescriptor->pucEthernetBuffer ) == eProcessBuffer )
 							{
 								xRxEvent.eEventType = eNetworkRxEvent;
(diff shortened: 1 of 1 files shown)


cobusve (approved):

htibosch:
Calling `vTaskNotifyGiveFromISR()` without doing a yield, may not result in an immediate wake-up of the receive task `rx_task()`.
~~~c
+    BaseType_t xSwitchRequired = pdFALSE;
     switch( event )
     {
         case kENET_TxIntEvent:
             break;
         case kENET_RxIntEvent:
             {
                 vTaskNotifyGiveFromISR( receiveTaskHandle, &( xSwitchRequired ) );
+                portEND_SWITCHING_ISR( xSwitchRequired );
             }
             break;
~~~

>     BaseType_t xGetPhyLinkStatus( void )
>     {
>         bool link;
> 
>         PHY_GetLinkStatus( &phyHandle, &link );
>         return link ? pdTRUE : pdFALSE;
>     }

Like I said earlier: fetching the link status is expensive, and it may disturb ongoing transfers.  It would be better just to return a variable here, and keep the variable up-to-date in the `rx_task()` by polling the PHY status while there is no reception.
As long as you receive packets, you can safely assume that the PHY link status is high, and no polling is needed.

About `xNetworkInterfaceInitialise()`: this function will be called until it returns pdPASS.
In your case, it may happen that the initialisation will be done multiple times, until the PHY status is high.

Now your `xNetworkInterfaceInitialise()` always returns `pdTRUE`. If there is no link status yet, that would mean that DHCP will fail.

Please have a look at e.g. ["STM32Fxx\NetworkInterface.c"](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/master/portable/NetworkInterface/STM32Fxx/NetworkInterface.c), in which a variable `xMacInitStatus` remembers if initialisation has taken place, and if it was successful..

~~~c
    BaseType_t xResult = pdFAIL;
    if( xMacInitStatus == eMACInit )
    {
        xTXDescriptorSemaphore = xSemaphoreCreateCounting( ( UBaseType_t ) ETH_TXBUFNB, ( UBaseType_t ) ETH_TXBUFNB );
        if( xTXDescriptorSemaphore == NULL )
        {
            xMacInitStatus = eMACFailed;
        }
        else
        {
            xMacInitStatus = eMACPass;
        }
    }
    if( xMacInitStatus != eMACPass )
    {
        /* EMAC initialisation failed, return pdFAIL. */
        xResult = pdFAIL;
    }
    else
    {
        if( xPhyObject.ulLinkStatusMask != 0uL )
        {
            xResult = pdPASS;
            FreeRTOS_printf( ( "Link Status is high\n" ) ) ;
        }
    }
    return xResult;
~~~

Mind you: the function `xNetworkInterfaceInitialise()` is called from the IP-task, while `rx_task()` is supposed to check the link status.
Make sure that `xNetworkInterfaceInitialise()` doesn't access the MMI bus, because that code is "non-reentrant".

Hein

PR #49, opened 2020-10-26
Link tuning

n9wxu:
link state change updates

Description
-----------
the phy link checks should be done infrequently in a single thread to avoid disrupting the link.  These changes only check the link if there has been no RX activity for 500ms.


cobusve (approved):

portable/NetworkInterface/LPC54018/NetworkInterface.c
```diff
@@ -129,197 +131,205 @@ void ENET_IntCallback( ENET_Type * base,
...
+
+		if(notified == pdFALSE) // no RX packets for a bit so check for a link
+		{
+			PHY_GetLinkStatus( &phyHandle, &g_linkStatus );
+		}
+		else // got a packet notification so process it.
+		{
+			BaseType_t receiving = pdTRUE;
+
+			while( ( receiving == pdTRUE ) && notified )
```
AniruddhaKanhere, line 168: Is the check on 'notified' necessary? If the control reached at the while, notified must be `pdTRUE`.
I did not find any instance of notified being modified inside the loop too.

portable/NetworkInterface/LPC54018/NetworkInterface.c
```diff
@@ -129,197 +131,205 @@ void ENET_IntCallback( ENET_Type * base,
...
+
+	while( pdTRUE )
+	{
+		BaseType_t notified = ulTaskNotifyTake( pdTRUE, pdMS_TO_TICKS(500) );
+
+		if(notified == pdFALSE) // no RX packets for a bit so check for a link
+		{
+			PHY_GetLinkStatus( &phyHandle, &g_linkStatus );
+		}
+		else // got a packet notification so process it.
```
AniruddhaKanhere, line 164: From perspective of complexity: Instead of an 'else' statement here, can we use a 'continue' statement after line 162 and remove this else?

AniruddhaKanhere (changes requested):

portable/NetworkInterface/LPC54018/NetworkInterface.c
AniruddhaKanhere: I tried making the changes locally and the new score for this file is 50 (currently it is in the range of 90).

portable/NetworkInterface/LPC54018/NetworkInterface.c
```diff
@@ -129,197 +131,205 @@ void ENET_IntCallback( ENET_Type * base,
...
-        PRINTF( "Network Receive Task creation failed!.\n" );
-
-        while( 1 )
-        {
-        }
-    }
-
-    return pdTRUE;
+BaseType_t returnValue = pdFALSE;
+static enum {initPhy, startReceiver, waitForLink, configurePhy }networkInitialisePhase = initPhy;
```
htibosch, line 247: The switch is a little confusing because it uses "fall through" without a warning.

I would prefer to see some comment:
~~~c
                }
            }
            /* Fall through. */
        case startReceiver:
            networkInitialisePhase = startReceiver;
~~~

portable/NetworkInterface/LPC54018/NetworkInterface.c
```diff
@@ -129,197 +131,205 @@ void ENET_IntCallback( ENET_Type * base,
...
+			/* Create the handler. */
+			ENET_CreateHandler( ENET, &g_handle, &config, &buffConfig[ 0 ], ENET_IntCallback, NULL );
+			NVIC_SetPriority( 65 - 16, 4 ); /* TODO this is a hack and I would expect a nice ENET API for priority. */
+
+			/* Active TX/RX. */
+			ENET_StartRxTx( ENET, 1, 1 );
+
+			networkInitialisePhase = initPhy;
+			returnValue = pdTRUE;
+		}
```
htibosch, line 330: That is a good solution: a switch that handles all initialisation in stages!

I do understand why you put the status to `initPhy` again? The task has been created, you don't want to do that again.

It is preferred to let `xNetworkInterfaceInitialise()` return pdPASS and pdFAIL, rather then pdTRUE and pdFALSE.

Also, I would call `xTaskCreate()` in the latest stage because the RX-talk will take over the access to the PHY. Now there might be concurrent access.

portable/NetworkInterface/LPC54018/NetworkInterface.c
htibosch: Yes, I noticed the same: `notified` will not change during that loop.

htibosch (commented):
Thanks for these changes again!

portable/NetworkInterface/LPC54018/NetworkInterface.c
n9wxu: Ooops, left in an earlier change. Thanks for noticing.

portable/NetworkInterface/LPC54018/NetworkInterface.c
n9wxu: Resolved complexity to 8 by breaking out the frame processing into a new function.

portable/NetworkInterface/LPC54018/NetworkInterface.c
n9wxu: added comments

portable/NetworkInterface/LPC54018/NetworkInterface.c
n9wxu: Moved xTaskCreate to the end.  Also switched for pdPASS/pdFAIL

The last initPhy was out of concern for reinitialising the network.  I removed it so you cannot reinitialise the network.  If this is needed we will need to think about it.

portable/NetworkInterface/LPC54018/NetworkInterface.c
```diff
@@ -129,197 +131,222 @@ void ENET_IntCallback( ENET_Type * base,
...
+            if( xTaskCreate( rx_task, "rx_task", 512, NULL, ( configMAX_PRIORITIES - 1 ), &receiveTaskHandle ) != pdPASS )
+            {
+                PRINTF( "Network Receive Task creation failed!.\n" );
+                break;
+            }
 
-        while( 1 )
-        {
-        }
+            returnValue = pdPASS;
```
htibosch, line 346: This is the last statement of the case `startReceiver`. Some compiler will expect a break here:
~~~c
     case startReceiver:
         networkInitialisePhase = startReceiver;
         if( xTaskCreate( ) )
         {
             PRINTF( "Network Receive Task creation failed!.\n" );
             break;
         }
         returnValue = pdPASS;
+        break;
 }
 
 return returnValue;
~~~
Otherwise it all looks good to me.
Thanks, Hein

n9wxu: Done

AniruddhaKanhere (dismissed):

cobusve (approved):

htibosch:
The changes look OK to me, but I am wondering why you declare the following global variables?
~~~c
uint8_t multicastAddr[ 6 ] = { 0x01, 0x00, 0x5e, 0x00, 0x01, 0x81 };
uint8_t g_frame[ ENET_EXAMPLE_PACKAGETYPE ][ ENET_EXAMPLE_FRAME_SIZE ];
uint8_t * g_txbuff[ ENET_TXBD_NUM ];
uint32_t g_txIdx = 0;
uint8_t g_txbuffIdx = 0;
uint8_t g_txCosumIdx = 0;
uint32_t g_testIdx = 0;
uint32_t g_rxIndex = 0;
uint32_t g_rxCheckIdx = 0;
~~~
They are not used in NetworkInterface.c, so why declare them here? And aren't their prototypes declared in a header file?

Further, `ipCONSIDER_FRAME_FOR_PROCESSING` is defined without being used?
You are supposed to call `ipCONSIDER_FRAME_FOR_PROCESSING()` in stead of `eConsiderFrameForProcessing()`.

Just curious: what is the value of `ENET_RXBUFF_SIZE`? Is it related to `ipconfigNETWORK_MTU`?

cobusve (approved):

cobusve (approved):

PR #52, opened 2020-10-27
Add #ifs to enable UDP only compilation

AniruddhaKanhere:
Description
-----------
This PR is in response to the issue https://github.com/FreeRTOS/FreeRTOS/issues/314.
FreeRTOS+TCP doesn't compile when UDP only configuration is enabled.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS_TCP_WIN.c
+++ b/FreeRTOS_TCP_WIN.c
@@ -23,6 +23,7 @@
  * http://www.FreeRTOS.org
  */
 
+#if( ipconfigUSE_TCP == 1 )
 /**
  * @file FreeRTOS_TCP_WIN.c
  * @brief Module which handles the TCP windowing schemes for FreeRTOS+TCP.  Many
@@ -2505,3 +2506,5 @@ void vTCPWindowInit( TCPWindow_t * pxWindow,
 
 #endif /* ipconfigUSE_TCP_WIN == 0 */
 /*-----------------------------------------------------------*/
+
+#endif /* ipconfigUSE_TCP == 1 */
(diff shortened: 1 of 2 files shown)


FreeRTOS_TCP_WIN.c
```diff
@@ -23,6 +23,8 @@
  * http://www.FreeRTOS.org
  */
 
+#if ( ipconfigUSE_TCP == 1 )
```
htibosch, line 26:
~~~c
 *
 * http://aws.amazon.com/freertos
 * http://www.FreeRTOS.org
 */

#if ( ipconfigUSE_TCP == 1 )
~~~
Oops, this is much too early in the file. Nothing has been included yet and `ipconfigUSE_TCP` is not yet defined.

I would put the test right after these includes:
~~~c
 /* FreeRTOS+TCP includes. */
 #include "FreeRTOS_UDP_IP.h"
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_Sockets.h"

-#include "FreeRTOSIPConfigDefaults.h"
+#if ( ipconfigUSE_TCP == 1 )

~~~

and please remove `#include "FreeRTOSIPConfigDefaults.h"`, because that file has already been included in the headers files. it is confusing to repeat it here.

You may try this by replacing `#include` with the following:
~~~c
#ifndef FREERTOS_DEFAULT_IP_CONFIG_H
    #error Uh?
#endif
~~~

AniruddhaKanhere: Ah! I missed that completely.
Done, thank you @htibosch.

AniruddhaKanhere: I had to also uncrustify the files which now show a lot of changes due to that.

aggarg (approved):

htibosch (approved):
That looks good now, thank you.
Hein

PR #56, opened 2020-10-29
Add build combination checks

AniruddhaKanhere:
Description
-----------
This PR adds build combination checks to the FreeRTOS-Plus-TCP repository to check the 'buildability' of the source code when various options are switched on or off in FreeRTOSIPConfig.h.
The checks included in this PR are just of two kinds: One which enables everything and one which disables everything.

Test Steps
-----------
- Go to the `test/build-combination` directory.
- Run `cmake -S . -B build -DTEST_CONFIGURATION=DISABLE_ALL` **OR**
                 `cmake -S . -B build -DTEST_CONFIGURATION=ENABLE_ALL`.
- Go to build directory: `cd build`.
- Run `make`.

The make command should finish successfully.

Related Issue
-----------


.github/workflows/ci.yml
```diff
@@ -104,6 +104,23 @@ jobs:
           doxygen docs/doxygen/config.doxyfile 2>&1 | tee doxyoutput.txt
           if [[ "$(wc -c < doxyoutput.txt | bc)" = "0" ]]; then exit 0; else exit 1; fi
 
+  build-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Update submodules
+        run: git submodule update --init --checkout
```
yanjos-dev, line 112: Should this have the --recursive flag in case there are nested submodules?

AniruddhaKanhere: This test just uses the FreeRTOS-Kernel sources which does not have any submodules itself.
It would be unnecessary to copy files other than the FreeRTOS-Kernel.

I would not mind adding that flag though. What is your opinion @yanjos-dev?

Thanks

yanjos-dev: If the flag isn't required for now then we can go ahead and not use it for now.

yanjos-dev (approved):

gkwicker (approved):

PR #60, opened 2020-11-03
Add CBMC proof for vSocketWakeUpUser

AniruddhaKanhere:
Description
-----------
This PR adds the CBM Cproof to prove the memory safety of the function vSocketWakeUpUser (present in FreeRTOS_Sockets.c).


markrtuttle (changes requested):
This proof omits xEventGroupSetBits from the list of exercised functions (no function body).  Is that intended?  If so, please document it in the proof harness or makefile.

In your abstraction of xQueueGenericSend, you might want to at least check (assert) that the function arguments are nonnull, even better assert that they satisfy whatever assumptions are made in the proof of xQueueGenericSend.  The only hole in this proof is that if xQueueGenericSend has side effects that the rest of this code depends on, those side effects are not being modeled.

AniruddhaKanhere:
Hey @markrtuttle,

> This proof omits xEventGroupSetBits from the list of exercised functions (no function body). Is that intended? If so, please document it in the proof harness or makefile.

I missed that. I have added the stub for the missing function and added checks to make it similar to the original version. Thank you for catching that.

> In your abstraction of xQueueGenericSend, you might want to at least check (assert) that the function arguments are nonnull, even better assert that they satisfy whatever assumptions are made in the proof of xQueueGenericSend. The only hole in this proof is that if xQueueGenericSend has side effects that the rest of this code depends on, those side effects are not being modeled.

Thanks again, I have added the asserts in the stub. I am not sure what do you mean by "side effects". How would I be able to model them?

- Aniruddha

markrtuttle (approved):

cobusve (approved):

PR #63, opened 2020-11-09
Make run script return zero on proof failure

karkhaz:
Prior to this commit, litani would return a non-zero code when any proof
job failed, and the run script would throw an exception when this
happened.

This commit adds the --exit-zero-on-fail flag to the litani run-build
invocation, so that litani returns zero even if there are proof failures
(but will continue to return a non-zero exit code if there was a crash
or some other error).

This is so that CI can distinguish between proof failures (which can be
detected in the litani JSON output), and litani crashes due to bugs in
litani, inconsistent builds, or other errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: ["**"]
   pull_request:
-    branches: [master, feature/ipv6_multi_beta]
+    branches: ["**"]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: ["**"]
   pull_request:
-    branches: [master]
+    branches: [master, feature/ipv6_multi_beta]
   workflow_dispatch:
 
 jobs:
--- a/FreeRTOS_TCP_IP.c
+++ b/FreeRTOS_TCP_IP.c
@@ -3194,7 +3194,7 @@
 
         /* Keep track of the highest sequence number that might be expected within
          * this connection. */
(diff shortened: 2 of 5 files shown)


karkhaz:
@htibosch this PR should ensure that the proof reports actually get posted even in the event of a proof failure. Cheers!

htibosch:
> 
> 
> @htibosch this PR should ensure that the proof reports actually get posted even in the event of a proof failure. Cheers!

I have not studied the CBMC scripts, but your change sounds logical to me. Thanks, Hein

test/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -149,7 +149,7 @@ def run_cmd(cmd, **args):
 
 
 def run_build(litani, jobs):
-    cmd = [str(litani), "run-build"]
+    cmd = [str(litani), "run-build", "--exit-zero-on-fail"]
```
markrtuttle, line 152: Is this a standard option on litani?  Does this mean we have to update every repository based on litani to include this flag in the run script?  Would it not be cleaner to make this is default on litani?

markrtuttle (changes requested):

test/cbmc/proofs/run-cbmc-proofs.py
karkhaz: > Is this a standard option on litani? 

It is. It seems that I wrote this flag for this exact reason, but didn't get a chance to add it to these run scripts early on.

> Does this mean we have to update every repository based on litani to include this flag in the run script?

Yes

> Would it not be cleaner to make this is default on litani?

This would require a PR to every litani-based repository anyway. I also don't think it's cleaner. Returning zero on failure isn't an intuitive or expected behaviour IMHO.

AniruddhaKanhere (approved):

markrtuttle (approved):

n9wxu (approved):

PR #97, opened 2020-11-19
Run CBMC property checking with kissat if available.

markrtuttle:
This pull request uses the kissat solver (if available) to accelerate CBMC property checking.  (It does not yet accelerate coverage checking.)

Two subsequent commits to this pull request also do some cleanup:  Run CBMC with the --flush flag to flush stdout and remove the use of the phony-clean touch file when building goto binaries.

### DIFF
--- a/test/cbmc/proofs/make_common_makefile.py
+++ b/test/cbmc/proofs/make_common_makefile.py
@@ -211,6 +211,15 @@ cbmc-batch.yaml:
     if opsys != "windows":
         _makefile.write(target)
 
+def write_solver(opsys, _makefile):
+    conditional = """
+ifneq ($(strip $(EXTERNAL_SAT_SOLVER)),)
+   SOLVER = --external-sat-solver $(EXTERNAL_SAT_SOLVER)
+endif
+"""
+    if opsys != "windows":
+        _makefile.write(conditional)
+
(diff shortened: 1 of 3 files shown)


test/cbmc/proofs/make_common_makefile.py
```diff
@@ -211,13 +211,23 @@ def write_cbmcbatchyaml_target(opsys, _makefile):
     if opsys != "windows":
         _makefile.write(target)
 
+def write_solver(opsys, _makefile):
+    conditional = """
+ifneq ($(strip $(EXTERNAL_SAT_SOLVER)),)
+   SOLVER ?= --external-sat-solver $(EXTERNAL_SAT_SOLVER)
```
karkhaz, line 217: Just as a matter of style: `SOLVER` is never meant to be set by the user (they're supposed to set `EXTERNAL_SAT_SOLVER`). So it's better to define the variable with `=` or `:=` as a signal that the Makefile will ignore any previously-set value.

karkhaz (approved):

test/cbmc/proofs/make_common_makefile.py
markrtuttle: Done.

karkhaz (approved):

AniruddhaKanhere (approved):

PR #101, opened 2020-11-29
Let 4 network interfaces use the function vPrintResourceStats()

htibosch:
Description
-----------
When developing an embedded application using TCP/IP, it can be very useful to get insight into the consumptions of resources.  Think of the number of available network buffers, the available heap, or the minimum space in the queue of the IP-task.

A while ago, I added a function `vPrintResourceStats()` to [FreeRTOS_IP.c](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/FreeRTOS_IP.c#L3041).

This PR changes 4 network interfaces to let them call `vPrintResourceStats()`.

Test Steps
-----------
Run a project from an of the platforms ESP32, Renesas' RX, STM32Fxx, or SAM, and look at the logging produced.
Note that `ipconfigHAS_PRINTF` must be defined as `1`, and `FreeRTOS_printf()` must be functional. 

### DIFF
--- a/portable/NetworkInterface/DriverSAM/NetworkInterface.c
+++ b/portable/NetworkInterface/DriverSAM/NetworkInterface.c
@@ -46,7 +46,7 @@
 /* gmac_SAM.[ch] is a combination of the gmac.[ch] for both SAM4E and SAME70. */
 #include "gmac_SAM.h"
 #include <sysclk.h>
-#include "phyhandling.h"
+#include "phyHandling.h"
 
 /* This file is included to see if 'CONF_BOARD_ENABLE_CACHE' is defined. */
 #include "conf_board.h"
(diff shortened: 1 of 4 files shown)


portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -46,7 +46,7 @@
 /* gmac_SAM.[ch] is a combination of the gmac.[ch] for both SAM4E and SAME70. */
 #include "gmac_SAM.h"
 #include <sysclk.h>
-#include "phyhandling.h"
+#include "phyHandling.h"
```
AniruddhaKanhere, line 49: Was this intentional?

htibosch: Yes intentional. The header file has always been called `phyHandling.h`.

aggarg (approved):

PR #102, opened 2020-12-01
Call closesocket() for a child-socket in case of a TCP protocol error

htibosch:
Call closesocket() for a child-socket in case of a TCP protocol error during the SYN phase.

Description
-----------
When an application creates a socket, the application is responsible for it deletion.  When done, it must call `FreeRTOS_closesocket()`.  This is true for sockets created with `FreeRTOS_socket()`, and child-sockets obtained by calling `FreeRTOS_accept()`.  Those two functions hand over the ownership of a socket.

However, a child socket in the TCP states between the first SYN and the second ACK ( SYN, SYN+ACK, ACK ), the socket is owned by the IP-stack.  This is true for any OS and for any IP-stack.
In FreeRTOS+TCP, these sockets have a **limited life-time** in this SYN-stage, by default 30 seconds (`ipconfigTCP_HANG_PROTECTION_TIME`).  This is needed in case the negotiation doesn't come to a good end. After a time-out, the socket will be deleted by the IP-task.

At one point in the SYN-stage, in case invalid flags are received, the socket will be give a new state `eCLOSE_WAIT`.  For an outgoing connection, this is OK.  The application will see a socket error, and it will close the socket.  This PR changes that for child-sockets to instead call `FreeRTOS_closesocket()` as it is the IP-task that owns the socket.

### DIFF
--- a/FreeRTOS_TCP_IP.c
+++ b/FreeRTOS_TCP_IP.c
@@ -499,18 +499,10 @@
                      * gets connected. */
                     if( pxSocket->u.xTCP.bits.bPassQueued != pdFALSE_UNSIGNED )
                     {
-                        if( pxSocket->u.xTCP.bits.bReuseSocket == pdFALSE_UNSIGNED )
-                        {
-                            /* As it did not get connected, and the user can never
-                             * accept() it anymore, it will be deleted now.  Called from
-                             * the IP-task, so it's safe to call the internal Close
-                             * function: vSocketClose(). */
-                            ( void ) vSocketClose( pxSocket );
-                        }
-
-                        /* Return a negative value to tell to inform the caller
-                         * xTCPTimerCheck()
-                         * that the socket got closed and may not be accessed anymore. */
+                        /* vTCPStateChange() has called FreeRTOS_closesocket()
+                         * in case the socket is not yet owned by the application.
+                         * Return a negative value to inform the caller that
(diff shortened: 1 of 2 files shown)


FreeRTOS_TCP_IP.c
```diff
@@ -2689,25 +2682,27 @@
...
-
         if( ( ucTCPFlags & ucFlagsMask ) != ucExpect )
         {
             /* eSYN_RECEIVED: flags 0010 expected, not 0002. */
             /* eSYN_RECEIVED: flags ACK  expected, not SYN. */
             FreeRTOS_debug_printf( ( "%s: flags %04X expected, not %04X\n",
                                      ( pxSocket->u.xTCP.ucTCPState == ( uint8_t ) eSYN_RECEIVED ) ? "eSYN_RECEIVED" : "eCONNECT_SYN",
                                      ucExpect, ucTCPFlags ) );
+
+            /* In case pxSocket is not yet onwed by the application, a closure
```
gkwicker, line 2704: spelling: __owned__

gkwicker (changes requested):
One misspelling, see comment

AniruddhaKanhere (dismissed):

AniruddhaKanhere (approved):

gkwicker (approved):

PR #106, opened 2020-12-05
Repair FreeRTOS_SendPingRequest

htibosch:
Description
-----------
`FreeRTOS_SendPingRequest()` first allocates a network buffer and then it does the sanity checks like length and the number of available network buffers. Under some circumstances, this could lead to the loss of network buffers.
The order should be the other way around: check if the ICMP/ping packet may be sent, and then allocate the buffer.

Test Steps
-----------
For instance, call `FreeRTOS_SendPingRequest()` with `uxNumberOfBytesToSend` set to zero.

### DIFF
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -1,5 +1,5 @@
 /*
- * FreeRTOS+TCP V2.3.1
+ * FreeRTOS+TCP V2.3.2
  * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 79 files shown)


FreeRTOS_IP.c
```diff
@@ -1346,22 +1346,22 @@ void FreeRTOS_SetAddressConfiguration( const uint32_t * pulIPAddress,
...
 
-            if( uxNumberOfBytesToSend < ( ipconfigNETWORK_MTU - ( sizeof( IPHeader_t ) + sizeof( ICMPHeader_t ) ) ) )
-            {
-                xEnoughSpace = pdTRUE;
-            }
-            else
-            {
-                xEnoughSpace = pdFALSE;
-            }
+        if( ( uxGetNumberOfFreeNetworkBuffers() >= 4U ) && ( uxNumberOfBytesToSend >= 1U ) && ( xEnoughSpace != pdFALSE ) )
```
AniruddhaKanhere, line 1360: I am curious, why was this (3 or 4) count of free network buffers chosen? And not 1 or 2?

htibosch: The number was 3, which was right after allocating a buffer. Now we check the number of buffer before allocating one.
The number was 3 because the IP-task needs at most 3 network buffers to handle a request.

AniruddhaKanhere: Thanks for the explanation @htibosch.

AniruddhaKanhere (approved):

cobusve (approved):

PR #107, opened 2020-12-08
DHCP network buffers

htibosch:
Description
-----------
In FreeRTOS_DHCP.c, a network buffer is obtained in a blocking way. This is no problem because during the DHCP negociation, the device will drop almost all incoming packets. But if it runs out of network buffers, the DHCP process could stall, thus stalling the IP-task.

In this PR, `pxGetNetworkBufferWithDescriptor()` will be called in a polling way ( with a time-out of zero ). When it fails, a new attempt will be done after 250 msec, a time-out defined by this macro:
~~~c
#define dhcpINITIAL_TIMER_PERIOD    ( pdMS_TO_TICKS( 250U ) )
~~~

Test Steps
-----------
I tested it by simulating a shortage of packages, using the following source of randomness:
~~~
static BaseType_t prvDoCreateMessage( void )
{
    uint32_t ulValue;

    xApplicationGetRandomNumber( &( ulValue ) );
    /* Put in the range 0..99 */
    ulValue = ulValue % 100U;
    if (EP_DHCPData.eDHCPState == eSendDHCPRequest)
    {
        return ulValue < 25;
    }
    else
    {
        return ulValue < 5;
    }
}
~~~

and have `pxGetNetworkBufferWithDescriptor()` fail if `prvDoCreateMessage()` returns a zero.

### DIFF
--- a/FreeRTOS_DHCP.c
+++ b/FreeRTOS_DHCP.c
@@ -1082,7 +1082,7 @@
         #endif /* if ( ipconfigDHCP_REGISTER_HOSTNAME == 1 ) */
 
         /* Obtain a network buffer with the required amount of storage.  It doesn't make much sense
-         * to use a time-out here, because that would cause the IP-task waiting for itself. */
+         * to use a time-out here, because that would cause the IP-task to wait for itself. */
         pxNetworkBuffer = pxGetNetworkBufferWithDescriptor( sizeof( UDPPacket_t ) + uxRequiredBufferSize, 0U );
 
         if( pxNetworkBuffer != NULL )
--- a/FreeRTOS_DHCP.c
+++ b/FreeRTOS_DHCP.c
@@ -390,7 +390,7 @@
                     }
                     else
                     {
-						/* Either the creation of a message buffer failed, or sendto().
+                        /* Either the creation of a message buffer failed, or sendto().
                          * Try again in the next cycle. */
                         FreeRTOS_debug_printf( ( "Send failed during eSendDHCPRequest.\n" ) );
                     }
@@ -421,7 +421,7 @@
                             }
                             else
                             {
-								/* Either the creation of a message buffer failed, or sendto().
+                                /* Either the creation of a message buffer failed, or sendto().
                                  * Try again in the next cycle. */
                                 FreeRTOS_debug_printf( ( "Send failed during eWaitingOffer/1.\n" ) );
                                 EP_DHCPData.eDHCPState = eSendDHCPRequest;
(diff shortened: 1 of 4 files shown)


FreeRTOS_DHCP.c
```diff
@@ -570,10 +627,27 @@
 
                         if( xDHCPSocket != NULL )
                         {
+                            uint32_t ulID;
+
+                            if( xApplicationGetRandomNumber( &( ulID ) ) != pdFALSE )
+                            {
+                                EP_DHCPData.ulTransactionId = ulID;
```
AniruddhaKanhere, line 634: What will be the transaction ID in case `xApplicationGetRandomNumber` fails? We will use the old transaction ID?

FreeRTOS_DHCP.c
```diff
@@ -1007,84 +1081,84 @@
...
-        ( void ) memset( pxDHCPMessage, 0x00, sizeof( DHCPMessage_IPv4_t ) );
-
-        /* Create the message. */
-        pxDHCPMessage->ucOpcode = ( uint8_t ) xOpcode;
-        pxDHCPMessage->ucAddressType = ( uint8_t ) dhcpADDRESS_TYPE_ETHERNET;
-        pxDHCPMessage->ucAddressLength = ( uint8_t ) dhcpETHERNET_ADDRESS_LENGTH;
-        pxDHCPMessage->ulTransactionID = FreeRTOS_htonl( EP_DHCPData.ulTransactionId );
-        pxDHCPMessage->ulDHCPCookie = ( uint32_t ) dhcpCOOKIE;
+        /* Obtain a network buffer with the required amount of storage.  It doesn't make much sense
+         * to use a time-out here, because that would cause the IP-task waiting for itself. */
```
AniruddhaKanhere, line 1085: Grammar: _that would cause the IP-task to wait for itself_

FreeRTOS_DHCP.c
htibosch: I had doubts here. The lease is expired. In the earlier versions, we would continue with the same transaction ID. But is is better to use a new ID. And what if the randomiser is broken?
It is very unlikely that it a randomiser breaks, once it works it will continue to work.

> We will use the old transaction ID?

So yes, extend the lease contract using the same transaction ID.

FreeRTOS_DHCP.c
htibosch: Changed, thanks.

AniruddhaKanhere (approved):

FreeRTOS_DHCP.c
AniruddhaKanhere: Makes sense. Thanks, @htibosch.

FreeRTOS_DHCP.c
AniruddhaKanhere: Thank you.

cobusve (approved):

PR #136, opened 2020-12-23
After sending a TCP packet, swap the source and target IP address ( I…

htibosch:
Description
-----------
When TCP sends out a packet in `prvTCPReturnPacket()`, it temporarily swaps the IPv6 addresses source and target. When the packet is part of the socket space, it will be reused later on. In that case, it is important to restore the value of `pxIPHeader_IPv6->xSourceAddress`

Related Issue
-----------
In some cases, a TCP socket would send a packet to its own IPv6-address.


AniruddhaKanhere (dismissed):

cobusve (dismissed):

FreeRTOS_TCP_IP.c
```diff
@@ -1153,9 +1150,12 @@
                 #if ( ipconfigUSE_IPv6 != 0 )
                     if( ( pxSocket != NULL ) && ( pxSocket->bits.bIsIPv6 != pdFALSE_UNSIGNED ) )
```
gkwicker, line 1151: The v4 logic on line 1158 verifies that `pxIPHeader != NULL` prior to dereferencing it at line 1160.  Should we do the same at line 1151, e.g.:

```c
if( ( pxSocket != NULL ) && 
    ( pxSocket->bits.bIsIPv6 != pdFALSE_UNSIGNED ) && 
    ( pxIPHeader_IPv6 != NULL ) )
```

gkwicker (commented):
Just one question, see my comment.

FreeRTOS_TCP_IP.c
htibosch: Thanks Gary, I added the extra test for `pxIPHeader_IPv6` not being NULL.

AniruddhaKanhere (approved):

cobusve (approved):

PR #154, opened 2021-01-14
Fix FreeRTOS_ClearARP CBMC proof for IPv6

AniruddhaKanhere:
Description
-----------
Fix FreeRTOS_ClearARP CBMC proof for IPv6.
Make the proof to be compatible with IPv6 source code. Add in some more checks/asserts etc.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/test/cbmc/proofs/ARP/ARP_FreeRTOS_ClearARP/ClearARP_harness.c
+++ b/test/cbmc/proofs/ARP/ARP_FreeRTOS_ClearARP/ClearARP_harness.c
@@ -13,7 +13,7 @@
 void harness()
 {
     /* Randomly assign a pointer to be sent to the function. */
-    struct xNetworkEndPoint *pxEndPoint = safeMalloc( sizeof( struct xNetworkEndPoint ) );
+    struct xNetworkEndPoint * pxEndPoint = safeMalloc( sizeof( struct xNetworkEndPoint ) );
 
     FreeRTOS_ClearARP( pxEndPoint );
 }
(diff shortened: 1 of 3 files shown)


test/cbmc/proofs/ARP/ARP_FreeRTOS_ClearARP/ClearARP_harness.c
```diff
@@ -1,12 +1,19 @@
...
 /* FreeRTOS+TCP includes. */
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_ARP.h"
+#include "FreeRTOS_Routing.h"
 
 void harness()
 {
-    FreeRTOS_ClearARP();
+    /* Randomly assign a pointer to be sent to the function. */
+    struct xNetworkEndPoint * pxEndPoint = safeMalloc( sizeof( struct xNetworkEndPoint ) );
```
markrtuttle, line 16: Might be more elegant to write safeMalloc(sizeof(*pxEndPoint))

Consider whether you want to consider using safeMalloc or not.  I believe the proofs are now being run with the cbmc flags that emulate malloc possibly returning null.  You might just want to use malloc here.

This not really doing random assignment.  It is merely returning a pointer to a structure in an unknown location.

markrtuttle (changes requested):

test/cbmc/proofs/ARP/ARP_FreeRTOS_ClearARP/ClearARP_harness.c
AniruddhaKanhere: Thanks for the review Mark!
The flag is not used in +TCP since we were concerned about the increasing time it would take to run on each commit/PR.

Also, I should make the comment much more clearer. By randomness, I meant that it can either be a valid location or NULL. I shall fix this.

markrtuttle (approved):

aggarg (approved):

PR #156, opened 2021-01-14
Fix the proof for FreeRTOS_OutputARPRequest in IPv6 branch

AniruddhaKanhere:
Description
-----------
Fix the proof for FreeRTOS_OutputARPRequest in IPv6 branch.
Make the proof to be compatible with IPv6 source code. Add in some more checks/asserts etc.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/test/cbmc/proofs/ARP/ARP_FreeRTOS_OutputARPRequest/OutputARPRequest_harness.c
+++ b/test/cbmc/proofs/ARP/ARP_FreeRTOS_OutputARPRequest/OutputARPRequest_harness.c
@@ -34,7 +34,7 @@
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_IP_Private.h"
 #include "FreeRTOS_ARP.h"
-
+#include "FreeRTOS_Routing.h"
 
 ARPPacket_t xARPPacket;
 NetworkBufferDescriptor_t xNetworkBuffer;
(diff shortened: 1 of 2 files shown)


test/cbmc/proofs/ARP/ARP_FreeRTOS_OutputARPRequest/OutputARPRequest_harness.c
```diff
@@ -73,9 +73,61 @@ NetworkBufferDescriptor_t * pxGetNetworkBufferWithDescriptor( size_t xRequestedS
...
+                                         NetworkBufferDescriptor_t * const pxNetworkBuffer,
+                                         BaseType_t bReleaseAfterSend )
+{
+    BaseType_t xReturn;
+
+    /* Assert that these parameters are non NULL. */
+    __CPROVER_assert( pxInterface != NULL, "pxInterface cannot be NULL here" );
+    __CPROVER_assert( pxNetworkBuffer != NULL, "The network buffer cannot be NULL" );
+
+    /* Return a random value. */
```
markrtuttle, line 88: Prefer the word arbitrary or unconstrained to random in this situation.  There is no probability space involved.

markrtuttle (approved):

markrtuttle (approved):

aggarg (approved):

PR #158, opened 2021-01-14
Fix Proof for ARPGetCacheEntry

AniruddhaKanhere:
Description
-----------

Fix Proof for ARPGetCacheEntry

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/test/cbmc/proofs/ARP/ARPGetCacheEntry/ARPGetCacheEntry_harness.c
+++ b/test/cbmc/proofs/ARP/ARPGetCacheEntry/ARPGetCacheEntry_harness.c
@@ -6,12 +6,57 @@
 #include "FreeRTOS_IP.h"
 #include "FreeRTOS_IP_Private.h"
 #include "FreeRTOS_ARP.h"
+#include "FreeRTOS_Routing.h"
 
+/* Stub: checks whether the IP address is Multicast or not. This stub
+ * will arbitrarily return a pdTRUE or pdFALSE. */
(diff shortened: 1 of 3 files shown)


test/cbmc/proofs/ARP/ARPGetCacheEntry/ARPGetCacheEntry_harness.c
```diff
@@ -6,12 +6,57 @@
...
+        FreeRTOS_FillEndPoint( &xNetworkInterface1,
+                               &xEndPoint1,
+                               ucIPAddress2,
+                               ucNetMask2,
+                               ucGatewayAddress2,
+                               ucDNSServerAddress2,
+                               ucMACAddress );
+    }
+
+    eARPGetCacheEntry( &ulIPAddress, &xMACAddress, &pxLocalEndPointPointer );
```
markrtuttle, line 61: If you really want these pointers to be nonnull, I would document this with a comment so people know this not a mistake.

markrtuttle (dismissed):

test/cbmc/proofs/ARP/ARPGetCacheEntry/ARPGetCacheEntry_harness.c
AniruddhaKanhere: I see. I will do that. Thanks.

divekarshubham (approved):

aggarg (approved):

markrtuttle (approved):

PR #159, opened 2021-01-15
Fix CBMC Proof for ARPGetCacheEntryByMAC

AniruddhaKanhere:
Description
-----------
Fix Proof for ARPGetCacheEntryByMAC

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
+++ b/test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
@@ -46,5 +46,7 @@ void harness()
                                ucMACAddress );
     }
 
+    /* The pointers to the MAC-Address, IP-Address and the End point cannot be NULL.
+     * Therefore, addresses to actual local variables are passed to the function. */
     eARPGetCacheEntryByMac( &xMACAddress, &ulIPAddress, &pxLocalEndPointPointer );
 }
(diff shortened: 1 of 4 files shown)


test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
```diff
@@ -6,11 +6,45 @@
...
+                               &xEndPoint1,
+                               ucIPAddress2,
+                               ucNetMask2,
+                               ucGatewayAddress2,
+                               ucDNSServerAddress2,
+                               ucMACAddress );
+    }
 
-    eARPGetCacheEntryByMac( &xMACAddress, &ulIPAddress );
+    eARPGetCacheEntryByMac( &xMACAddress, &ulIPAddress, &pxLocalEndPointPointer );
```
markrtuttle, line 49: As before, if you really want these to be nonnull, you should document this so no one thinks you just made a silly mistake.

markrtuttle (approved):

test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
AniruddhaKanhere: Indeed! I fixed that
Thanks Mark!

markrtuttle (approved):

n9wxu (approved):

PR #172, opened 2021-01-25
Fix the CBMC proof of DNSgetHostByName_a

AniruddhaKanhere:
Description
-----------

Fix CBMC proof of DNSgetHostByName_a function.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS_DNS.c
+++ b/FreeRTOS_DNS.c
@@ -377,7 +377,7 @@
     }
     #include "pack_struct_end.h"
     typedef struct xDNSAnswerRecord DNSAnswerRecord_t;
-/** @brief this is for debugging only. */
+/** @brief Used for additional error checking when asserts are enabled. */
     _static struct freertos_addrinfo * pxLastInfo = NULL;
 
 
--- a/FreeRTOS_DNS.c
+++ b/FreeRTOS_DNS.c
@@ -378,7 +378,7 @@
     #include "pack_struct_end.h"
     typedef struct xDNSAnswerRecord DNSAnswerRecord_t;
 /** @brief this is for debugging only. */
-    static struct freertos_addrinfo * pxLastInfo = NULL;
+    _static struct freertos_addrinfo * pxLastInfo = NULL;
 
 
 /**
--- a/test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
+++ b/test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
@@ -14,6 +14,7 @@
 #include "FreeRTOS_UDP_IP.h"
 #include "FreeRTOS_DNS.h"
 #include "FreeRTOS_DHCP.h"
+#include "FreeRTOS_Routing.h"
 #include "NetworkBufferManagement.h"
 #include "NetworkInterface.h"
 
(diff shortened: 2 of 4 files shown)


test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -37,6 +38,76 @@
...
+ */
+NetworkBufferDescriptor_t * pxGetNetworkBufferWithDescriptor( size_t xRequestedSizeBytes,
+                                                              TickType_t xBlockTimeTicks )
+{
+    __CPROVER_assert(
+        xRequestedSizeBytes + ipBUFFER_PADDING < CBMC_MAX_OBJECT_SIZE,
+        "pxGetNetworkBufferWithDescriptor: request too big" );
+
+    /* Arbitrarily return NULL or a valid pointer. */
+    NetworkBufferDescriptor_t * pxReturn = nondet_bool() ? NULL : malloc( sizeof( *pxReturn ) );
```
markrtuttle, line 61: Are you using the malloc can fail flags?  If so, then this conditional expression is not necessary, the flags will do it.

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -37,6 +38,76 @@
...
+
+    if( pxReturn != NULL )
+    {
+        /* Allocate the ethernet buffer. */
+        pxReturn->pucEthernetBuffer = malloc( xRequestedSizeBytes + ipBUFFER_PADDING );
```
markrtuttle, line 66: Again, unclear if you are using the malloc can fail flags.  If not, then malloc will never return fail.

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -37,6 +38,76 @@
...
+             * ahead. This space is used to store data by the TCP stack for its own
+             * use whilst processing the packet. */
+            pxReturn->pucEthernetBuffer += ipBUFFER_PADDING;
+        }
+        else
+        {
+            /* If the ethernet buffer is not allocated, then the network buffer
+             * descriptor is freed and a NULL is returned. */
+            free( pxReturn );
+            pxReturn = NULL;
```
markrtuttle, line 82: Is it enough to nondeterministically choose at line 61 whether you will return NULL, and if not then malloc the buffers and assume the mallocs do not return null.

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -47,8 +118,8 @@
 * function.
 ****************************************************************/
 
-uint32_t prvParseDNSReply( uint8_t * pucUDPPayloadBuffer,
-                           size_t xBufferLength,
+uint32_t __CPROVER_file_local_FreeRTOS_DNS_c_prvParseDNSReply( uint8_t * pucUDPPayloadBuffer,
+                           size_t xBufferLength,struct freertos_addrinfo **ppxAddressInfo,
```
markrtuttle, line 122: Doesn't your style guide require a new line and indention here?

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
```diff
@@ -90,24 +162,48 @@ void func( const char * pcHostName,
...
-
 void harness()
 {
     size_t len;
+    __CPROVER_assume( ( len <= MAX_HOSTNAME_LEN ) && ( len > 0 ) );
 
-    __CPROVER_assume( len <= MAX_HOSTNAME_LEN );
-    char * pcHostName = safeMalloc( len );
+    /* Arbitrarily assign a proper memory or NULL to the hostname. */
+    char * pcHostName = nondet_bool() ? safeMalloc( len ) : NULL;
```
markrtuttle, line 182: I'm pretty sure safeMalloc is already doing this nondeterministic choice of NULL.

markrtuttle (changes requested):

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
AniruddhaKanhere: We are not yet using malloc can fail flags in the TCP repository. Thus, this is necessary.

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
AniruddhaKanhere: Same as above.

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
AniruddhaKanhere: Yes! Nice catch! I will make this change.

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
AniruddhaKanhere: It does... I was going to merge a commit :)

test/cbmc/proofs/DNS/DNSgetHostByName_a/DNSgetHostByName_a_harness.c
AniruddhaKanhere: Nice catch!
I will change that to malloc then just to be inline with the code.

markrtuttle (approved):

FreeRTOS_DNS.c
```diff
@@ -378,7 +378,7 @@
     #include "pack_struct_end.h"
     typedef struct xDNSAnswerRecord DNSAnswerRecord_t;
 /** @brief this is for debugging only. */
-    static struct freertos_addrinfo * pxLastInfo = NULL;
+    _static struct freertos_addrinfo * pxLastInfo = NULL;
```
gkwicker, line 381: If this is only for debugging, then why is it always compiled in?  Shouldn't it be in something like:

```
#ifdef _FREERTOS_DNS_DEBUG_
_static struct freertos_addrinfo * pxLastInfo = NULL; 
#endif 
```

gkwicker (changes requested):
See comment

FreeRTOS_DNS.c
AniruddhaKanhere: Thanks for the review @gkwicker.
Since IPv6 is under development, this is something which is can be used a breakpoint to check the value.
While I agree that the variable should be conditionally compiled, it is later used in below format:
```
configASSERT( pxLastInfo != pxInfo );
        pxLastInfo = pxInfo;

        while( pxInfo != NULL )
        {
            pxNext = pxInfo->ai_next;
            vPortFree( pxInfo );
            pxInfo = pxNext;
        }

        pxLastInfo = NULL;
```

I think that if all instances are conditionally compiled using `#ifdef`, then it would make the code difficult to read and a bit messy. I believe that we can get rid of this (and all other debugging variables/prints) once this library has been thoroughly tested.

Or shall I use the `#ifdef`?

gkwicker: Then the comment at line 380 is incorrect, `pxLastInfo` is _not_ used for debugging only; it is used to implement additional error checking via `configASSERT()`.  The correct thing to do here is to update the comment so that it is accurate, e.g.: _"Used for additional error checking when asserts are enabled."_

AniruddhaKanhere: Done!

gkwicker (approved):

PR #177, opened 2021-01-26
Fix CBMC proof of ProcessDHCPReplies

AniruddhaKanhere:
Description
-----------

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/test/cbmc/proofs/ProcessDHCPReplies/ProcessDHCPReplies_harness.c
+++ b/test/cbmc/proofs/ProcessDHCPReplies/ProcessDHCPReplies_harness.c
@@ -13,13 +13,13 @@
 #include "FreeRTOS_UDP_IP.h"
 #include "FreeRTOS_DHCP.h"
 #include "FreeRTOS_ARP.h"
-
+#include "FreeRTOS_Routing.h"
 
 /****************************************************************
 * Signature of function under test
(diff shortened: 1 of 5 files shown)


test/cbmc/proofs/ProcessDHCPReplies/ProcessDHCPReplies_harness.c
```diff
@@ -32,6 +33,8 @@ void harness()
     /* transaction id is an outgoing message */
 
     BaseType_t xExpectedMessageType;
+    NetworkEndPoint_t xEndPoint;
+    NetworkEndPoint_t * pxEndPoint = &xEndPoint;
```
markrtuttle, line 37: Sure this must be nonnull?

markrtuttle (changes requested):

test/cbmc/proofs/ProcessDHCPReplies/ProcessDHCPReplies_harness.c
AniruddhaKanhere: Yes, it must be Non-NULL. I added some comments to make that clear to the reader.

markrtuttle (approved):

cobusve (approved):

PR #178, opened 2021-01-26
Fix CBMC proof of vSocketBind

AniruddhaKanhere:
Description
-----------

Additionally, the PR also removes different configurations and consolidates them into just one file for easy reading. These configurations will be made into different sets of folders while CBMC proofs are being run.

Test Steps
-----------

Related Issue
-----------


test/cbmc/proofs/Socket/vSocketBind/vSocketBind_harness.c
```diff
@@ -55,8 +57,18 @@ void harness()
     __CPROVER_assume( pxSocket != NULL );
     __CPROVER_assume( pxSocket != FREERTOS_INVALID_SOCKET );
 
+    pxSocket->pxEndPoint = safeMalloc( sizeof( *( pxSocket->pxEndPoint ) ) );
+    __CPROVER_assume( pxSocket->pxEndPoint != NULL );
```
markrtuttle, line 61: Want to get rid of safeMalloc like you have other places?

test/cbmc/proofs/Socket/vSocketBind/vSocketBind_harness.c
```diff
@@ -55,8 +57,18 @@ void harness()
...
+
     struct freertos_sockaddr * pxBindAddress = safeMalloc( sizeof( struct freertos_sockaddr ) );
 
+    #if ( ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND == 0 )
+
+        /* The library asserts that pxBindAddress cannot be
+         * NULL if the above configuration is set. */
+        __CPROVER_assume( pxBindAddress != NULL );
```
markrtuttle, line 69: Maybe try 
```
__CPROVER_assume( IMPLIES (ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND == 0, 
                                                     pxBindAddress != NULL))
```
Does that work?

markrtuttle (changes requested):

test/cbmc/proofs/Socket/vSocketBind/vSocketBind_harness.c
AniruddhaKanhere: Will do.

test/cbmc/proofs/Socket/vSocketBind/vSocketBind_harness.c
AniruddhaKanhere: It does. Thanks.

test/cbmc/proofs/Socket/vSocketBind/vSocketBind_harness.c
markrtuttle: Well, now you have changed the semantics because (unless the cbmc flag is used) malloc will never return NULL.  I was coaching in the direction of using the conditional expression (that implements safemalloc).

markrtuttle (changes requested):

test/cbmc/proofs/Socket/vSocketBind/vSocketBind_harness.c
AniruddhaKanhere: Oh. Something else was on my mind. Thanks, I updated

markrtuttle (approved):

cobusve (approved):

PR #183, opened 2021-01-28
Fix CBMC proof of ProcessReceivedUDPPacket

AniruddhaKanhere:
Description
-----------

Test Steps
-----------

Related Issue
-----------


markrtuttle (changes requested):
Same comment about nondet_bool followed by assume nonnull.

AniruddhaKanhere:
> Same comment about nondet_bool followed by assume nonnull.

I have added that in anticipation of using the --malloc-can-fail flag. But when I just use a malloc, the use of __CPROVER_assume seems weird.
I am thinking to add the flag and clean up all the proofs all at one with an additional PR - since there were 12 or so proofs which did not need modification.
One such simple PR doing just one thing will be extremely easy to review.
What do you think @markrtuttle?

markrtuttle:
> One such simple PR doing just one thing will be extremely easy to review.
> What do you think @markrtuttle?

Famous last words.  I've said the same thing once upon a time.  :-)

markrtuttle (approved):

n9wxu (approved):

PR #187, opened 2021-02-01
IPv6 MISRA checks and changes.

htibosch:
Description
-----------
When checking the labs/ipv6_multi branch with Coverity for MISRA appliance, it couldn't find any violations against mandatory rules.

In this patch I solved a couple of violation against "required" rules, such as 2.2: "There shall be no dead code", code that isn't ever used. I did that by adding the proper `#ifdef`'s.

In the FreeRTOS+TCP library, heavy use was made of "static inline functions". MISRA complains about them. I'm afraid that the quality of the library would become less when they are all removed, both in safety ( type checking ! ), and performance. ( in-lining code that is often used ).
The alternative would be using global functions ( slow ) , the used or function-like macro's ( risky ).

Also there are macro's that were used to allow certain casts from `uint_8*` to a struct pointer. Coverity thinks that these are functions that are never used: dead code.

Beside MISRA changes, I copied a few changes to the logging.

### DIFF
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -477,6 +477,7 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
  *
  * @param[in] pxMACAddress: The MAC-address of the entry of interest.
  * @param[out] pulIPAddress: set to the IP-address found, or unchanged when not found.
+ * @param[out] ppxInterface: Will get a pointer to the interface that holds the MAC-address.
  *
  * @return Either eARPCacheMiss or eARPCacheHit.
  */
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -498,6 +498,12 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
             if( memcmp( pxMACAddress->ucBytes, xARPCache[ x ].xMACAddress.ucBytes, sizeof( MACAddress_t ) ) == 0 )
             {
                 *pulIPAddress = xARPCache[ x ].ulIPAddress;
+
+                if( ppxInterface != NULL )
+                {
+                    *( ppxInterface ) + pxEndPoint->pxNetworkInterface;
+                }
+
                 eReturn = eARPCacheHit;
                 break;
             }
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -140,7 +140,7 @@ eFrameProcessingResult_t eARPProcessPacket( NetworkBufferDescriptor_t * const px
     /* Some extra logging while still testing. */
     if( pxARPHeader->usOperation == ( uint16_t ) ipARP_REQUEST )
     {
-        /*if( ulSenderProtocolAddress != ulTargetProtocolAddress ) */
+        if( ulSenderProtocolAddress != ulTargetProtocolAddress )
         {
             if( pxTargetEndPoint != NULL )
             {
@@ -646,10 +646,14 @@ eARPLookupResult_t eARPGetCacheEntry( uint32_t * pulIPAddress,
             {
                 eReturn = prvCacheLookup( ulAddressToLookup, pxMACAddress, ppxEndPoint );
 
-                FreeRTOS_printf( ( "ARP %lxip %s using %lxip\n",
-                                   FreeRTOS_ntohl( ulOrginal ),
-                                   ( eReturn == eARPCacheHit ) ? "hit" : "miss",
-                                   FreeRTOS_ntohl( ulAddressToLookup ) ) );
+                if( ( eReturn != eARPCacheHit ) || ( ulOrginal != ulAddressToLookup ) )
+                {
+                    FreeRTOS_printf( ( "ARP %lxip %s using %lxip\n",
+                                       FreeRTOS_ntohl( ulOrginal ),
+                                       ( eReturn == eARPCacheHit ) ? "hit" : "miss",
+                                       FreeRTOS_ntohl( ulAddressToLookup ) ) );
+                }
+
                 /* It might be that the ARP has to go to the gateway. */
                 *pulIPAddress = ulAddressToLookup;
             }
@@ -912,6 +916,7 @@ BaseType_t xARPWaitResolution( uint32_t ulIPAddress,
     NetworkEndPoint_t * pxEndPoint;
     size_t uxSendCount = ipconfigMAX_ARP_RETRANSMISSIONS;
 
+    /* The IP-task is not supposed to call this function. */
     configASSERT( xIsCallingFromIPTask() == 0 );
 
     xLookupResult = eARPGetCacheEntry( &( ulIPAddress ), &( xMACAddress ), &( pxEndPoint ) );
@@ -925,7 +930,7 @@ BaseType_t xARPWaitResolution( uint32_t ulIPAddress,
 
         while( uxSendCount > 0 )
         {
-            FreeRTOS_printf( ( "OutputARPRequest %lxip", FreeRTOS_ntohl( ulIPAddress ) ) );
+            FreeRTOS_printf( ( "OutputARPRequest %lxip\n", FreeRTOS_ntohl( ulIPAddress ) ) );
             FreeRTOS_OutputARPRequest( ulIPAddress );
 
             vTaskDelay( uxSleepTime );
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -492,7 +492,7 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
 
         if( ppxInterface != NULL )
         {
-            *( ppxInterface ) + NULL;
+            *( ppxInterface ) = NULL;
         }
 
         /* Loop through each entry in the ARP cache. */
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -490,6 +490,11 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
         configASSERT( pxMACAddress != NULL );
         configASSERT( pulIPAddress != NULL );
 
+        if( ppxInterface != NULL )
+        {
+            *( ppxInterface ) + NULL;
+        }
+
         /* Loop through each entry in the ARP cache. */
         for( x = 0; x < ipconfigARP_CACHE_ENTRIES; x++ )
         {
@@ -501,7 +506,7 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
 
                 if( ppxInterface != NULL )
                 {
-                    *( ppxInterface ) + pxEndPoint->pxNetworkInterface;
+                    *( ppxInterface ) + xARPCache[ x ].pxEndPoint->pxNetworkInterface;
                 }
 
                 eReturn = eARPCacheHit;
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -506,7 +506,7 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
 
                 if( ppxInterface != NULL )
                 {
-                    *( ppxInterface ) + xARPCache[ x ].pxEndPoint->pxNetworkInterface;
+                    *( ppxInterface ) = xARPCache[ x ].pxEndPoint->pxNetworkInterface;
                 }
 
                 eReturn = eARPCacheHit;
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -504,7 +504,8 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
             {
                 *pulIPAddress = xARPCache[ x ].ulIPAddress;
 
-                if( ppxInterface != NULL )
+                if( ( ppxInterface != NULL ) &&
+                    ( xARPCache[ x ].pxEndPoint != NULL ) )
                 {
                     *( ppxInterface ) = xARPCache[ x ].pxEndPoint->pxNetworkInterface;
                 }
--- a/FreeRTOS_ARP.c
+++ b/FreeRTOS_ARP.c
@@ -349,7 +349,7 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
             }
 
             /* Does this line in the cache table hold an entry for the IP
-             * address	being queried? */
+             * address being queried? */
             if( xARPCache[ x ].ulIPAddress == ulIPAddress )
             {
                 if( pxMACAddress == NULL )
@@ -870,8 +870,6 @@ void FreeRTOS_OutputARPRequest( uint32_t ulIPAddress )
                         {
                             BaseType_t xIndex;
 
-                            /*					FreeRTOS_printf( ( "OutputARPRequest: length %lu -> %lu\n", */
-                            /*						pxNetworkBuffer->xDataLength, ipconfigETHERNET_MINIMUM_PACKET_BYTES ) ); */
                             for( xIndex = ( BaseType_t ) pxNetworkBuffer->xDataLength; xIndex < ( BaseType_t ) ipconfigETHERNET_MINIMUM_PACKET_BYTES; xIndex++ )
                             {
                                 pxNetworkBuffer->pucEthernetBuffer[ xIndex ] = 0U;
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -4681,7 +4681,7 @@ uint16_t usChar2u16( const uint8_t * pucPtr )
 #else /* if ( ipconfigUSE_IPv6 != 0 ) */
 
 /**
- * @brief Get the size of an IPv4-header, independant of the network buffer.
+ * @brief Get the size of an IPv4-header, independent of the network buffer.
  * @param[in] pxNetworkBuffer: The network buffer ( ignored ).
  * @return The size of the corresponding IP-header.
  */
@@ -4713,7 +4713,7 @@ uint16_t usChar2u16( const uint8_t * pucPtr )
 #else /* if ( ipconfigUSE_IPv6 != 0 ) */
 
 /**
- * @brief Get the size of an IPv4-header, independant of the socket setting.
+ * @brief Get the size of an IPv4-header, independent of the socket setting.
  * @param[in] pxSocket: The socket ( ignored ).
  * @return The size of an IPv4-header.
  */
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -4713,6 +4713,9 @@ uint16_t usChar2u16( const uint8_t * pucPtr )
  */
     size_t uxIPHeaderSizePacket( const NetworkBufferDescriptor_t * pxNetworkBuffer )
     {
+		/* As this is IPv4-only code, the parameter 'pxNetworkBuffer' is not used. */
+		( void ) pxNetworkBuffer;
+
         return ipSIZE_OF_IPv4_HEADER;
     }
 #endif /* if ( ipconfigUSE_IPv6 != 0 ) */
@@ -4748,6 +4751,9 @@ uint16_t usChar2u16( const uint8_t * pucPtr )
  */
     size_t uxIPHeaderSizeSocket( const FreeRTOS_Socket_t * pxSocket )
     {
+		/* As this is IPv4-only code, the parameter 'pxSocket' is not used. */
+		( void ) pxSocket;
+
         return ipSIZE_OF_IPv4_HEADER;
     }
 #endif /* if ( ipconfigUSE_IPv6 != 0 ) */
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -543,7 +543,7 @@ static void prvIPTask( void * pvParameters )
                     else
                 #endif
                 {
-                    struct freertos_sockaddr * pxAddress = ipPOINTER_CAST( struct freertos_sockaddr *, & ( xAddress ) );
+                    struct freertos_sockaddr * pxAddress = ipPOINTER_CAST( struct freertos_sockaddr *, &( xAddress ) );
 
                     pxAddress->sin_family = FREERTOS_AF_INET;
                     pxAddress->sin_addr = FreeRTOS_htonl( pxSocket->ulLocalAddress );
@@ -553,7 +553,7 @@ static void prvIPTask( void * pvParameters )
                 /* 'ulLocalAddress' and 'usLocalPort' will be set again by vSocketBind(). */
                 pxSocket->ulLocalAddress = 0;
                 pxSocket->usLocalPort = 0;
-                ( void ) vSocketBind( pxSocket, ipPOINTER_CAST( struct freertos_sockaddr *, & ( xAddress ) ), sizeof( xAddress ), pdFALSE );
+                ( void ) vSocketBind( pxSocket, ipPOINTER_CAST( struct freertos_sockaddr *, &( xAddress ) ), sizeof( xAddress ), pdFALSE );
 
                 /* Before 'eSocketBindEvent' was sent it was tested that
                  * ( xEventGroup != NULL ) so it can be used now to wake up the
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -86,8 +86,8 @@
 #endif
 
 /* IPv4 multi-cast addresses range from 224.0.0.0.0 to 240.0.0.0. */
-#define ipFIRST_MULTI_CAST_IPv4             0xE0000000UL /**< Lower bound of the IPv4 multicast address. */
-#define ipLAST_MULTI_CAST_IPv4              0xF0000000UL /**< Higher bound of the IPv4 multicast address. */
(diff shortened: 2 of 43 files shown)


htibosch:
It gets a build error because the enum `ePingReplyStatus_t` is unknown in:
~~~c
void vApplicationPingReplyHook( ePingReplyStatus_t eStatus, uint16_t usIdentifier )
~~~

This can be solved by making the declaration dependent on the macro `ipconfigSUPPORT_OUTGOING_PINGS`. Could you do that Aniruddha?

htibosch:
A summary of what I did since commit "Removed dead code and solved all required MISRA violations" :

The struct `NetworkAddressingParameters_t` was a left-over from the IPv4 release. It is not used in the /multi release.

The structs `IGMPHeader_t` and `PseudoHeader_t` were prepared but not yet in use, thus dropped.

One field name coincided with a struct name: `ICMPHeader_IPv6_t::xIPv6_Address` with `struct xICMPHeader_IPv6`. I changed the name of the field to `xICMPHeaderIPv6`.

The field name `ICMPPacket_IPv6_t::x'ICMPHeader_IPv6` coincided with `struct xICMPHeader_IPv6`. I changed the field name to `xICMPHeaderIPv6`.

Much the same for `ProtocolHeaders_t::xICMPHeader_IPv6` and `struct xICMPHeader_IPv6`, I renamed the field.

The field name `NetworkBufferDescriptor_t::xIPv6_Address` also changed its name to `xIPv6Address`.

These and many other inline function became global function:
~~~c
	uint32_t ulChar2u32( const uint8_t * pucPtr );
	uint16_t usChar2u16( const uint8_t * pucPtr );
	BaseType_t xSocketValid( Socket_t xSocket );
~~~
Also the min/max and round functions became global in stead of `static inline`.

Added dependencies on config macros, for instance here:

    #if ( ipconfigUSE_RA != 0 )
        void vReceiveNA( NetworkBufferDescriptor_t * const pxNetworkBuffer );
    #endif

Removed several functions because they were never used (dead code), for instance 'CONFIRM_EP_v4' and 'CONFIRM_EP_v6'.

Added some macro's to 'FreeRTOSIPConfigDefaults.h' to make sure they are defined when tested.

AniruddhaKanhere:
> It gets a build error because the enum `ePingReplyStatus_t` is unknown in:
> 
> ```c
> void vApplicationPingReplyHook( ePingReplyStatus_t eStatus, uint16_t usIdentifier )
> ```
> 
> This can be solved by making the declaration dependent on the macro `ipconfigSUPPORT_OUTGOING_PINGS`. Could you do that Aniruddha?
> 
> The declaration of `vApplicationPingReplyHook` is in main.c.

Done

FreeRTOS_IP.c
```diff
@@ -3889,53 +3868,157 @@ uint32_t FreeRTOS_GetIPAddress( void )
...
+ * @brief Get the gateway address of the subnet.
+ *
+ * @return The IP-address of the gateway, zero if a gateway is
+ *         not used/defined.
+ */
     uint32_t FreeRTOS_GetGatewayAddress( void )
     {
-        return xNetworkAddressing.ulGatewayAddress;
+        uint32_t ulIPAddress = 0U;
+        NetworkEndPoint_t * pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
```
AniruddhaKanhere, line 3904: Would returning a gateway address based on the endpoint and interface be better rather than returning the first endpoint always?

FreeRTOS_IP.c
```diff
@@ -3889,53 +3868,157 @@ uint32_t FreeRTOS_GetIPAddress( void )
...
+/**
+ * @brief Get the DNS server address.
+ *
+ * @return The IP address of the DNS server.
+ */
     uint32_t FreeRTOS_GetDNSServerAddress( void )
     {
-        return xNetworkAddressing.ulDNSServerAddress;
+        uint32_t ulIPAddress = 0U;
+        NetworkEndPoint_t * pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
```
AniruddhaKanhere, line 3923: Same comment as the one regarding gateway address.

FreeRTOS_IP.c
```diff
@@ -3889,53 +3868,157 @@ uint32_t FreeRTOS_GetIPAddress( void )
...
+/**
+ * @brief Get the netmask for the subnet.
+ *
+ * @return The 32 bit netmask for the subnet.
+ */
     uint32_t FreeRTOS_GetNetmask( void )
     {
-        return xNetworkAddressing.ulNetMask;
+        uint32_t ulIPAddress = 0U;
+        NetworkEndPoint_t * pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
```
AniruddhaKanhere, line 3942: Same comment as that of the gateway.

FreeRTOS_IP.c
```diff
@@ -3889,53 +3868,157 @@ uint32_t FreeRTOS_GetIPAddress( void )
...
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Update the MAC address.
+ *
+ * @param[in] ucMACAddress: the MAC address to be set.
+ */
+    void FreeRTOS_UpdateMACAddress( const uint8_t ucMACAddress[ ipMAC_ADDRESS_LENGTH_BYTES ] )
+    {
+        NetworkEndPoint_t * pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
```
AniruddhaKanhere, line 3960: Same comment as that of the gateway.

FreeRTOS_IP.c
```diff
@@ -3889,53 +3868,157 @@ uint32_t FreeRTOS_GetIPAddress( void )
...
+/**
+ * @brief Get the MAC address.
+ *
+ * @return The pointer to MAC address.
+ */
     const uint8_t * FreeRTOS_GetMACAddress( void )
     {
-        return ipLOCAL_MAC_ADDRESS;
+        const uint8_t * pucReturn = NULL;
+        NetworkEndPoint_t * pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
```
AniruddhaKanhere, line 3978: Same comment as that of the gateway.

FreeRTOS_IP.c
```diff
@@ -3889,53 +3868,157 @@ uint32_t FreeRTOS_GetIPAddress( void )
...
 
+/**
+ * @brief Set the netmask for the subnet.
+ *
+ * @param[in] ulNetmask: The 32 bit netmask of the subnet.
+ */
     void FreeRTOS_SetNetmask( uint32_t ulNetmask )
     {
-        xNetworkAddressing.ulNetMask = ulNetmask;
+        NetworkEndPoint_t * pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
```
AniruddhaKanhere, line 3997: Same comment as that of the gateway.

FreeRTOS_IP.c
```diff
@@ -3889,53 +3868,157 @@ uint32_t FreeRTOS_GetIPAddress( void )
...
 
+/**
+ * @brief Set the gateway address.
+ *
+ * @param[in] ulGatewayAddress: The gateway address.
+ */
     void FreeRTOS_SetGatewayAddress( uint32_t ulGatewayAddress )
     {
-        xNetworkAddressing.ulGatewayAddress = ulGatewayAddress;
+        NetworkEndPoint_t * pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
```
AniruddhaKanhere, line 4013: Same comment as that of the gateway.

FreeRTOS_IP.c
```diff
@@ -4204,6 +4287,471 @@ const char * FreeRTOS_strerror_r( BaseType_t xErrnum,
...
+        return uxResult;
+    }
+#else /* if ( ipconfigUSE_IPv6 != 0 ) */
+
+/**
+ * @brief Get the size of an IPv4-header, independent of the socket setting.
+ * @param[in] pxSocket: The socket ( ignored ).
+ * @return The size of an IPv4-header.
+ */
+    size_t uxIPHeaderSizeSocket( const FreeRTOS_Socket_t * pxSocket )
```
AniruddhaKanhere, line 4749: Why not make this a #define like above?

FreeRTOS_ARP.c
```diff
@@ -490,6 +490,11 @@ void vARPRefreshCacheEntry( const MACAddress_t * pxMACAddress,
         configASSERT( pxMACAddress != NULL );
         configASSERT( pulIPAddress != NULL );
 
+        if( ppxInterface != NULL )
```
AniruddhaKanhere, line 493: Not a comment about this line.
We should add the parameter description of `ppxInterface` in the Doxygen comments on line 483.

FreeRTOS_RA.c
```diff
@@ -109,12 +109,12 @@
         pxICMPPacket->xEthernetHeader.usFrameType = ipIPv6_FRAME_TYPE;
 
         /* Set IP-header. */
-        pxICMPPacket->xIPHeader.ucVersionTrafficClass = 0x60;
-        pxICMPPacket->xIPHeader.ucTrafficClassFlow = 0;
-        pxICMPPacket->xIPHeader.usFlowLabel = 0;
+        pxICMPPacket->xIPHeader.ucVersionTrafficClass = 0x60U;
```
AniruddhaKanhere, line 112: Maybe change this to a MACRO?

FreeRTOS_RA.c
```diff
@@ -109,12 +109,12 @@
...
+        pxICMPPacket->xIPHeader.ucTrafficClassFlow = 0U;
+        pxICMPPacket->xIPHeader.usFlowLabel = 0U;
         pxICMPPacket->xIPHeader.usPayloadLength = FreeRTOS_htons( sizeof( ICMPRouterSolicitation_IPv6_t ) ); /*lint !e845: (Info -- The right argument to operator '|' is certain to be 0. */
         pxICMPPacket->xIPHeader.ucNextHeader = ipPROTOCOL_ICMP_IPv6;
-        pxICMPPacket->xIPHeader.ucHopLimit = 255;
+        pxICMPPacket->xIPHeader.ucHopLimit = 255U;
```
AniruddhaKanhere, line 117: Maybe we should also change this to a macro?

FreeRTOS_RA.c
```diff
@@ -461,9 +461,9 @@
 
                    /* Send a Router Solicitation to ff02::2 */
                    ( void ) memset( xIPAddress.ucBytes, 0, sizeof xIPAddress.ucBytes );
-                   xIPAddress.ucBytes[ 0 ] = 0xff;
-                   xIPAddress.ucBytes[ 1 ] = 0x02;
-                   xIPAddress.ucBytes[ 15 ] = 0x02;
+                   xIPAddress.ucBytes[ 0 ] = 0xffU;
+                   xIPAddress.ucBytes[ 1 ] = 0x02U;
+                   xIPAddress.ucBytes[ 15 ] = 0x02U;
```
AniruddhaKanhere, line 466: I don't think so but still confirming: Are these values worthy of being converted to a macro?

portable/BufferManagement/BufferAllocation_2.c
```diff
@@ -61,15 +59,21 @@
     #define baMINIMAL_BUFFER_SIZE    sizeof( ARPPacket_t )
 #endif /* ipconfigUSE_TCP == 1 */
 
-/*_RB_ This is too complex not to have an explanation. */
 #if defined( ipconfigETHERNET_MINIMUM_PACKET_BYTES )
-    #define ASSERT_CONCAT_( a, b )    a ## b
-    #define ASSERT_CONCAT( a, b )     ASSERT_CONCAT_( a, b )
-    #define STATIC_ASSERT( e ) \
+
+/* Crteating a static assert that can be used with sizeof()
```
AniruddhaKanhere, line 64:
```suggestion
/* Creating a static assert that can be used with sizeof()
```

I wonder why spell-check did not catch this...

AniruddhaKanhere (commented):
@htibosch thank you for this PR.
Please have a look at the comments/questions I have.

FreeRTOS_IP.c
```diff
@@ -1932,7 +1906,7 @@ BaseType_t xSendEventStructToIPTask( const IPStackEvent_t * pxEvent,
  *         succeeded.
  * @param pxEndPoint: The end-point that needs DHCP.
  */
-    BaseType_t xSendDHCPEvent( NetworkEndPoint_t * pxEndPoint )
+    BaseType_t xSendDHCPEvent( struct xNetworkEndPoint * pxEndPoint )
```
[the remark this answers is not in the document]
htibosch: No the typedef is still there.
The function header `xSendDHCPEvent()` appears in FreeRTOS_DHCP.h, but `NetworkEndPoint_t` is defined in Routing.h, which is normally included as the last one. Therefore, I had to make some forward declarations like:
~~~
struct xNetworkEndPoint;
~~~

FreeRTOS_IP.c
htibosch: Absolutely! But this function is declared only for backward compatibility, i.e. when `ipconfigCOMPATIBLE_WITH_SINGLE` is defined.
In that mode, there is only 1 interface, which has 1 IPv4 end-point. `ipv4_settings.ulGatewayAddress` is either a valid IP-address of a router, or it is 0.0.0.0.

FreeRTOS_IP.c
htibosch: The same answer as to `FreeRTOS_GetGatewayAddress()` : the function is only defined when `ipconfigCOMPATIBLE_WITH_SINGLE` is used.

FreeRTOS_IP.c
htibosch: See above.

FreeRTOS_IP.c
htibosch: See response here above.

FreeRTOS_IP.c
htibosch: See above.

FreeRTOS_IP.c
htibosch: See above.

FreeRTOS_IP.c
htibosch: `uxIPHeaderSizeSocket` has become a global function, used from other modules as well.

portable/BufferManagement/BufferAllocation_2.c
htibosch: I have now changed the text, thanks.

htibosch (commented):
Aniruddha, thank you for reviewing!

FreeRTOS_IP.c
AniruddhaKanhere: Makes sense! Thanks!

AniruddhaKanhere (approved):

FreeRTOS_IP.c
```diff
@@ -4184,7 +4267,7 @@ const char * FreeRTOS_strerror_r( BaseType_t xErrnum,
 
         default:
             /* Using function "snprintf". */
-            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( printf_signed ) xErrnum );
+            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", xErrnum );
```
dan4thewin, line 4270: We've removed snprintf from other code due to MISRA 21.6.  We should consider doing so here too.  The justification provided in MISRA.md is not persuasive.

FreeRTOS_Stream_Buffer.c
```diff
@@ -43,6 +43,226 @@
...
+ */
+BaseType_t xStreamBufferLessThenEqual( const StreamBuffer_t * pxBuffer,
+                                       const size_t uxLeft,
+                                       const size_t uxRight )
+{
+    BaseType_t xReturn;
+    size_t uxTail = pxBuffer->uxTail;
+
+    /* Returns true if ( uxLeft < uxRight ) */
+    if( ( ( ( uxLeft < uxTail ) ? 1U : 0U ) ^ ( ( uxRight < uxTail ) ? 1U : 0U ) ) != 0U )
```
dan4thewin, line 216: Is there a simpler way to express this?

FreeRTOS_TCP_IP.c
```diff
@@ -527,6 +527,8 @@
  *
  * @note Sequence of calling (normally) :
  * IP-Task:
+ *		prvIPTask()						// The IP-task
+ *		prvCheckNetworkTimers()			// The regular checks
```
dan4thewin, line 531: Does this code still use tabs?

test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
```diff
@@ -22,14 +24,19 @@ const uint8_t ucMACAddress[ 6 ];
...
-    struct xNetworkEndPoint * pxLocalEndPointPointer = &xLocalEndPoint;
+    struct xNetworkEndPoint * pxLocalEndPointPointer = nondet_bool() ? &xLocalEndPoint : NULL;
 
     /* Assume that the list of interfaces/endpoints is not initialized. */
     __CPROVER_assume( pxNetworkInterfaces == NULL );
     __CPROVER_assume( pxNetworkEndPoints == NULL );
 
+    for( x = 0; x < ipconfigARP_CACHE_ENTRIES; x++ )
+    {
+        xARPCache[ x ].pxEndPoint = nondet_bool() ? NULL : malloc( sizeof( NetworkEndPoint_t ) );
```
dan4thewin, line 37: CBMC malloc() can return NULL, as I recall.

dan4thewin (changes requested):

FreeRTOS_IP.c
```diff
@@ -3501,12 +3475,17 @@ uint16_t usGenerateProtocolChecksum( const uint8_t * const pucEthernetBuffer,
...
-        ( usChecksum == ipINVALID_LENGTH ) )
-    {
-        /* NOP if ipconfigHAS_PRINTF != 0 */
-        FreeRTOS_printf( ( "CRC error: %04x location %ld\n", usChecksum, xLocation ) );
-    }
+    #if ( ipconfigHAS_PRINTF == 1 )
+        if( ( usChecksum == ipUNHANDLED_PROTOCOL ) ||
+            ( usChecksum == ipINVALID_LENGTH ) )
+        {
+            /* NOP if ipconfigHAS_PRINTF != 0 */
```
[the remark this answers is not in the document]
htibosch: As `ipconfigHAS_PRINTF` has just been tested, I think the comment has become useless and can be removed.

FreeRTOS_IP.c
```diff
@@ -3824,7 +3803,7 @@ void vReturnEthernetFrame( NetworkBufferDescriptor_t * pxNetworkBuffer,
         else if( ( uxMinLastSize * ipMONITOR_PERCENTAGE_90 ) > ( uxMinSize * ipMONITOR_PERCENTAGE_100 ) )
         {
             uxMinLastSize = uxMinSize;
-            FreeRTOS_printf( ( "Heap: current %d lowest %u\n", ( printf_signed ) xPortGetFreeHeapSize(), ( printf_unsigned ) uxMinSize ) );
+            FreeRTOS_printf( ( "Heap: current %d lowest %u\n", xPortGetFreeHeapSize(), uxMinSize ) );
```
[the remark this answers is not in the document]
htibosch: Last week I spent a lot of time to sort out how printf compiler warnings can be avoided while getting the format correct, see my comment in [PR #197](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/197#pullrequestreview-595091959). That works for both 32 and 64-bit platforms.

This is one example:
~~~
    uint32_t ulValue = 0x80004000U;
    printf( "int32 = %x\n", ( unsigned int ) ulValue );
~~~
Earlier, I have been playing with `printf_signed ` and some other typedef's, but Coverity didn't like them.
I would like to propose to use casts `( unsigned int )` as in the example, because that is what `%x` is expecting.

FreeRTOS_IP.c
htibosch: Luckily there are very few occurrences of `snprintf()`. What alternative would you recommend? I suppose that `itoa()` is worse than `snprintf()`?

FreeRTOS_Stream_Buffer.c
htibosch: It use to be much simpler, but MISRA didn't like it because the XOR was applied to two boolean expressions:
~~~c
if( ( uxLeft < uxTail ) ^ ( uxRight < uxTail ) )
~~~
and also: XOR results in a number, not a boolean :-(

FreeRTOS_TCP_IP.c
htibosch: Sorry yes, I found some tabs within the comments. I will remove them.

htibosch (commented):
Dan, thank you very much for this review. Aniruddha will look at the `malloc()` returning NULL.

test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
AniruddhaKanhere: It can, but we have not switched to the flag that enables it since it made proofs slower - leading to slow PR check runs. Thus, we need to have that kind of statement to mimic the behavior of `mallocCanFail`.

htibosch:
I pushed some formatting changes ( which removes tabs in comments ).
And then I applied Uncrustify again.
Now I see many more formatting problems, however those files are not part of this PR:
~~~
/test/unit-test/ConfigFiles/FreeRTOSConfig.h
/portable/BufferManagement/BufferAllocation_1.c
/portable/NetworkInterface/pic32mzef/NetworkInterface_wifi.c
/portable/NetworkInterface/STM32Hxx/stm32hxx_hal_eth.h
/portable/NetworkInterface/ATSAM4E/ethernet_phy.c
/portable/NetworkInterface/ATSAM4E/gmac.c
/portable/NetworkInterface/ksz8851snl/ksz8851snl.c
/portable/NetworkInterface/DriverSAM/gmac_SAM.c
/tools/tcp_utilities/plus_tcp_demo_cli.c
~~~

AniruddhaKanhere (dismissed):

FreeRTOS_IP.c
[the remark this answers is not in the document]
dan4thewin: That seems fine for unsigned values.

FreeRTOS_IP.c
dan4thewin: Is this list not already exhaustive?  What value could xErrnum have that is not one of the cases?

test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
dan4thewin: How much slower?

FreeRTOS_Stream_Buffer.c
dan4thewin: maybe something like:
```c
xReturn = pdFalse;
if( ( uxLeft - uxTail ) <= ( uxRight - uxTail ) )
{
    xReturn = pdTRUE;
}
```

FreeRTOS_IP.c
htibosch: Yes good answer, in this case. All other uses of `snprintf()` ( not many ) have to do with debug logging, which is not an essential part of the library.

FreeRTOS_Stream_Buffer.c
htibosch: That's great! 
Just to be sure, I tested it for all 1000 combinations under 10 and got the same results as with the former complicated formula. Thanks!

FreeRTOS_IP.c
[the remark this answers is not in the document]
htibosch: In a meanwhile I also tested the formats with signed numbers. The outputs are as expected.

FreeRTOS_IP.c
htibosch: There was a second snprintf() with a `%s`: I have replaced that with a `strncpy()`:
~~~
    if( uxLength > 0U )
    {
        ( void ) strncpy( pcBuffer, pcName, uxLength - 1U );

        /* Make sure the result is nul-terminated. */
        pcBuffer[ uxLength - 1U ] = '\0';
    }
~~~

htibosch (commented):
@dan4thewin :
I have removed the last `snprintf()`, simplified the expression for stream-buffer, and re-tested the `printf()` formats in logging statements.

Thanks, Hein

htibosch:
PS. The function `FreeRTOS_inet_ntop6()` still contains a call to `snprintf()`, in order to print a hex number in an IPv6 address.
Will will keep it and submit that after this PR.

test/cbmc/proofs/ARP/ARPGetCacheEntryByMac/ARPGetCacheEntryByMac_harness.c
AniruddhaKanhere: At that time, the proofs were taking 16 minutes, and using the flag pushed that to 25 minutes.
That was very slow for a person waiting on the results of the PR checks. We decided that we rather not switch to that flag - at least in the IPv6 branch where the proofs are slower than the one in the IPv4 branch.

dan4thewin (approved):

AniruddhaKanhere (approved):

PR #193, opened 2021-02-16
Addition of MSP432E401Y EMAC driver 

nkinar:
Third-party code for an MSP432E401Y NetworkInterface driver.

Description
-----------
This is an example driver for the Texas Instruments SimpleLink MSP432E401Y microcontroller with internal Ethernet transceiver.  Code to setup the driver is provided in the README.md file.  Additional middleware is also provided. 
Some changes were made for robustness after some additional testing.

Test Steps
-----------
Tested using UDP and TCP sockets, ping floods, nmap and in the context of a FreeRTOS+TCP application.

Related Issue
-----------
https://forums.freertos.org/t/freertos-tcp-crashing-on-cortex-m4f-and-possible-bug-msp432e401y/11606/17


AniruddhaKanhere:
Hello @nkinar!

Thank you for this PR and the interesting discussion on the freertos.org forums.
I see that the formatting check is failing. That check makes sure that the source is formatted similar to the FreeRTOS+TCP source files. Would you mind fixing the formatting? Or if you like, I will be more than happy to do it for you.

If you would like to do it yourself, then the below information will be useful:
- A tool called Uncrustify is used for this check.
- We do need a specific version of the tool -> 0.66.1
- The config file can be found in the `./tools` directory [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/tools/uncrustify.cfg).
- All the source .[ch] files that are modified in this PR are required to be formatted with this tool.

Regards,
Aniruddha

nkinar:
@AniruddhaKanhere thanks for your help.  I ran uncrustify on the code and then added the updated commits.  The formatting is still failing - can you check and see where the fail is occurring?  I will update accordingly if required.

AniruddhaKanhere:
@nkinar - I see two files failing the check:
`/portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c`
and
`portable/NetworkInterface/ThirdParty/MSP432/NetworkMiddleware.c`

nkinar:
@AniruddhaKanhere, sure that sounds good; thanks for your help.  Can you fix this?  Thanks.

nkinar:
@AniruddhaKanhere that is great!!  Thanks.

AniruddhaKanhere (approved):

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+    BaseType_t txResp;
+
+    NIDataOutput.pxDescriptor = pxDescriptor;
+    NIDataOutput.xReleaseAfterSend = xReleaseAfterSend;
+    txResp = xQueueSend( xQueueTX, &NIDataOutput, 0 );
+
+    if( txResp == pdTRUE )
+    {
+        xTaskNotifyGiveIndexed( xTaskToNotifyEthernetTX, 0 );
+        return pdTRUE;
```
gkwicker, line 838: nit: you can simplify this function to use a single `return` statement by introducing a helper variable, e.g.:

```c
BaseType_t returnValue = pdFALSE;
.
.
.
if( txResp == pdTRUE )
{
    xTasknotifyGiveIndexed( xTaskToNotifyEthernetTX, 0 );
    returnValue = pdTRUE;
}
return returnValue;
```

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+bool isEMACLinkUp()
+{
+    uint8_t uci8PHYAddr = 0; /* refers to the internal PHY */
+    uint16_t check;
+
+    check = EMACPHYRead( EMAC0_BASE, uci8PHYAddr, EPHY_BMSR );
+
+    if( ( ( check & EPHY_BMSR_LINKSTAT ) == 0 ) || ( check & EPHY_BMSR_RFAULT ) )
+    {
+        return false; /* link is not up */
```
gkwicker, line 864: See previous comment about rewriting with a single `return`, that would work here as well.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+    check = false;
+
+    for( ; ; )
+    {
+        check = isEMACLinkUp();
+
+        if( ( check == true ) && ( networkUP == false ) )
+        {
+            networkUP = true;
+            vTaskDelay( pdMS_TO_TICKS( 10 ) );
```
gkwicker, line 885: Should this delay value be configurable?  Do all applications need link up/down notifications within 10ms, or would some applications (i.e. low power apps) prefer to wake up less frequently here?

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+        }
+        else if( ( networkUP == true ) && ( check == false ) )
+        {
+            /*   FreeRTOS will poll xNetworkInterfaceInitialise() to check if the network is up.
+             *  After FreeRTOS_NetworkDown() is called, so there is no corresponding FreeRTOS_NetworkUp() function... */
+            networkUP = false;
+            FreeRTOS_NetworkDown();
+            vTaskDelay( pdMS_TO_TICKS( 10 ) );
+        }
```
gkwicker, line 894: No delay happens here `if( ( check == true ) && ( networkUP == true ) )`.  Is that desired?  The code would be simpler if lines 885/893 were removed and the delay was placed after the tests on lines 882 & 887, e.g.:

```c
if( ( check == true ) && ( networkUP == false ) )
{
    networkUP = true;
}
else if( ( networkUP == true ) && ( check == false ) )
{
    networkUP = false;
    FreeRTOS_NetworkDown();
}
vTaskDelay( pdMS_TO_TICKS( 10 ) );
```

gkwicker (changes requested):
See my comments inline for specific items.

General comment – it might be helpful to list all of the tasks this network interface creates and their purpose in the documentation (I counted 5 tasks).  Also, would it be possible to combine some of the tasks related to checking link up/down, network state & reset, etc.?  Each task requires a stack, so if some of these could be combined it could help app developers save memory.

yourslab:
Hi @nkinar. Thank you for your contribution. We would like to merge it to main but can you please address the last few comments made?

nkinar:
@yourslab Thanks.  I am just testing the driver on the hardware and I can respond to the comments within the next few days.  I am so sorry about the delay.

yourslab:
No worries. Thanks for the hard work and getting it tested on hardware 🙂

nkinar:
> See my comments inline for specific items.
> 
> General comment – it might be helpful to list all of the tasks this network interface creates and their purpose in the documentation (I counted 5 tasks). Also, would it be possible to combine some of the tasks related to checking link up/down, network state & reset, etc.? Each task requires a stack, so if some of these could be combined it could help app developers save memory.

Tasks created for the network interface and the purpose of each task are now listed in the documentation.
See the "List of Tasks" section.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+
+/* Queue handles */
+QueueHandle_t xQueueRX;
+QueueHandle_t xQueueTX;
+
+/* MAC address */
+uint8_t ucpui8MACAddr[ ipMAC_ADDRESS_LENGTH_BYTES ];
+
+/* State variable that indicates whether the network is up or down */
+bool networkUP = false;
```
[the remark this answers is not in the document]
nkinar: > See my comments inline for specific items.
> 
> General comment – it might be helpful to list all of the tasks this network interface creates and their purpose in the documentation (I counted 5 tasks). Also, would it be possible to combine some of the tasks related to checking link up/down, network state & reset, etc.? Each task requires a stack, so if some of these could be combined it could help app developers save memory.

The link up/down, network state & reset tasks have now been combined as suggested.  As provided in the
README.md table, there are now only three tasks used in the driver code (NetworkInterface.c).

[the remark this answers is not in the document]
nkinar: networkUP is now static and external code can obtain the value of networkUP by calling vPublicCheckNetworkInterfaceUp().
The vPublicCheckNetworkInterfaceUp() is declared in the NetworkInterface.h file.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+
+bool loadMACInternal()
+{
+    uint32_t ui32User0, ui32User1;
+
+    /* Read the MAC address from internal flash. Bit[23:0] are stored in user register0, and Bit[47:24] are stored in user register1.
+     * The MAC address can be loaded from an external IC or can be loaded from the internal registers if the registers are not zero.
+     * The evaluation kit hardware uses reading from internal flash variables but this can also be used for production as well
+     * if the internal flash is programmed on the assembly line. */
+    FlashUserGet( &ui32User0, &ui32User1 );
```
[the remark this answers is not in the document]
nkinar: This is good, and it is indeed safer to use the code that you suggest.  Thank you.
The code has been changed accordingly.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+    /* enable and reset the internal MAC */
+    SysCtlPeripheralPowerOn( SYSCTL_PERIPH_EMAC0 );
+    SysCtlPeripheralPowerOn( SYSCTL_PERIPH_EPHY0 );
+    SysCtlPeripheralEnable( SYSCTL_PERIPH_EMAC0 );
+    SysCtlPeripheralEnable( SYSCTL_PERIPH_EPHY0 );
+    SysCtlPeripheralReset( SYSCTL_PERIPH_EMAC0 );
+    SysCtlPeripheralReset( SYSCTL_PERIPH_EPHY0 );
+
+    while( !SysCtlPeripheralReady( SYSCTL_PERIPH_EMAC0 ) )
+    {
```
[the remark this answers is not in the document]
nkinar: Yes, a timeout has been added here.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+
+    /* register the interrupt handler for the Ethernet MAC */
+    EMACIntRegister( EMAC0_BASE, ethernetIntHandler );
+
+    /* enable the interrupts for the EMAC */
+    EMACIntClear( EMAC0_BASE, EMACIntStatus( EMAC0_BASE, false ) );
+    IntEnable( INT_EMAC0 );
+    EMACIntEnable( EMAC0_BASE, EMAC_INTERRUPTS );
+
+    if( xTaskGetSchedulerState() == taskSCHEDULER_RUNNING )
```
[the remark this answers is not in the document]
nkinar: You are right - the scheduler might indeed start running.  The code that you suggest has
been added in and and utilized.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+    ulstatus = EMACStatusGet( EMAC0_BASE );
+
+    while( ulstatus )
+    {
+        ulstatus = EMACStatusGet( EMAC0_BASE );
+        ulcnt++;
+
+        if( ulcnt == ETH_STARTUP_TIMEOUT )
+        {
+            if( xTaskGetSchedulerState() == taskSCHEDULER_RUNNING )
```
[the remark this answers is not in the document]
nkinar: It is safer to use a helper variable.  The code has been updated accordingly.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,981 @@
...
+    uint32_t ui32Temp;
+
+    ui32Temp = EMACIntStatus( EMAC0_BASE, true );
+
+    if( ui32Temp & EMAC_INT_RECEIVE )
+    {
+        /* RX */
+        processReceivedPacket();
+    }
+    else if( ui32Temp & EMAC_INT_TRANSMIT )
```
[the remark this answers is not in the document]
nkinar: Yes, EMAC_INT_RECEIVE and EMAC_INT_TRANSMIT could be true in the same interrupt and the code has been updated to check at the same time.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: Yes, this is a good idea.  The code is updated.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: Yes, the code is updated.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: After more investigation, the delay is not required, so it has been removed from the code.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: No delay is required, so it was removed from the code after testing on the hardware.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -271,8 +285,13 @@ bool setupEMAC()
     SysCtlPeripheralReset( SYSCTL_PERIPH_EMAC0 );
     SysCtlPeripheralReset( SYSCTL_PERIPH_EPHY0 );
 
-    while( !SysCtlPeripheralReady( SYSCTL_PERIPH_EMAC0 ) )
+    /* The peripheral should always start, but there is a wait with timeout here to prevent an infinite loop*/
+    ul32Loop = 0;
+
+    while( !SysCtlPeripheralReady( SYSCTL_PERIPH_EMAC0 ) && ( ul32Loop < ETH_STARTUP_TIMEOUT ) )
```
gkwicker, line 291: If you pass line 296 with `ul32Loop > ETH_STARTUP_TIMEOUT`, this code continues as if everything is fine when the EMAC never became ready.  In this case it should not continue with the initialization and should return `false`.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -247,10 +257,14 @@ bool setupEMAC()
 {
     uint32_t ul32Loop;
     bool rv;
+    bool interruptsMasked;
```
gkwicker, line 260: maintainability: it's better to initialize these inline in case a future maintainer inserts code between declaration and assignment, e.g.:

```
bool interruptsMasked = false;
```

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -827,6 +824,8 @@ BaseType_t xNetworkInterfaceOutput( NetworkBufferDescriptor_t * const pxDescript
 {
     struct NetworkInterfaceDataOut NIDataOutput;
     BaseType_t txResp;
+    BaseType_t returnValue;
```
gkwicker, line 827: maintainability: prefer `BaseType_t returnValue = pdFALSE;` here.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -856,20 +855,22 @@ bool isEMACLinkUp()
 {
     uint8_t uci8PHYAddr = 0; /* refers to the internal PHY */
     uint16_t check;
+    BaseType_t returnValue;
+    returnValue = pdTRUE;
```
gkwicker, line 859: maintainability: it's better to initialize return values to fail/error rather than success, e.g.:

```c
BaseType_t returnValue = pdFALSE;

check = EMACPHYRead( EMAC0_BASE, uci8PHYAddr, EPHY_BMSR );

if( ( ( check & EPHY_BMSR_LINKSTAT ) != 0 ) && !( check & EPHY_BMSR_RFAULT ) )
{
    returnValue = pdTRUE; /* link is up */
}

return returnValue; /* return link status */

File: portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c [hunk 15]
```diff
@@ -856,20 +855,22 @@ bool isEMACLinkUp()
...
     check = EMACPHYRead( EMAC0_BASE, uci8PHYAddr, EPHY_BMSR );
 
     if( ( ( check & EPHY_BMSR_LINKSTAT ) == 0 ) || ( check & EPHY_BMSR_RFAULT ) )
     {
-        return false; /* link is not up */
+        returnValue = pdFALSE; /* link is not up */
     }
 
-    return true; /* link is up */
+    return returnValue; /* link is up */
```
gkwicker, line 868: Comment is incorrect, link may be either up or down here.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -856,20 +855,22 @@ bool isEMACLinkUp()
...
 }
 
 
-/* A task to check and see if the link is up or down by polling an EMAC register*/
-void prvCheckLinkUpOrDownTask( void * pvParameters )
+/* A task to check and see if the link is up or down by polling an EMAC register */
+void prvCheckLinkUpOrDownNetStateTask( void * pvParameters )
 {
     bool check;
```
gkwicker, line 875: see previous comments, prefer: `bool check = false;`.

nit: this variable could have a more descriptive name, e.g. `checkLinkStatus`.

gkwicker (changes requested):
The code is looking better, just a few small comments.

yourslab:
> I am also seeing that the formatting check is failing. I ran uncrustify on the files, but I am uncertain why the formatting check fails when run on the server.

Can you verify that you're using uncrustify 0.67 and using the follow uncrustify config? https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/tools/uncrustify.cfg

nkinar:
@yourslab  I was using uncrustify 0.66 on Windows, and I am now using uncrustify 0.67.  I also downloaded the config file that you provided in the link.  Thanks.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: Done, this has been updated.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: I agree, sometimes I think in code prior to C99.  This has now been updated.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: Yes, good idea.  This has been changed.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: I agree, and the code has been updated.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: This has now been changed.  Thanks.

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
nkinar: I changed the variable name to `checkLinkStatus` and also the boolean.  Thanks.

nkinar:
@gkwicker Thanks for the comments.  I've now updated the code again and tested it on the hardware.

htibosch:
Hello @nkinar, thanks a lot for these contributions.

I am sorry that somehow we didn't get in touch earlier.

Normally, a Network Interface starts a single helping task, which takes care of reception, transmission, PHY status polling, clearing DMA descriptors after a packet has been sent.

There are a few reasons to use a single task, also called a "deferred interrupt handler":

- You need less resources when creating less tasks
- The peripheral can not be shared among tasks without introducing a mutex
- Especially the PHY access must be protected against "parallel access"
- No need for critical sections, no mutex.

The ETH (RX/TX/ERR) interrupts are kept as short as possible. They will set a flag and call a function like `vTaskNotifyGiveFromISR()` to wake up the handler.

About `EMACPHYConfigSet()` : if your PHY is 100 Mbps, you could also consider to use the generic [PHY handler](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/portable/NetworkInterface/Common/phyHandling.c). You find an example of how it is used [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/portable/NetworkInterface/STM32Fxx/NetworkInterface.c#L217).
~~~c
>    interruptsMasked = false;
~~~
In FreeRTOS projects, the words `true` and `false` are normally coded as `pdTRUE` and `pdFALSE`. Also many C-compilers might not implement the type `bool`.

About `xNetworkInterfaceOutput()` : normally the sending is done within this function. After the function returns, you are not supposed to access `pxDescriptor` any more ( unless `xReleaseAfterSend` is true ).
When the function exits, the packet is queued for transmission, possibly in a DMA descriptor.
I suppose that you have read about `xReleaseAfterSend`? It indicates whether the ownership is passed to the driver or not. When `ipconfigZERO_COPY_TX_DRIVER` is `1`, `xReleaseAfterSend` will always be true, meaning that the driver will own the network buffer.

~~~c
    if( xTaskGetSchedulerState() == taskSCHEDULER_RUNNING )
    {
        FreeRTOS_NetworkDown();
~~~
Normally, all network interfaces may assume that the scheduler is running, otherwise `xNetworkInterfaceInitialise()` can not be called, it is called from the IP-task.

And also "normally", the function `xNetworkInterfaceInitialise()` is the first point of action in the driver: it initialises the EMAC, the PHY, DMADMA descriptors, and buffers.
~~~c
        vTaskDelay( pdMS_TO_TICKS( ETH_UP_DELAY_MS ) );
~~~
There is no need to put a `vTaskDelay()` here, the IP-task already pauses 3 seconds between the calls:
~~~c
#ifndef ipINITIALISATION_RETRY_DELAY
	#define ipINITIALISATION_RETRY_DELAY    ( pdMS_TO_TICKS( 3000U ) )
#endif
~~~
This is indeed correct:
~~~c
if( ( checkLinkStatus == true ) && ( networkUP == false ) )
{
    networkUP = true;
}
~~~
meaning that `xNetworkInterfaceInitialise()` will return pdPASS only when the PHY link status is up.

About transmission: normally when the link status is low, all packets to be transmitted **will be dropped**. If you queue them for transmission you might get many DMA errors and see some long time-outs.

Finally I want to show an example of `xNetworkInterfaceInitialise()`:
~~~c
    BaseType_t xNetworkInterfaceInitialise( void )
    {
        BaseType_t xResult;
    
        if( xMacInitStatus == eMACInit )
        {
    		if( initialise_emac_and_phy() == pdPASS )
    		{
    			xMacInitStatus = eMACPass;
    		}
    		else
    		{
    			xMacInitStatus = eMACFail;
    		}
        } /* if( xMacInitStatus == eMACInit ) */
    
        if( xMacInitStatus != eMACPass )
        {
            /* EMAC initialisation failed, return pdFAIL. */
            xResult = pdFAIL;
        }
        else
        {
            xResult = ( xPhyObject.ulLinkStatusMask != 0U ) ? pdPASS : pdFAIL;
        }
    
        /* When returning non-zero, the stack will become active and
         * start DHCP (if configured) */
        return xResult;
    }
~~~
The above function `xNetworkInterfaceInitialise()` will be called from the IP-task.
It makes sure that `initialise_emac_and_phy()` will only be called once.

The MSP432E401Y is a nice CPU, and I think it is worth to give it a great network driver.
Thank you for this effort!
Hein

portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c
```diff
@@ -0,0 +1,987 @@
...
+{
+    BaseType_t rv;
+    struct NetworkInterfaceDataIn NIDataIn;
+
+    NIDataIn.ucbuff = uc8Buf;
+    NIDataIn.ulbuff_siz = ul32FrameLen;
+    NIDataIn.ulindx = ulindex;
+
+    /* send the data into the queue */
+    rv = xQueueSendFromISR( xQueueRX, &NIDataIn, &xHigherPriorityTaskWokenRX );
```
htibosch, line 588: Oops, `xHigherPriorityTaskWokenRX` is cleared when the application starts. You driver will never clear it again.
If is declared as a global variable, is it cleared elsewhere?
~~~c
     /* send the data into the queue */
     rv = xQueueSendFromISR( xQueueRX, &NIDataIn, &xHigherPriorityTaskWokenRX );

     if( rv == pdTRUE )
     {
         vTaskNotifyGiveFromISR( xTaskToNotifyEthernetRX, &xHigherPriorityTaskWokenRX );
-        portYIELD_FROM_ISR( xHigherPriorityTaskWokenRX );
     }
     else
     {
         DMAFreeDescriptorRX( ulindex );
     }
+    portYIELD_FROM_ISR( xHigherPriorityTaskWokenRX );
~~~
Suppose that `xQueueSendFromISR()` sets `xHigherPriorityTaskWokenRX ` while returning `pdFALSE`, then `portYIELD_FROM_ISR()` will not be called?

nkinar:
@htibosch thanks so much for your detailed comments; this is much appreciated.  However, due to other commitments, I do not have the time to make additional changes.  If you have the time and if you or anyone else would like to do so, then that would be great.  I do not mind if the code is not contributed to the kernel.  Thank you.

gkwicker (dismissed):

htibosch:
> @htibosch thanks so much for your detailed comments; this is much appreciated. However, due to other commitments, I do not have the time to make additional changes.

All right, no problem.

I just went through NetworkInterface.c again and made some essential changes, and also a few changes to please MISRA checks, like the 'U' after a few constant values. And finally I inserted "static" in front of all static functions, some compilers could complain about that.

I uploaded the adapted [NetworkInterface.c](https://raw.githubusercontent.com/htibosch/FreeRTOS-Plus-TCP/adding_driver_MSP432/portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c) to my github repo.

Could you please have a look with e.g. Beyond Compare or WinMerge to check the changes? The most important change is in `xNetworkInterfaceOutput()`.

Thanks,

nkinar:
@htibosch you are awesome.  Thank you so much my friend!  I will check the changes and also run a test on the hardware and then let you know.

nkinar:
@htibosch thanks for the intelligent changes to the code.  These are great.  I tested on actual hardware and had to make minor changes (as listed below).  IMHO, the MSP432 platform can be finnicky.

1.  In `prvCheckLinkUpOrDownNetStateTask`, the `vTaskDelay()` had to be removed for the Ethernet plug event to be detected.  I think this is because when the EMAC phy register is read by polling via the `isEMACLinkUp()` function, and a delay causes the change in the register to be missed.  Unfortunately, there is no way on this microcontroller to set up an interrupt to detect the Ethernet plug event.

2. vPublicTurnOnEMAC() should not be static since it can be called from external code?  This is declared in NetworkInterface.h. 

Thanks again so much for your expertise.

htibosch:
> thanks for the intelligent changes to the code.

It has yet to be proven if the proposed changes are really intelligent :-)

>  1. In `prvCheckLinkUpOrDownNetStateTask`, the `vTaskDelay()` had to be removed for the Ethernet plug event to be detected.  I think this is because when the EMAC phy register is read by polling via the `isEMACLinkUp()` function, and a delay causes the change in the register to be missed. 

Would it matter if you miss a short drop in the link status? I don't think so. Only if the Link Status drops for longer that a few seconds, I would respond to it.

> Unfortunately, there is no way on this microcontroller to set up an interrupt to detect the Ethernet plug event.

Yes right, normally the PHY can trigger an interrupt. I have never used it though.

A while ago I played with a Xilinx Zynq. When I polled the PHY very frequently ( 100 Hz ), the transfers were **heavily disturbed**. When I lowered the PHY polling speed, the transfers were good again. And then I decided not to poll the PHY for as long as packets are being received. With that method there were zero disruptions left.
( I was testing this while running IPERF3 in 2 directions ).
I am not sure if this disturbance is observed in other EMACs as well.

> 2. `vPublicTurnOnEMAC()` should not be static since it can be called from external code?  This is declared in NetworkInterface.h.

I tend to make the deferred interrupt handler task the **owner** of the EMAC and the PHY. I wouldn't give direct access to the hardware from a public function.

PS. you have not implemented the following function:
~~~c
BaseType_t xGetPhyLinkStatus( void )
{
    return networkUP;
}
~~~

It might be called from an application. Can you add it please?

@AniruddhaKanhere, I approve this PR!

nkinar:
@htibosch thanks for the comments; again this is much appreciated.

> Would it matter if you miss a short drop in the link status? I don't think so. Only if the Link Status drops for longer that a few 
> seconds, I would respond to it.

For some bizarre reason, polling using the `isEMACLinkUp` function misses the plug event if a delay is added to the `prvCheckLinkUpOrDownNetStateTask` function.  I suspect that the Ethernet plug event register is automatically cleared in a few processor cycles by the hardware.  MSP432 is unique and there are often some strange things that occur that might be hardware or compiler-bug related.  We could definitely look into this for a future patch.

> A while ago I played with a Xilinx Zynq. When I polled the PHY very frequently ( 100 Hz ), the transfers were heavily disturbed. When I lowered the PHY polling speed, the transfers were good again. And then I decided not to poll the PHY for as long as packets are being received. With that method there were zero disruptions left.
> ( I was testing this while running IPERF3 in 2 directions ).
> I am not sure if this disturbance is observed in other EMACs as well.

I wish MSP432 was similar to the Zynq.  The plug event is detected by checking a bit in the Ethernet PHY Basic Mode Status register and it appears that the bit is automatically cleared after some processor cycles.

> I tend to make the deferred interrupt handler task the owner of the EMAC and the PHY. I wouldn't give direct access to the hardware from a public function.

Yes, this is a good idea.  I changed the function names.

I also replaced all `bool` values with pdFALSE or pdTRUE.

nkinar:
Also, the `xGetPhyLinkStatus` has been added.

htibosch:
> For some bizarre reason, polling using the `isEMACLinkUp` function misses the plug event if a delay is added to the `prvCheckLinkUpOrDownNetStateTask` function

[phyHandler.c](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/portable/NetworkInterface/Common/phyHandling.c) is using:
~~~c
#define phyREG_01_BMSR          0x01U  /* Basic Mode Status Register. */

#define phyBMSR_AN_COMPLETE     0x20U  /* Auto-Negotiation process completed */
#define phyBMSR_LINK_STATUS     0x04U  /* PHY Link Status. */
#define phyBMSR_REMOTE_FAULT    0x10U  /* Remote fault. */

~~~
which corresponds to the TI definitions:
~~~c
#define EPHY_BMSR               0x01  // Ethernet PHY Basic Mode Status

#define EPHY_BMSR_ANC           0x20  // Auto-Negotiation Complete
#define EPHY_BMSR_LINKSTAT      0x04  // Link Status
#define EPHY_BMSR_RFAULT        0x10  // Remote Fault
~~~

I just tested on a different board, an STM32Fxx DISCO, to observe the PHY handling.

When I disconnect the device from the LAN, the bits `AN_COMPLETE` and `LINK_STATUS` both drop.
The driver stops sending packets, and it will wait for the `LINK_STATUS` to become high again.
Once it is high, the device starts a new Auto Negotiation (AN). 

The auto negotiation process typically lasts 1650 ms. `AN_COMPLETE` and `LINK_STATUS` both become high, and the device will re-program the PHY properties into the EMAC. These properties consists of speed (10/100) and (half/full) duplex, 

Maybe, when you have time, you can drive a LED that follows the `LINK_STATUS`? I can not imagine that the EMAC has any knowledge about the attached PHY. It should provide you a transparent data bus to the PHY, and let the application do what it wants to do.

PS. phyHandler.c has always ignored `BMSR_RFAULT`, but it might be good to observe it as well.

nkinar:
@htibosch thanks, I will take a look at this next weekend.

nkinar:
I am also not sure why the CI Checks / formatting is failing since I am using uncrustify to format the files.

AniruddhaKanhere:
> I am also not sure why the CI Checks / formatting is failing since I am using uncrustify to format the files.

I think my latest commit should fix that.

nkinar:
@AniruddhaKanhere thanks so much.  Uncrustify is really sensitive to whitespace!!

nkinar:
@htibosch, I set up the code to drive a LED that follows the LINK_STATUS.  The LED remains on when the network cable is plugged in, and off when the network cable is unplugged.  

Here is what I now think is happening after some investigation.  The MSP432 architecture has a "Tickless" Low Power Operation (https://www.freertos.org/TI_MSP432_Free_RTOS_Demo.html).  Apparently when a sleep statement is added to the task that detects the plug event, the microcontroller can enter a power-saving state and the change in the LINK_STATUS is missed.  The transition between the power-saving state and the operational state occurs when there is an interrupt.  However, there is no interrupt that can be associated with the LINK_STATUS, so the microcontroller does not wake up.

nkinar:
@AniruddhaKanhere I would be happy to help integrate any code as required or provide tests on hardware.

htibosch:
> @AniruddhaKanhere I would be happy to help integrate any code as required or provide tests on hardware.

@nkinar, @AniruddhaKanhere, I think that the driver is good as it is now, and that it can be merged with the main branch. Let's get experience with it, and please report if you find any issues.

Thanks, Hein

nkinar:
Thanks @htibosch, and we can definitely update as time goes along.  I will be using the driver for a number of projects, so the additional testing will continue to explore the operation.  @AniruddhaKanhere @gkwicker thanks for your reviews, help and comments; this is much appreciated.

AniruddhaKanhere (approved):

n9wxu (approved):

PR #196, opened 2021-02-18
Minor bug fix

AniruddhaKanhere:
Description
-----------
When the `FreeRTOS_sendto` function fails, the status returned by the `prvSendDHCPDiscover` function should be pdFAIL. Instead, a pdPASS was being returned.
This PR fixes that minor bug with an `else` statement.

Additionally, return value description was missing from the Doxygen comments. It was added as well.

Test Steps
-----------

Related Issue
-----------


FreeRTOS_DHCP.c
```diff
@@ -1449,8 +1451,10 @@
                  * returned to the stack. */
                 FreeRTOS_ReleaseUDPPayloadBuffer( pucUDPPayloadBuffer );
             }
-
-            xResult = pdTRUE;
+            else
+            {
+                xResult = pdPASS;
+            }
```
htibosch, line 1457: You are right, it makes no sense to wait for a reply to a messages that was never sent.

htibosch (approved):
@AniruddhaKanhere : thanks for this mini bug-fix. It looks good to me.

FreeRTOS_DHCP.c
```diff
@@ -1449,8 +1451,10 @@
                  * returned to the stack. */
                 FreeRTOS_ReleaseUDPPayloadBuffer( pucUDPPayloadBuffer );
             }
-
-            xResult = pdTRUE;
+            else
```
gkwicker, line 1454: Is there a unit test for this code?  If so, can it be updated to cover this condition?

gkwicker (commented):
One question, see comment.

FreeRTOS_DHCP.c
AniruddhaKanhere: There are no unit-tests as of now. We are working on adding unit-tests to get 100% coverage.

gkwicker (approved):

PR #197, opened 2021-02-20
fix compiler warnings about casting to format string

twpedersen:
Fix various warnings like:

/FreeRTOS-Plus-TCP/FreeRTOS_IP.c: In function 'FreeRTOS_strerror_r':
/FreeRTOS-Plus-TCP/FreeRTOS_IP.c:3395:60: error: format '%d' expects argument of type 'int', but argument 4 has type 'long int' [-Werror=format=]
             ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int32_t ) xErrnum );
                                                           ~^   ~~~~~~~~~~~~~~~~~~~
                                                           %ld

This was tested against an ARM a9 platform (zynq).

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -3977,13 +3977,13 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                     if( xResult != ( int32_t ) ulByteCount )
                     {
                         FreeRTOS_debug_printf( ( "lTCPAddRxdata: at %u: %d/%u bytes (tail %u head %u space %u front %u)\n",
-                                                 ( UBaseType_t ) uxOffset,
-                                                 ( BaseType_t ) xResult,
-                                                 ( UBaseType_t ) ulByteCount,
-                                                 ( UBaseType_t ) pxStream->uxTail,
-                                                 ( UBaseType_t ) pxStream->uxHead,
-                                                 ( UBaseType_t ) uxStreamBufferFrontSpace( pxStream ),
-                                                 ( UBaseType_t ) pxStream->uxFront ) );
+                                                 uxOffset,
+                                                 (int) xResult,
+                                                 (unsigned int) ulByteCount,
+                                                 pxStream->uxTail,
+                                                 pxStream->uxHead,
+                                                 uxStreamBufferFrontSpace( pxStream ),
+                                                 pxStream->uxFront ) );
                     }
                 }
             #endif /* ipconfigHAS_DEBUG_PRINTF */
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -3978,7 +3978,7 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                     {
                         FreeRTOS_debug_printf( ( "lTCPAddRxdata: at %u: %d/%u bytes (tail %u head %u space %u front %u)\n",
                                                  ( unsigned int ) uxOffset,
-                                                 ( int) xResult,
+                                                 ( int ) xResult,
                                                  ( unsigned int ) ulByteCount,
                                                  ( unsigned int ) pxStream->uxTail,
                                                  ( unsigned int ) pxStream->uxHead,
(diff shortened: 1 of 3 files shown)


htibosch:
Hello @twpedersen, thanks for this PR.
You are trying to tackle a problem that I have been seeing for years: compiler warnings about `printf` formats. In most cases, these are harmless, in some rare cases, they can lead to a crash, when `%s` is used.
To me it seems a big challenge to find a solution that works for all platforms, and for all compilers.
I also thought of using casts to the standard types `unsigned` and `int`. That works, except that MISRA doesn't like the use of these types.
The type `long` is confusing: it can be 32- or 64-bits.
I will compile your changes for several platforms and report back here.
Hein

FreeRTOS_Sockets.c
```diff
@@ -3977,13 +3977,13 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
...
                     {
                         FreeRTOS_debug_printf( ( "lTCPAddRxdata: at %u: %d/%u bytes (tail %u head %u space %u front %u)\n",
-                                                 ( UBaseType_t ) uxOffset,
-                                                 ( BaseType_t ) xResult,
-                                                 ( UBaseType_t ) ulByteCount,
-                                                 ( UBaseType_t ) pxStream->uxTail,
-                                                 ( UBaseType_t ) pxStream->uxHead,
-                                                 ( UBaseType_t ) uxStreamBufferFrontSpace( pxStream ),
-                                                 ( UBaseType_t ) pxStream->uxFront ) );
+                                                 uxOffset,
```
htibosch, line 3980: On a 64-bit platform I get several warnings like: 
```argument 2 has type 'size_t' {aka 'long unsigned int'}```
`uxOffset` is 64-bits and should be printed as a long.

**Review by htibosch — COMMENTED** [review]
I have been trying and testing on a 32-bit and on a 64-bits platform. I think that with the following casts and formats, there will be no compiler warnings, and the output will be as expected::
~~~c
	uint8_t ucValue = 0x80U;
	printf( "char = %x\n", ucValue );

	uint16_t usValue = 0x8040U;
	printf( "int16 = %x\n", usValue );

	uint32_t ulValue = 0x80004000U;
	printf( "int32 = %x\n", ( unsigned int ) ulValue );

	TickType_t uxTicks = 0x80004000U;
	printf( "TickType = %lx\n", uxTicks );

	UBaseType_t uxBase = 0x8000000040000000UL; /* may be truncated, no problem. */
	printf( "UBaseType = %lx\n", ( unsigned long ) uxBase );

	size_t uxValue = 0x8000000040000000UL; /* may be truncated, no problem. */
	printf( "size_t = %lx\n", ( unsigned long ) uxValue );

	printf( "sizeof = %lx\n", ( unsigned long ) sizeof ( uxValue ) );


                           32-bits  64-bits
type char             size    1        1  // equal size
type unsigned char    size    1        1
type short            size    2        2
type unsigned short   size    2        2
type int              size    4        4
type unsigned int     size    4        4

type long             size    4        8  // different size
type unsigned long    size    4        8
type size_t           size    4        8
type BaseType_t       size    4        8
type UBaseType_t      size    4        8
type TickType_t       size    4        8

~~~
As `TickType`, `UBaseType`, and `size_t` are 32-bits on an STM32, they are printed as such.
On UltraScale they are printed as 64-bit numbers.

File: FreeRTOS_Sockets.c (hunk 1 above)
Comment (twpedersen) [reply FreeRTOS_Sockets.c:3980]: thanks for testing. I'll see if casting to unsigned works.

File: FreeRTOS_TCP_WIN.c [hunk 2]
```diff
@@ -1183,9 +1183,9 @@
                         FreeRTOS_debug_printf( ( "lTCPWindowRxCheck[%d,%d]: seqnr %u exp %u (dist %d) SACK to %u\n",
                                                  ( int ) pxWindow->usPeerPortNumber,
                                                  ( int ) pxWindow->usOurPortNumber,
```
htibosch, line 1185: A cast from `short` or `char` to `int` would not be necessary as the compiler will up-cast them automatically when ellipsis are used.

htibosch (approved):

htibosch (approved):
I approve your PR, thanks for the patience. In a later PR, we should check all logging statements.

Thanks, Hein

AniruddhaKanhere (approved):

aggarg (approved):

PR #205, opened 2021-03-08
IPv6/Multi : A new usGenerateChecksum() that is faster and MISRA compliant

htibosch:
Description
-----------
The function `usGenerateChecksum()` calculates all checksums for all purposes: for the IP-headers ( IPv4 and IPv6 ), and for the protocol headers ( TCP, UDP, and ICMP ).

The current version of [`usGenerateChecksum()`](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/6ae29a62db3342182739999476c7d972d3dcf0e3/FreeRTOS_IP.c#L2820) was developed a long time ago. It was heavily optimised for an AVR32 processor ( UC3A ), which has a few handy instructions that ARM doesn't have.

I created a new version that is a lot **simpler** **and** **faster**. Also it is MISRA compatible. What it does is the following:

1. When the pointer to the buffer is not aligned, the first byte will be stored in `usTerm`, and the bytes in the accumulator will be swapped.
2. Now the pointer is 16-bit aligned, it will sum up 16 short values in each loop. When an optimisation flag is passed to the compiler ( `-O1` or `-Os` ), the code will be very efficient.
3. Store the remaining short values.
4. Store the last single byte in `usTerm`, if any.
5. Add the value of `usTerm` to the accumulator.
6. If the accumulator was swapped in step-1, swap it back.

The algorithm is only used when any of these two macros is defined as zero:
~~~c
#define ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM		( 1 )
#define ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM		( 1 )
~~~
For SAM ( MicroChip ), there are additional macros that must be set in order to activate the hardware offloading:
~~~c
#define ipconfigHAS_RX_CRC_OFFLOADING 1
#define ipconfigHAS_TX_CRC_OFFLOADING 1
~~~

Test Steps
-----------
I tested the new algorithm by comparing it with the existing one.

In the test I created packets with all sorts of alignments, with random contents, and with a length varying from 0 to 1514 bytes.  Each combination was tested.

I tested the new algorithm on real hardware:

- STM32F4x ( ST electronics )
- Zynq Ultrascale ( Xilinx, 64-bits )
- Zynq 7000 ( Xilinx, 32-bits )
- AT32UC3A0512 ( **big-endian**, Microchip )

The new algorithm is faster on an ARM-M4, same speed on the Xilinx parts, and slower on AVR32.

### DIFF
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -206,37 +206,6 @@ static void prvCallDHCP_RA_Handler( NetworkEndPoint_t * pxEndPoint );
 
 /*-----------------------------------------------------------*/
 
-/**
- * Used in checksum calculation.
- */
-typedef union _xUnion32
-{
-    uint32_t u32;      /**< The 32-bit member of the union. */
-    uint16_t u16[ 2 ]; /**< The array of 2 16-bit members of the union. */
-    uint8_t u8[ 4 ];   /**< The array of 4 8-bit members of the union. */
-} xUnion32;
-
-/**
- * Used in checksum calculation.
- */
-typedef union _xUnionPtr
-{
-    const uint32_t * u32ptr; /**< The pointer member to a 32-bit variable. */
-    const uint16_t * u16ptr; /**< The pointer member to a 16-bit variable. */
-    const uint8_t * u8ptr;   /**< The pointer member to an 8-bit variable. */
-} xUnionPtr;
-
-
-/**
- * @brief Utility function to cast pointer of a type to pointer of type NetworkBufferDescriptor_t.
- *
- * @return The casted pointer.
- */
-static portINLINE ipDECL_CAST_PTR_FUNC_FOR_TYPE( NetworkBufferDescriptor_t )
-{
-    return ( NetworkBufferDescriptor_t * ) pvArgument;
-}
-
 /*-----------------------------------------------------------*/
 
 /*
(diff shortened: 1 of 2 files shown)


FreeRTOS_IP.c
```diff
@@ -3510,144 +3541,94 @@ uint16_t usGenerateChecksum( uint16_t usSum,
...
-        /* Now add the secondary sum to the major sum, and remember if there was
-         * a carry. */
-        xSum.u32 = xSum2.u32 + xSource.u32ptr[ 1 ];
+        /* The alignment of 'pucData' has just been tested and corrected
+         * when necessary.
+         */
+        usPointer = ( const uint16_t * ) uxBufferAddress;
 
-        if( xSum2.u32 > xSum.u32 )
+        /* Sum 8 shorts in each loop. */
```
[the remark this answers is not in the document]
htibosch: I changed it as:
~~~c
-    /* Sum 8 shorts in each loop. */
+    /* Sum 'uxUnrollCount' shorts in each loop. */
~~~

FreeRTOS_IP.c
```diff
@@ -3510,144 +3541,94 @@ uint16_t usGenerateChecksum( uint16_t usSum,
...
-    uxAlignBits = ( ( ( uintptr_t ) pucNextData ) & 0x03U );
-
-    /*
-     * If pucNextData is non-aligned then the checksum is starting at an
-     * odd position and we need to make sure the usSum value now in xSum is
-     * as if it had been "aligned" in the same way.
-     */
-    if( ( uxAlignBits & 1UL ) != 0U )
+    uint32_t ulAccum = FreeRTOS_htons( usSum );
+    const uint16_t * usPointer;
```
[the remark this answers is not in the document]
htibosch: Sorry, yes of course!

htibosch (commented):
Thanks for your comments.
Hein

AniruddhaKanhere (dismissed):

AniruddhaKanhere:
Thanks for the update @htibosch.
Can you have a look at the failing checks? If you'd like, I can have a look at that for you.

AniruddhaKanhere (dismissed):

FreeRTOS_IP.c
```diff
@@ -3495,159 +3495,117 @@ uint16_t usGenerateProtocolChecksum( const uint8_t * const pucEthernetBuffer,
...
+            ulAccum += pusPointer[ 9 ];
+            ulAccum += pusPointer[ 10 ];
+            ulAccum += pusPointer[ 11 ];
+            ulAccum += pusPointer[ 12 ];
+            ulAccum += pusPointer[ 13 ];
+            ulAccum += pusPointer[ 14 ];
+            ulAccum += pusPointer[ 15 ];
+
+            uxBytesLeft -= sizeof( *pusPointer ) * uxUnrollCount;
+            pusPointer = &( pusPointer[ uxUnrollCount ] );
```
n9wxu, line 3574: MISRA 18.4 (pointer math) allows the ++ on the operator.  So using 
ulAccum += *pusPointer++;
is compliant AND it will suggest to the compiler to use an indirect post increment addressing mode.

FreeRTOS_IP.c
```diff
@@ -3495,159 +3495,117 @@ uint16_t usGenerateProtocolChecksum( const uint8_t * const pucEthernetBuffer,
...
-        xSum2.u32 = xSum.u32 + xSource.u32ptr[ 2 ];
-
-        if( xSum2.u32 < xSum.u32 )
+        /* Between 0 and 7 shorts might be left. */
+        while( uxBytesLeft >= sizeof( *pusPointer ) )
         {
-            ulCarry++;
+            ulAccum += *pusPointer;
+            uxBytesLeft -= sizeof( *pusPointer );
+            pusPointer = &( pusPointer[ 1 ] );
```
n9wxu, line 3582: see previous comment on the ++ operator.

FreeRTOS_IP.c
n9wxu: MISR compliant version is:
ulAccum += *(pusPointer++);

n9wxu (commented):
The new checksum function creates some unusual patterns that could be more clear and still follow MISRA.  It is possible it will be faster.

FreeRTOS_IP.c
htibosch: I didn't use the indirect post increment addressing mode, because I feared that the compiler would need 3 instructions: load, add, and increment, times 16.

But with optimisation enabled, the compiler only increments `pusPointer` with 32 at the end of each loop. Thus giving the same performance.

It certainly looks better than:
~~~c
pusPointer = &( pusPointer[ 1 ] );
~~~
Although Coverity did not complain about this type of assignment.

Thanks, Hein

htibosch:
@n9wxu wrote: 

> The new checksum function creates some unusual patterns that could be more clear and still follow MISRA. It is possible it will be faster.

What do you mean with "creates some unusual patterns" ?

FreeRTOS_IP.c
n9wxu: Depending upon how the compiler optimizes I would expect pusPointer to stay in a register for this function.  Then that addressing mode will leave the loop with pusPointer updated as a "side effect" of the single instruction.  I am doing a test.

n9wxu:
> @n9wxu wrote:
> 
> > The new checksum function creates some unusual patterns that could be more clear and still follow MISRA. It is possible it will be faster.
> 
> What do you mean with "creates some unusual patterns" ?

Sorry, I could have been more clear.  The expected idiom for stepping through an array with a pointer *ptr++ and not ptr = &ptr[1] so I was surprised to find it.  I think the array element dereference method is a natural pattern considering the examples shown in the MISRA documentation but the ++ pattern is still allowed and is more clear.

n9wxu (approved):

PR #209, opened 2021-03-10
Add unit-tests for FreeRTOS_ARP.c

AniruddhaKanhere:
Description
-----------
This PR adds unit-tests for the FreeRTOS_ARP file. It also lays the foundation for future unit-tests.
With this PR the line, function, and branch coverage is at 100%.

Test Steps
-----------

Related Issue
-----------


test/unit-test/CMakeLists.txt
```diff
@@ -1,13 +1,18 @@
-cmake_minimum_required ( VERSION 3.13.0 )
-project ( "FreeRTOS-Plus-TCP test"
-          VERSION 1.0.0
-          LANGUAGES C )
+# gcc -E -fpreprocessed -I include/ -I test/FreeRTOS-Kernel/include/ -I test/unit-test/ConfigFiles/ -fdirectives-only include/FreeRTOS_IP.h -o tmp.c
```
yanjos-dev, line 1: What is this comment for?

test/unit-test/ConfigFiles/FreeRTOSConfig.h
```diff
@@ -29,7 +29,8 @@
 #ifndef FREERTOS_CONFIG_H
 #define FREERTOS_CONFIG_H
 
-#include <unity.h>
+/*#include "unity.h" */
```
yanjos-dev, line 32: Can you remove this comment if it's not needed anymore?

test/unit-test/ConfigFiles/FreeRTOSConfig.h
```diff
@@ -117,7 +118,9 @@ void vConfigureTimerForRunTimeStats( void );    /* Prototype of function that in
  * uses the same semantics as the standard C assert() macro. */
 extern void vAssertCalled( unsigned long ulLine,
                            const char * const pcFileName );
-#define configASSERT( x )    if( x == 0 ) TEST_ABORT()
+#define configASSERT( x )    assert( x )
+/*TEST_ASSERT( x ) */
+/*if( x == 0 ) { TEST_FAIL() } */
```
yanjos-dev, line 123: Can you remove these if they are not needed?

test/unit-test/cmock/project.yml
```diff
@@ -2,7 +2,7 @@
 :cmock:
   :mock_prefix: mock_
   :when_no_prototypes: :warn
-  :enforce_strict_ordering: TRUE
+#  :enforce_strict_ordering: TRUE
```
yanjos-dev, line 5: Please remove this if it's no longer needed.

test/unit-test/cmock/project.yml
```diff
@@ -22,10 +22,13 @@
     - <stdbool.h>
     - <stdint.h>
     - <fcntl.h>
-    - "FreeRTOS.h"
-    - "FreeRTOS_IP.h"
-    - "FreeRTOS_IP_Private.h"
-    - "FreeRTOS_UDP_IP.h"
+    - <unity.h>
```
yanjos-dev, line 25: Is unity still needed? I saw earlier that it was removed from a file.

FreeRTOS_ARP.c
```diff
@@ -725,17 +725,16 @@ void FreeRTOS_OutputARPRequest( uint32_t ulIPAddress )
...
-                    for( xIndex = ( BaseType_t ) pxNetworkBuffer->xDataLength; xIndex < ( BaseType_t ) ipconfigETHERNET_MINIMUM_PACKET_BYTES; xIndex++ )
-                    {
-                        pxNetworkBuffer->pucEthernetBuffer[ xIndex ] = 0U;
-                    }
+                /* At this stage, the xDataLength is would be equal to sizeof( ARPPacket_t )
+                 * due to the call to vARPGenerateRequestPacket above. */
+                BaseType_t xIndex;
 
-                    pxNetworkBuffer->xDataLength = ( size_t ) ipconfigETHERNET_MINIMUM_PACKET_BYTES;
+                for( xIndex = ( BaseType_t ) pxNetworkBuffer->xDataLength; xIndex < ( BaseType_t ) ipconfigETHERNET_MINIMUM_PACKET_BYTES; xIndex++ )
```
htibosch, line 732: It is true that `xDataLength >= sizeof( ARPPacket_t )`, but it has not been proven that `xDataLength < ipconfigETHERNET_MINIMUM_PACKET_BYTES`.

An ARP packet is 42 bytes long, the minimum packet length is typically 60 bytes. So 18 bytes must be cleared.

The check for `xDataLength < ipconfigETHERNET_MINIMUM_PACKET_BYTES` is important, otherwise we might truncate longer packets.

**Background**: the macro `ipconfigETHERNET_MINIMUM_PACKET_BYTES` normally has a value of 60 bytes. That is the minimum length of an Ethernet packet. Shorter packets can be sent, but the collision detection might not work correctly. I have also seen that shorter packets got dropped.

Some users commented that the minimum should be 64 bytes: they are right, but the 4 Frame Check Sequence bytes (FCS) are generated automatically by the peripheral. The stack doesn't see them.

htibosch (commented):
Thanks for this long PR which covers all source code in FreeRTOS_ARP.c.

FreeRTOS_ARP.c
AniruddhaKanhere: How would the code reach a point where `xDataLength > ipconfigETHERNET_MINIMUM_PACKET_BYTES`?
The call to vARPGenerateRequestPacket unconditionally modifies the `xDataLength = sizeof( ARPPacket_t )`.
Thus, the value of `xDataLength` would always be less than `ipconfigETHERNET_MINIMUM_PACKET_BYTES`.

Am I missing something here?

htibosch: You are right that `ipconfigETHERNET_MINIMUM_PACKET_BYTES` ought to be longer than `sizeof( ARPPacket_t )`, but it is not necessarily true. If it is defined as 40, we should not trim the packet here:
~~~c
pxNetworkBuffer->xDataLength = ( size_t ) ipconfigETHERNET_MINIMUM_PACKET_BYTES;
~~~

htibosch: By the way, the packet length is being checked in 4 locations: ARP, TCP, UDP, and in the function `vReturnEthernetFrame()` ( FreeRTOS_IP.c ).
It is not much code, but we could create a function that is called from those 4 locations. The logic will remain the same, symbolically:
~~~
if( xDataLength < ipconfigETHERNET_MINIMUM_PACKET_BYTES )
{
    /* Clear the extra bytes. */
    /* Set the new, extended length. */
    xDataLength = ipconfigETHERNET_MINIMUM_PACKET_BYTES;
}
~~~
And only if `ipconfigETHERNET_MINIMUM_PACKET_BYTES` is defined as a macro.

test/unit-test/cmock/project.yml
AniruddhaKanhere: Yes, here it is needed... :)

FreeRTOS_ARP.c
```diff
@@ -808,6 +808,7 @@ BaseType_t xARPWaitResolution( uint32_t ulIPAddress,
                 break;
             }
 
+            /* Decrement the count. */
             uxSendCount--;
```
htibosch, line 812: I compared your branch [AddHeaders](https://github.com/AniruddhaKanhere/FreeRTOS-Plus-TCP/tree/AddHeaders) with the HEAD version of my repo, and now it looks all good.

htibosch (dismissed):
I only looked at the changes in the +TCP library sources ( FreeRTOS_ARP.c ), and that looks all good, thanks.
Hein

yanjos-dev (dismissed):

.github/workflows/ci.yml
```diff
@@ -35,11 +33,7 @@ jobs:
       - name: Coverage
         run: |
           make -C test/unit-test/build/ coverage
-          lcov --rc lcov_branch_coverage=1 --remove test/unit-test/build/coverage.info '*test*' --output-file test/unit-test/build/coverage.info
```
alfred2g, line 38: not resetting the coverage with each run, will give erroneous results when making changes
and it adds coverage with each run.
so if you change something in the code, that results in no longer covering a certain line..it will still show as covered while it is actually not.
coverage.info needs to be zeroed/removed with each run

test/unit-test/ConfigFiles/FreeRTOSIPConfig.h
```diff
@@ -300,11 +297,20 @@ extern uint32_t ulRand();
...
+#define ipconfigETHERNET_MINIMUM_PACKET_BYTES    ( 200 )
 
-void vApplicationMQTTGetKeys( const char ** ppcRootCA,
-                              const char ** ppcClientCert,
-                              const char ** ppcClientPrivateKey );
+#define ipconfigARP_STORES_REMOTE_ADDRESSES      ( 1 )
+
+#define ipconfigARP_USE_CLASH_DETECTION          ( 1 )
+
+#define portINLINE
```
alfred2g, line 314: this should be defined in a port or portmacro.h
It should be part of the port layer for the kernel port in use (hence the 'port' prefix).  Adding it here is really dangerous because it means our demo will build but end user code won't

test/unit-test/ConfigFiles/portmacro.h
```diff
@@ -1,6 +1,6 @@
 /*
- * FreeRTOS Kernel V10.3.0
- * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ * FreeRTOS Kernel V10.4.3
+ * Copyright 2020 Cambridge Consultants Ltd.
```
alfred2g, line 3: mmm i thought it should be the reverse Amazon.com instead of Cambridge Consultants ..

test/unit-test/cmock/create_test.cmake
```diff
@@ -151,9 +152,9 @@ function(create_real_library target
             ${real_include_list}
         )
     set_target_properties(${target} PROPERTIES
-                COMPILE_FLAGS "-Wextra -Wpedantic \
+                COMPILE_FLAGS "-O0 -Wextra -Wpedantic \
                     -fprofile-arcs -ftest-coverage -fprofile-generate \
-                    -Wno-unused-but-set-variable"
+                    -Wno-unused-but-set-variable -ggdb"
```
alfred2g, line 157: these you can add them only with debug builds
in case we want to add some speed tests you want to be with O3
cmake -DCMAKE_BUILD_TYPE=Release/Debug
check: CMAKE_C_FLAGS_[DEBUG|RELEASE|RELWITHDEBINFO|MINSIZEREL|...].

test/unit-test/FreeRTOS_ARP/FreeRTOS_ARP_utest.c
```diff
@@ -0,0 +1,947 @@
...
+
+#define ARPCacheEntryToCheck    2
+
+#if ARPCacheEntryToCheck >= ipconfigARP_CACHE_ENTRIES
+    #error "ARPCacheEntryToCheck cannot be greater than ipconfigARP_CACHE_ENTRIES"
+#endif
+
+extern ARPCacheRow_t xARPCache[ ipconfigARP_CACHE_ENTRIES ];
+
+void FillARPCache( void )
```
alfred2g, line 29: static void

test/unit-test/ConfigFiles/portmacro.h
```diff
@@ -19,146 +19,105 @@
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  *
- * http://www.FreeRTOS.org
- * http://aws.amazon.com/freertos
+ * https://www.FreeRTOS.org
+ * https://github.com/FreeRTOS
  *
  * 1 tab == 4 spaces!
```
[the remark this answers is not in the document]
AniruddhaKanhere: Removed the file

test/unit-test/ConfigFiles/portmacro.h
AniruddhaKanhere: Removed the file

test/unit-test/FreeRTOS_ARP/FreeRTOS_ARP_utest.c
AniruddhaKanhere: This is not required, so removed the function too.

test/unit-test/ConfigFiles/FreeRTOSIPConfig.h
AniruddhaKanhere: Done.

test/unit-test/cmock/create_test.cmake
AniruddhaKanhere: Removed

alfred2g (approved):

test/unit-test/catch_assert.h
```diff
@@ -0,0 +1,77 @@
...
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
+static void catchHandler_( int signal )
+{
+    longjmp( CATCH_JMPBUF, signal );
+}
+#pragma GCC diagnostic pop
+
+#define catch_assert( x )                    \
```
gkwicker, line 54: Why do we need this `setjmp()/longjmp()` magic?  Would it be possible to simplify redefine `assert()` in the unit test to set a variable that the unit test code checks?

gkwicker (commented):
One question, see comment.

test/unit-test/catch_assert.h
AniruddhaKanhere: Thanks for the review.
In some cases, setting a variable is not enough. The code is supposed to abort in such situations. Such as:
```
configASSERT( MyPtr != NULL );
MyPtr->member = 2;
```
In the above case, the `setjmp` magic would work. But if we had just set a variable, then this will cause a segmentation fault (since MyPtr is NULL).

gkwicker: makes sense, thank you

gkwicker (approved):

PR #215, opened 2021-03-23
Repair buffer leak when replying to NBNS requests

htibosch:
Description
-----------
As reported by [niconrok](https://forums.freertos.org/u/niconrok) in [this post](https://forums.freertos.org/t/memory-leak-during-nbns-llmnr-hostname-resolution/12062) the following:

when the application gives a positive reply to an NBNS requests, and only in case `BufferAllocation_2.c` is used, a network buffer will not be released.

The patch goes like this:
~~~c
 usLength = ( uint16_t ) ( sizeof( NBNSAnswer_t ) + ( size_t ) offsetof( NBNSRequest_t, usType ) );

 prvReplyDNSMessage( pxNetworkBuffer, ( BaseType_t ) usLength );

+if( pxNewBuffer != NULL )
+{
+    vReleaseNetworkBufferAndDescriptor( pxNewBuffer );
+}
~~~

Test Steps
-----------
Enabled NBNS and disable LLMNR:

    #define ipconfigUSE_NBNS   1
    #define ipconfigUSE_LLMNR  0

Write a version of `xApplicationDNSQueryHook()` that always returns `pdPASS`. In that case, the application will reply to all NBNS requests.
Issue some logging about the free network buffers ( `uxGetNumberOfFreeNetworkBuffers()` ).

Note that the LLMNR driver did not have this problem.

Related Issue
-----------
Both protocols can be very useful on a LAN: it allows to work with device names in stead of IP-addresses.
However, both protocols are not considered safe any more. There will be re better replacement: mDNS.


AniruddhaKanhere (approved):

FreeRTOS_DNS.c
```diff
@@ -1871,6 +1871,11 @@
                         usLength = ( uint16_t ) ( sizeof( NBNSAnswer_t ) + ( size_t ) offsetof( NBNSRequest_t, usType ) );
 
                         prvReplyDNSMessage( pxNetworkBuffer, ( BaseType_t ) usLength );
+
+                        if( pxNewBuffer != NULL )
```
muneebahmed10, line 1875: Reading the most recent [post](https://forums.freertos.org/t/memory-leak-during-nbns-llmnr-hostname-resolution/12062/7) it looks like `pxNewBuffer` is out of scope by this time? Line 1823 would need to be moved up.

AniruddhaKanhere: I agree! Missed that completely.
@htibosch, would you mind making the requested changes?

AniruddhaKanhere (changes requested):
See the comments.

FreeRTOS_DNS.c
htibosch: Sorry yes, you are right. I moved the declaration of `pxNewBuffer` upward and initialised it to NULL. Thanks

htibosch:
About my commit "Variable 'pxNewBuffer' was out of scope" :

I moved the declaration of `pxNewBuffer` upward so that it can be released.

AniruddhaKanhere (approved):
@htibosch, thank you for the changes.

cobusve (approved):

PR #229, opened 2021-04-21
Move FreeRTOS_closesocket() out of vTCPStateChange

bernd-edlinger:
Description
-----------
This is a possible fix for issues reported here https://forums.freertos.org/t/issues-with-freertos-closesocket/12195

The idea is to move the call to `FreeRTOS_closesocket()` out of `vTCPStateChange()`.
use vCloseSocket instead, which is simple for vTCPStateChange (eCLOSED),
but not for vTCPStateChange(eCLOSE_WAIT) since these are done from a deeply
nested call.  Instead I propose to signal the user application to "accept" the new socket
and just pretend that the connection dropped a short moment later, therefore the socket
will be closed by the user.  These errors happen rarely so it is not a significant overhead.

* Use vCloseSocket() instead of FreeRTOS_closesocket(),

* Rework the commit 86ee487a so that a state change
from eSYN_RECEIVED -> eCLOSE_WAIT makes the select
wake up and the user task can accept and close the
socket.

* Fix build with ipconfigTCP_HANG_PROTECTION == 0.

The problem is that a select will not wake up when the
child socket is accepted, but only when some data is received.
However if the server wants to send data immediately after
accpet that will not work.

* Also remove the code in prvTCPSocketCopy() which makes the
child socket inherit the socket set of the parent.
This is most likely just a work-around that makes the select
wake up when the first user data is received.

Test Steps
-----------
test without ipconfigTCP_HANG_PROTECTION:
have a server that sends first some data before receiving data.
that is rarely used, except for instance in "timed".
test with the following test program that sends SYN/SYN+ACK/RST
sequence:
```C
#include <sys/socket.h>
#include <netinet/in.h>
#include <arpa/inet.h>
#include <string.h>
#include <unistd.h>
#include <fcntl.h>
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
int main(int argc, char **argv)
{
  int i,x;
  struct sockaddr_in s;
  struct linger li = {1,0};
  if (argc != 3)
  {
    printf("args: ip port\n");
    return 1;
  }
  i = socket(AF_INET, SOCK_STREAM, 0);
  if (i<0)
  {
    printf("err\n");
    return 1;
  }
  fcntl(i, F_SETFL, fcntl(i, F_GETFL) | O_NONBLOCK);
  x=setsockopt(i,SOL_SOCKET, SO_LINGER, &li, sizeof(li));
  if(x!= 0)
  {
    printf("setsockopt->%d\n",x);
  }

  memset(&s,0, sizeof(s));
  s.sin_family = AF_INET;
  s.sin_port = htons(atoi(argv[2]));
  s.sin_addr.s_addr = inet_addr(argv[1]);
  x=connect(i, (struct sockaddr*)&s, sizeof(s));
  if(x!= -1 && errno != EINPROGRESS)
  {
    printf("connect->%d\n",x);
  }

  x=close(i);
  if(x!= 0)
  {
    printf("close->%d\n",x);
  }

}
```

this does however not test the SYN/SYN-ACK/X sequence,
I used a "hack" to test it, when the RST is received, 
```C
                {
                    /* This is not a socket in listening mode. Check for the RST
                     * flag. */
                    ucTCPFlags &= ~tcpTCP_FLAG_RST;
                    if( ( ucTCPFlags & tcpTCP_FLAG_RST ) != 0U )
                    {
```

### DIFF
--- a/FreeRTOS_TCP_IP.c
+++ b/FreeRTOS_TCP_IP.c
@@ -499,10 +499,19 @@
                      * gets connected. */
                     if( pxSocket->u.xTCP.bits.bPassQueued != pdFALSE_UNSIGNED )
                     {
-                        /* vTCPStateChange() has called FreeRTOS_closesocket()
+                        if( pxSocket->u.xTCP.bits.bReuseSocket == pdFALSE_UNSIGNED )
+                        {
+                            /* As it did not get connected, and the user can never
+                             * accept() it anymore, it will be deleted now.  Called from
+                             * the IP-task, so it's safe to call the internal Close
+                             * function: vSocketClose(). */
+                            vSocketClose( pxSocket );
+                        }
+
+                        /* prvTCPStatusAgeCheck() has called vSocketClose()
                          * in case the socket is not yet owned by the application.
                          * Return a negative value to inform the caller that
-                         * the socket will be closed in the next cycle. */
+                         * the socket got closed and may not be accessed anymore. */
                         xResult = -1;
                     }
                 }
@@ -1720,6 +1729,17 @@
             FreeRTOS_Socket_t * xConnected = NULL;
         #endif
 
+        /* If a socket is orphaned and does a state change to eCLOSE_WAIT,
+         * we cannot simply call vSocketClose here, as the caller is not yet done
+         * with the socket in this state. This makes sure select/accept gets
+         * signalled, so the user can accept and close the socket normally. */
+        if( ( eTCPState == eCLOSE_WAIT ) && ( bBefore == pdFALSE ) &&
+            ( pxSocket->u.xTCP.bits.bPassQueued != pdFALSE_UNSIGNED ) &&
+            ( pxSocket->u.xTCP.bits.bReuseSocket == pdFALSE_UNSIGNED ) )
+        {
+            bAfter = pdTRUE;
+        }
+
         /* Has the connected status changed? */
         if( bBefore != bAfter )
         {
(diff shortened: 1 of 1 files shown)


htibosch:
Hello Bernd, we studied your proposed changes, and although I have a long experience with this library, I can not be convinced that we won't break anything by applying you PR. It will be very hard to test.

Let me go back to the original issue: the IP-task queue `xNetworkEventQueue` only has space for `ipconfigEVENT_QUEUE_LENGTH` messages. When more messages are sent, the function `xQueueSendToBack()` will either block and/or fail.

I do agree that the application should block (`portMAX_DELAY`) when sending a message.
And we also agree that it is useless to let the IP-task block while waiting for itself.

We already had this ( unrelated ) escape mechanism:
~~~c
    /* Simply send the network task the appropriate event. */
    if( xSendEventStructToIPTask( &xNetworkDownEvent, xDontBlock ) != pdPASS )
    {
        /* Could not send the message, so it is still pending. */
        xNetworkDownEventPending = pdTRUE;
    }
~~~
It says: when the queue is full, remember what has to be done.

I would like to do the same with orphaned sockets that must be closed.

In stead of calling `FreeRTOS_closesocket()`, I would like to call a new function `vMarkSocketForClosure()`, and have it closed in the next loop of the IP-task.
The reason for this choice is that it will not change anything about the logic and flow of the functions.

FreeRTOS_TCP_IP.c
```diff
@@ -3841,48 +3854,14 @@
...
-                if( pxSocket->u.xTCP.pxPeerSocket == NULL )
-                {
-                    pxSocket->u.xTCP.pxPeerSocket = pxNewSocket;
-                }
-            }
-        #endif /* if ( ipconfigTCP_HANG_PROTECTION == 1 ) */
+        /* When bPassQueued is true, the socket is an orphan until it gets
+         * connected. */
+        pxNewSocket->u.xTCP.bits.bPassQueued = pdTRUE_UNSIGNED;
+        pxNewSocket->u.xTCP.pxPeerSocket = pxSocket;
```
bernd-edlinger, line 3864: Okay, but maybe this remains as in independent issue.

htibosch: @bernd-edlinger wrote:

> Also remove the code in `prvTCPSocketCopy()` which makes the
> child socket inherit the socket set of the parent.
> This is most likely just a work-around that makes the select
> wake up when the first user data is received.

I am sure you have explained it earlier, but what was the reason of these changes?
Did you observe a race between the IP-task and the application using select?

bernd-edlinger: From the beginning I was wondering where/how the select is supposed to wake up when
`bPassAccept` is set to true here (the case w/o hang protection).
I see the only code path which does this in `vTCPStateChange()`
https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/d425fbd70cb9459f4b9f722668be8cf6de42f75d/FreeRTOS_TCP_IP.c#L1754-L1759
which is guarded by https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/d425fbd70cb9459f4b9f722668be8cf6de42f75d/FreeRTOS_TCP_IP.c#L1730
My other concern, is, when a socket that is not yet in the eESTABLISHED is allowed
to be returned from `FreeRTOS_accept()` would it work as expected or would it not.
So I tried to setup a test for this, I have a server which uses select to wait for connections,
and usually the protocol requires the client to send something first.
So I was surprised that the select did only wake up, after the first bytes arrived,
but if I change the protocol make the server send a "hello" first, the protocol gets stuck.

bernd-edlinger:
This is now replaced by #238 and #239

PR #231, opened 2021-04-25
Delete the SocketSet_t in the IP task

bernd-edlinger:
Description
-----------
This prevents FreeRTOS_DeleteSocketSet to delete the
SocketSet, while a eSocketCloseEvent is still
queued and the Socket_t refers to a deleted
SocketSet_t for a short time.

Related Issue
-----------

https://forums.freertos.org/t/issues-with-freertos-closesocket/12195

### DIFF
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -590,6 +590,18 @@ static void prvIPTask( void * pvParameters )
                 #endif /* ipconfigUSE_TCP */
                 break;
 
+            case eSocketSetDeleteEvent:
+                #if ( ipconfigSUPPORT_SELECT_FUNCTION == 1 )
+                    {
+                        SocketSelect_t * pxSocketSet = ( SocketSelect_t * ) ( xReceivedEvent.pvData );
+
+                        iptraceMEM_STATS_DELETE( pxSocketSet );
+                        vEventGroupDelete( pxSocketSet->xSelectGroup );
+                        vPortFree( ( void * ) pxSocketSet );
+                    }
+                #endif /* ipconfigSUPPORT_SELECT_FUNCTION == 1 */
+                break;
+
             case eNoEvent:
                 /* xQueueReceive() returned because of a normal time-out. */
                 break;
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -550,13 +550,16 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
  */
     void FreeRTOS_DeleteSocketSet( SocketSet_t xSocketSet )
     {
-        SocketSelect_t * pxSocketSet = ( SocketSelect_t * ) xSocketSet;
+        IPStackEvent_t xCloseEvent;
 
 
-        iptraceMEM_STATS_DELETE( pxSocketSet );
+        xCloseEvent.eEventType = eSocketSetDeleteEvent;
+        xCloseEvent.pvData = ( void * ) xSocketSet;
 
-        vEventGroupDelete( pxSocketSet->xSelectGroup );
-        vPortFree( pxSocketSet );
+        if( xSendEventStructToIPTask( &xCloseEvent, ( TickType_t ) portMAX_DELAY ) == pdFAIL )
+        {
+            FreeRTOS_printf( ( "FreeRTOS_DeleteSocketSet: xSendEventStructToIPTask failed\n" ) );
+        }
     }
 
 #endif /* ipconfigSUPPORT_SELECT_FUNCTION == 1 */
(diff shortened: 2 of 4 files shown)


FreeRTOS_IP.c
```diff
@@ -590,6 +590,18 @@ static void prvIPTask( void * pvParameters )
                 #endif /* ipconfigUSE_TCP */
                 break;
 
+            case eSocketSetDeleteEvent:
+                #if ( ipconfigSUPPORT_SELECT_FUNCTION == 1 )
+                    {
+                        SocketSelect_t * pxSocketSet = ( SocketSelect_t * ) ( xReceivedEvent.pvData );
+
+                        iptraceMEM_STATS_DELETE( pxSocketSet );
+                        vEventGroupDelete( pxSocketSet->xSelectGroup );
```
AniruddhaKanhere, line 599: @bernd-edlinger, thank you for this PR and this change.
This change is certainly an improvement over the previous code, but there can still be a case where this would be an issue.
- If two tasks try to delete a socket set at once, they shall both queue the deletion events to the IP task.
- One event would be processed before the other one (since there is no actual parallelism - just an illusion).
- I am not sure of the impact of calling `vEventGroupDelete()` on a socket set which has already been deleted.

Let me know what you think about this.
Also @htibosch, what do you think about my concern.

Thanks!

AniruddhaKanhere (commented):
Just one concern. See comments.

FreeRTOS_IP.c
bernd-edlinger: That would be undefined behaviour already, since one of the two tasks would delete already deleted memory.

bernd-edlinger: see the vPortFree below for instance.

AniruddhaKanhere: Let me take that back, Actually, in this case, the application should take care of this. I see the err in my argument there.
The socket would be a shared object and the application should take care to use mutex when accessing the shared object.

AniruddhaKanhere (approved):

FreeRTOS_Sockets.c
```diff
@@ -550,13 +550,16 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
...
+        IPStackEvent_t xCloseEvent;
 
 
-        iptraceMEM_STATS_DELETE( pxSocketSet );
+        xCloseEvent.eEventType = eSocketSetDeleteEvent;
+        xCloseEvent.pvData = ( void * ) xSocketSet;
 
-        vEventGroupDelete( pxSocketSet->xSelectGroup );
-        vPortFree( pxSocketSet );
+        if( xSendEventStructToIPTask( &xCloseEvent, ( TickType_t ) portMAX_DELAY ) == pdFAIL )
```
htibosch, line 559: Yes, it is good to let the IP-task delete the socket-set.

htibosch (approved):
I give my approval of the proposed changes.
Thanks, Hein

FreeRTOS_IP.c
htibosch: Indeed, just like sockets and memory blocks can only be closed/released once. A second attempt will indeed lead to undefined behaviour.

Maybe it is worth noting that the application is responsible for detaching existing sockets from the select group by calling `FreeRTOS_FD_CLR( socket, set, eSELECT_ALL )`.

bernd-edlinger: Hmm, yes,

that would be cleaner, but `FreeRTOS_FD_CLR()` does not synchronize with the IP-TASK which may
still be using the socket set *after* `FreeRTOS_FD_CLR` returned.

cobusve (approved):

FreeRTOS_IP.c
htibosch: It is hard for me to imagine a situation in which you want to delete a socket group, **while** there are still active sockets. It sounds like you're closing the shop while there are still customers inside.

What do you want to do with the open client sockets? Does the server socket continue to exists?

Or do you want to stop a TCP service bound to a certain port number?

Is that a problem?

In order to stop a TCP service I would:

- Detach all sockets from the socket group, which happens immediately.
- Close the socket group ( which is an API since PR #231 ).
- Close the server socket.
- Close the clients sockets in a graceful way.

There is a small chance of receiving a new client during this process. Until now that orphaned socket would get cleared after 30 seconds when `ipconfigTCP_HANG_PROTECTION` is defined. But yes, we could as well close any unattended socket while deleting the server socket.

bernd-edlinger: No the test case which fails, is when the listening socket is closed, and the socket set is deleted
after FreeRTOS_closesocket returns.  Now the to be deleted socket receives quickly a new connection
and traps while accessing the already deleted socket set.
I admit I have not tried to detach, the socket set before the close.
But I know the socket group pointer is copied into orphans, and is therefore inaccesible ATM.
#239 would (if accepted) remove that copy.
Anyway simply changing a pointer value is not an atomic operation.  We will have to fix that too.

PR #232, opened 2021-04-25
When a listening socket is closed, close also all orphans

bernd-edlinger:
Description
-----------

Close all not yet accepted sockets when the listening socket is closed.
Otherwise these sockets become orphans and cannot be accepted any more.
Already accepted sockets must be closed by the application.

Test Steps
-----------
Close a listening socket when there are new connections which are not accepted.
See it is impossible to bind a new server socket on the same port again, since the
orphans occupy the port.

Related Issue
-----------

https://forums.freertos.org/t/issues-with-freertos-closesocket/12195

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -1597,8 +1597,8 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
 
 /**
  * @brief When a child socket gets closed, make sure to update the child-count of the
- *        parent. When a listening parent socket is closed, make sure no child-sockets
- *        keep a pointer to it.
+ *        parent. When a listening parent socket is closed, make sure to close also
+ *        all orphaned child-sockets.
  *
  * @param[in] pxSocketToDelete: The socket being closed.
  */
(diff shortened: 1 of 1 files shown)


FreeRTOS_Sockets.c
```diff
@@ -1599,6 +1599,25 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+
+                if( ( pxOtherSocket->u.xTCP.ucTCPState != ( uint8_t ) eTCP_LISTEN ) &&
+                    ( pxOtherSocket->usLocalPort == usLocalPort ) &&
+                    ( pxOtherSocket->u.xTCP.bits.bPassQueued || pxOtherSocket->u.xTCP.bits.bPassAccept ) )
+                {
+                    vSocketClose( pxOtherSocket );
+                }
+            }
+
+            return;
```
htibosch, line 1618: MISRA doesn't like a `return` in the middle of a function, so could you change it into an else?

htibosch (approved):
Until now, orphaned sockets would only be removed after a time-out, by the IP-task when `ipconfigTCP_HANG_PROTECTION` is defined.

After this PR, when the listening socket is closed, any pending sockets ( in the SYN phase ) will get closed immediately.

Thanks, Hein

FreeRTOS_Sockets.c
```diff
@@ -1599,23 +1599,42 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
-             pxIterator != pxEnd;
-             pxIterator = listGET_NEXT( pxIterator ) )
+        if( pxSocketToDelete->u.xTCP.ucTCPState == ( uint8_t ) eTCP_LISTEN )
         {
-            pxOtherSocket = ipCAST_PTR_TO_TYPE_PTR( FreeRTOS_Socket_t, listGET_LIST_ITEM_OWNER( pxIterator ) );
+            for( pxIterator = listGET_NEXT( pxEnd );
+                 pxIterator != pxEnd; )
+            {
+                pxOtherSocket = ipCAST_PTR_TO_TYPE_PTR( FreeRTOS_Socket_t, listGET_LIST_ITEM_OWNER( pxIterator ) );
+                pxIterator = listGET_NEXT( pxIterator );
```
AniruddhaKanhere, line 1608: This is a very minor suggestion to keep both the if and else clauses similar to each other.  Why not move this up in the for loop condition above?
`pxIterator = listGET_NEXT( pxIterator );`

i.e., the loop would become:
```
            for( pxIterator = listGET_NEXT( pxEnd );
                 pxIterator != pxEnd; 
                 pxIterator = listGET_NEXT( pxIterator )    )
```

AniruddhaKanhere: I don't mind this how it is. I shall approve the PR.

AniruddhaKanhere (approved):

FreeRTOS_Sockets.c
bernd-edlinger: Hmm, I think that would not work, because of the 
vSocketClose below I need to do the listGET_NEXT
here, the original form of the loop would access free
memory.

However, I agree that the reason why I wrote it this way
is probably not so clear as I thought it would be, so I should
definitely add a comment here, to prevent such "code-cleanup"
in the future.

bernd-edlinger: I've added a comment to hopefully make the code more clear:
"This needs to be done here, before calling vSocketClose."
Thanks for bringing this to may attention.

AniruddhaKanhere (approved):

FreeRTOS_Sockets.c
AniruddhaKanhere: Ah! Yes, I did not see the cast. Thanks for explaining that.
I have approved the PR. But it needs to be rebased before it can be merged.

AniruddhaKanhere:
@bernd-edlinger, please allow FreeRTOS members to contribute to your repository. That would allow us to rebase your PR on top of main. This is required since we only allow linear history of commits.
This is outlined in the Contributing document [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/.github/CONTRIBUTING.md#contributing-via-pull-request). Paraphrasing: "*NOTE: Please make sure the default option (Allow edits from maintainers) is left checked.*"

FreeRTOS_Sockets.c
```diff
@@ -1599,23 +1599,44 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
 
-            if( ( pxOtherSocket->u.xTCP.ucTCPState == ( uint8_t ) eTCP_LISTEN ) &&
-                ( pxOtherSocket->usLocalPort == usLocalPort ) &&
-                ( pxOtherSocket->u.xTCP.usChildCount != 0U ) )
+                /* This needs to be done here, before calling vSocketClose. */
+                pxIterator = listGET_NEXT( pxIterator );
+
+                if( ( pxOtherSocket->u.xTCP.ucTCPState != ( uint8_t ) eTCP_LISTEN ) &&
+                    ( pxOtherSocket->usLocalPort == usLocalPort ) &&
+                    ( pxOtherSocket->u.xTCP.bits.bPassQueued || pxOtherSocket->u.xTCP.bits.bPassAccept ) )
```
htibosch, line 1614: For me the expression is perfect, but MISRA would like to see:
~~~c
( ( pxOtherSocket->u.xTCP.bits.bPassQueued != 0U ) ||
  ( pxOtherSocket->u.xTCP.bits.bPassAccept != 0U ) )
~~~

bernd-edlinger:
Done.

FreeRTOS_Sockets.c
bernd-edlinger: okay, done.

bernd-edlinger:
rebased to master and enabled the allow edits checkbox

AniruddhaKanhere (approved):

n9wxu (approved):

AniruddhaKanhere:
@bernd-edlinger, thanks a lot for this PR and for your patience! We are working on (reviewing and) merging your other PRs as well. 
Merging the PR now.

PR #236, opened 2021-05-02
IPv6/Multi: Reducing the complexity of FreeRTOS_Sockets.c

htibosch:
Description
-----------
In this PR, the following long API calls are split-up into smaller functions: `FreeRTOS_recv()`, `FreeRTOS_send()`, and `FreeRTOS_setsockopt()`.

In total 11 new functions have been added. These are all declared static, so it is expected that an optimising compile will inline them fully.

The complexity score before this PR:
~~~
31     180     110   FreeRTOS_Sockets.c(3763): FreeRTOS_send
34     182     144   FreeRTOS_Sockets.c(3448): FreeRTOS_recv
46     336     253   FreeRTOS_Sockets.c(1745): FreeRTOS_setsockopt
~~~
The score after applying the PR:
~~~
 4      44      32   FreeRTOS_Sockets.c(3854): FreeRTOS_recv
 9      45      33   FreeRTOS_Sockets.c(4047): FreeRTOS_send
14     122      78   FreeRTOS_Sockets.c(2060): FreeRTOS_setsockopt
~~~

After this PR, there is still some more work to do on the complexity. I like to do that in a second PR for FreeRTOS_Sockets.c.

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -91,10 +91,18 @@
     #define ipTCP_TIMER_PERIOD_MS    ( 1000U )
 #endif
 
-/* Some helper macro's for defining the 20/80 % limits of uxLittleSpace / uxEnoughSpace. */
-#define sock20_PERCENT     20U  /**< 20% of the defined limit. */
-#define sock80_PERCENT     80U  /**< 80% of the defined limit. */
-#define sock100_PERCENT    100U /**< 100% of the defined limit. */
+/** @brief Some helper macro's for defining the 20/80 % limits of uxLittleSpace / uxEnoughSpace. */
+#define sock20_PERCENT            20U  /**< 20% of the defined limit. */
+#define sock80_PERCENT            80U  /**< 80% of the defined limit. */
+#define sock100_PERCENT           100U /**< 100% of the defined limit. */
+
+/** @brief When ucASCIIToHex() can not convert a character,
+ *         the value 255 will be returned.
+ */
+#define socketINVALID_HEX_CHAR    0xffU
+
+/** @brief Convert a readable character to its hex value. */
+static uint8_t ucASCIIToHex( char cChar );
 
 #if ( ipconfigUSE_CALLBACKS != 0 )
     static portINLINE ipDECL_CAST_PTR_FUNC_FOR_TYPE( F_TCP_UDP_Handler_t )
@@ -130,39 +138,34 @@ static portINLINE ipDECL_CAST_PTR_FUNC_FOR_TYPE( StreamBuffer_t )
 #endif
 /*-----------------------------------------------------------*/
 
-/*
- * Allocate the next port number from the private allocation range.
+/** @brief Allocate the next port number from the private allocation range.
  * TCP and UDP each have their own series of port numbers
- * ulProtocol is either ipPROTOCOL_UDP or ipPROTOCOL_TCP
+ *         ulProtocol is either ipPROTOCOL_UDP or ipPROTOCOL_TCP.
  */
 static uint16_t prvGetPrivatePortNumber( BaseType_t xProtocol );
 
-/*
- * Return the list item from within pxList that has an item value of
+/** @brief Return the list item from within pxList that has an item value of
  * xWantedItemValue.  If there is no such list item return NULL.
  */
 static const ListItem_t * pxListFindListItemWithValue( const List_t * pxList,
                                                        TickType_t xWantedItemValue );
 
-/*
- * Return pdTRUE only if pxSocket is valid and bound, as far as can be
- * determined.
+/** @brief Return pdTRUE only if pxSocket is valid and bound, as far as
+ *         can be determined.
  */
 static BaseType_t prvValidSocket( const FreeRTOS_Socket_t * pxSocket,
                                   BaseType_t xProtocol,
                                   BaseType_t xIsBound );
 
-/*
- * Internal function prvSockopt_so_buffer(): sets FREERTOS_SO_SNDBUF or
- * FREERTOS_SO_RCVBUF properties of a socket.
+/** @brief Internal function prvSockopt_so_buffer(): sets FREERTOS_SO_SNDBUF
+ *         or FREERTOS_SO_RCVBUF properties of a socket.
  */
 static BaseType_t prvSockopt_so_buffer( FreeRTOS_Socket_t * pxSocket,
                                         int32_t lOptionName,
                                         const void * pvOptionValue );
 
-/*
- * Before creating a socket, check the validity of the parameters used
- * and find the size of the socket space, which is different for UDP and TCP
+/** @brief Before creating a socket, check the validity of the parameters used
+ *         and find the size of the socket space, which is different for UDP and TCP.
  */
 static BaseType_t prvDetermineSocketSize( BaseType_t xDomain,
                                           BaseType_t xType,
@@ -171,8 +174,8 @@ static BaseType_t prvDetermineSocketSize( BaseType_t xDomain,
 
 #if ( ipconfigUSE_TCP == 1 )
 
-/*
- * Create a txStream or a rxStream, depending on the parameter 'xIsInputStream'
+/** @brief Create a txStream or a rxStream, depending on the parameter
+ *         'xIsInputStream'.
  */
     static StreamBuffer_t * prvTCPCreateStream( FreeRTOS_Socket_t * pxSocket,
                                                 BaseType_t xIsInputStream );
@@ -180,9 +183,8 @@ static BaseType_t prvDetermineSocketSize( BaseType_t xDomain,
 
 #if ( ipconfigUSE_TCP == 1 )
 
-/*
- * Called from FreeRTOS_send(): some checks which will be done before
- * sending a TCP packed.
+/** @brief Called from FreeRTOS_send(): some checks which will be done
+ *         before sending a TCP packed.
  */
     static int32_t prvTCPSendCheck( FreeRTOS_Socket_t * pxSocket,
                                     size_t uxDataLength );
@@ -190,17 +192,17 @@ static BaseType_t prvDetermineSocketSize( BaseType_t xDomain,
 
 #if ( ipconfigUSE_TCP == 1 )
 
-/*
- * When a child socket gets closed, make sure to update the child-count of the parent
+/** @brief When a child socket gets closed, make sure to update the
+ *         child-count of the parent.
  */
     static void prvTCPSetSocketCount( FreeRTOS_Socket_t const * pxSocketToDelete );
 #endif /* ipconfigUSE_TCP == 1 */
 
 #if ( ipconfigUSE_TCP == 1 )
 
-/*
- * Called from FreeRTOS_connect(): make some checks and if allowed, send a
- * message to the IP-task to start connecting to a remote socket
+/** @brief Called from FreeRTOS_connect(): make some checks and if
+ *         allowed, send a message to the IP-task to start connecting
+ *         to a remote socket.
  */
     static BaseType_t prvTCPConnectStart( FreeRTOS_Socket_t * pxSocket,
                                           struct freertos_sockaddr const * pxAddress );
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (approved):

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
-                /* Receive time out. */
-                pxSocket->xReceiveBlockTime = *( ( const TickType_t * ) pvOptionValue );
-                xReturn = 0;
-                break;
+        BaseType_t xReturn = 0;
 
-            case FREERTOS_SO_SNDTIMEO:
-                pxSocket->xSendBlockTime = *( ( const TickType_t * ) pvOptionValue );
+        #if ( ipconfigUSE_TCP == 1 )
+            {
```
paulbartell, line 2005: These scope brackets are not strictly necessary.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+        #endif /* ipconfigUSE_TCP */
 
-                #if ( ipconfigUDP_MAX_RX_PACKETS > 0U )
-                    case FREERTOS_SO_UDP_MAX_RX_PACKETS:
+        if( xReturn == 0 )
+        {
+            switch( lOptionName )
+            {
+                #if ipconfigUSE_TCP == 1
+                    case FREERTOS_SO_TCP_CONN_HANDLER:
```
paulbartell, line 2035: This extra indentation is a bit confusing.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
 
-                            /* Each socket can have a callback function that is executed
-                             * when there is an event the socket's owner might want to
-                             * process. */
-                            /* The type cast of the pointer expression "A" to type "B" removes const qualifier from the pointed to type. */
-                            pxSocket->pxUserWakeCallback = ipCAST_CONST_PTR_TO_CONST_TYPE_PTR( SocketWakeupCallback_t, pvOptionValue );
-                            xReturn = 0;
-                            break;
-                    #endif /* ipconfigSOCKET_HAS_USER_WAKE_CALLBACK */
+            if( xReturn != 0 )
```
paulbartell, line 2099: This violates MISRA Rule 15.4 which advises against multiple break statements inside a loop.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+            }
 
-                           if( ( pxLowHighWater->uxLittleSpace >= pxLowHighWater->uxEnoughSpace ) ||
-                               ( pxLowHighWater->uxEnoughSpace > pxSocket->u.xTCP.uxRxStreamSize ) )
-                           {
-                               /* Impossible values. */
-                               FreeRTOS_debug_printf( ( "FREERTOS_SO_SET_LOW_HIGH_WATER: bad values\n" ) );
-                               break; /* will return -pdFREERTOS_ERRNO_EINVAL */
-                           }
+            #if ( ipconfigUSE_TCP_WIN == 1 )
```
paulbartell, line 2111: A local pointer to pxSocket->u.xTCP would increase clarity here.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
-                           {
-                               FreeRTOS_debug_printf( ( "Set SO_WIN_PROP: buffer already created\n" ) );
-                               break; /* will return -pdFREERTOS_ERRNO_EINVAL */
-                           }
+        if( pxSocket->ucProtocol != ( uint8_t ) FREERTOS_IPPROTO_TCP )
+        {
+            /* It is not allowed to access 'pxSocket->u.xTCP'. */
+            FreeRTOS_debug_printf( ( "FREERTOS_SO_SET_LOW_HIGH_WATER: wrong socket type\n" ) );
+        }
+        else if( ( pxLowHighWater->uxLittleSpace >= pxLowHighWater->uxEnoughSpace ) ||
```
paulbartell, line 2159: Could this be handled with an assert?

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+#if ( ipconfigUSE_TCP != 0 )
 
-                           if( xReturn != 0 )
-                           {
-                               break; /* will return an error. */
-                           }
+/**
+ * @brief Handle the socket option FREERTOS_SO_SET_FULL_SIZE.
+ *        When enabled, the IP-stack will only send packets when
+ *        there are at least MSS bytes to sent.
```
paulbartell, line 2184: s/sent/send/

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+    {
+        BaseType_t xReturn = -pdFREERTOS_ERRNO_EINVAL;
+        const WinProperties_t * pxProps;
 
-                        xReturn = 0;
-                        break;
-                #endif /* ipconfigUSE_CALLBACKS */
+        do
+        {
+            if( pxSocket->ucProtocol != ( uint8_t ) FREERTOS_IPPROTO_TCP )
```
paulbartell, line 2083: MISRA Rule 15.4: "There should be no more than one break or goto statement used to terminate any iteration statement"

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
 
-                           if( ( pxLowHighWater->uxLittleSpace >= pxLowHighWater->uxEnoughSpace ) ||
-                               ( pxLowHighWater->uxEnoughSpace > pxSocket->u.xTCP.uxRxStreamSize ) )
-                           {
-                               /* Impossible values. */
-                               FreeRTOS_debug_printf( ( "FREERTOS_SO_SET_LOW_HIGH_WATER: bad values\n" ) );
-                               break; /* will return -pdFREERTOS_ERRNO_EINVAL */
-                           }
+            #if ( ipconfigUSE_TCP_WIN == 1 )
+                {
```
paulbartell, line 2112: Extra scope brackets are not necessary.

FreeRTOS_Sockets.c
paulbartell: Nevermind. I realize that this is part of the kernel style.

FreeRTOS_Sockets.c
paulbartell: This falls under an exception documented in MISRA.md.

FreeRTOS_Sockets.c
paulbartell: This falls under an exception documented in MISRA.md.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+    static BaseType_t prvSetOptionStopRX( FreeRTOS_Socket_t * pxSocket,
+                                          const void * pvOptionValue )
+    {
+        BaseType_t xReturn = -pdFREERTOS_ERRNO_EINVAL;
+
+        if( pxSocket->ucProtocol == ( uint8_t ) FREERTOS_IPPROTO_TCP )
+        {
+            if( *( ( const BaseType_t * ) pvOptionValue ) != 0 )
+            {
+                pxSocket->u.xTCP.bits.bRxStopped = pdTRUE_UNSIGNED;
```
paulbartell, line 2241: A local pointer to "pxSocket->u.xTCP" could improve clarity.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+
+/**
+ * @brief Handle the socket options FREERTOS_SO_RCVTIMEO and
+ *        FREERTOS_SO_SNDTIMEO.
+ *        Used in applications with streaming audio: tell the peer
+ *        to stop or continue sending data.
+ *
+ * @param[in] pxSocket: The TCP socket used for the connection.
+ * @param[in] pvOptionValue: The option name like FREERTOS_SO_xxx_HANDLER.
+ * @param[in] xForSend: when true, handle 'FREERTOS_SO_SNDTIMEO',
```
paulbartell, line 2266: I would prefer if the FREERTOS_SO_RCVTIMEO case was handled in a separate function or directly in FreeRTOS_setsockopt.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+        }
+
+        pxSocket->xSendBlockTime = xBlockTime;
+    }
+    else
+    {
+        pxSocket->xReceiveBlockTime = xBlockTime;
+    }
+
+    return 0;
```
paulbartell, line 2302: The return type from this function is constant.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+
+    if( xForSend == pdTRUE )
+    {
+        if( pxSocket->ucProtocol == ( uint8_t ) FREERTOS_IPPROTO_UDP )
+        {
+            /* The send time out is capped for the reason stated in the
+             * comments where ipconfigUDP_MAX_SEND_BLOCK_TIME_TICKS is defined
+             * in FreeRTOSIPConfig.h (assuming an official configuration file
+             * is being used. */
+            if( xBlockTime > ipconfigUDP_MAX_SEND_BLOCK_TIME_TICKS )
```
paulbartell, line 2283: This could be combined into a single compound if statement.

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+            /* For TCP socket, it isn't necessary to limit the blocking time
+             * because	the FreeRTOS_send() function does not wait for a network
+             * buffer to become available. */
+        }
+
+        pxSocket->xSendBlockTime = xBlockTime;
+    }
+    else
+    {
+        pxSocket->xReceiveBlockTime = xBlockTime;
```
paulbartell, line 2299: Does xReceiveBlockTime need to be bounds checked?

FreeRTOS_Sockets.c
```diff
@@ -1878,371 +1980,539 @@ void * vSocketClose( FreeRTOS_Socket_t * pxSocket )
...
+ * @param[in] pxSocket: The TCP socket used for the connection.
+ * @param[in] pvOptionValue: The option name like FREERTOS_SO_xxx_HANDLER.
+ * @param[in] xForSend: when true, handle 'FREERTOS_SO_SNDTIMEO',
+ *            otherwise handle the option `FREERTOS_SO_RCVTIMEO`.
+ */
+static int32_t prvSetOptionTimeout( FreeRTOS_Socket_t * pxSocket,
+                                    const void * pvOptionValue,
+                                    BaseType_t xForSend )
+{
+    TickType_t xBlockTime = *( ( const TickType_t * ) pvOptionValue );
```
paulbartell, line 2273: It would be cleaner if this cast occurred in FreeRTOS_setsockopt for each of the socket options.

FreeRTOS_Sockets.c
```diff
@@ -171,63 +174,162 @@ static BaseType_t prvDetermineSocketSize( BaseType_t xDomain,
...
+#if ( ipconfigUSE_TCP != 0 )
+
+/**
+ * @brief Handle the socket option FREERTOS_SO_WIN_PROPERTIES.
+ */
+    static BaseType_t prvSetOptionTCPWindows( FreeRTOS_Socket_t * pxSocket,
+                                              const void * pvOptionValue );
+#endif /* ( ipconfigUSE_TCP != 0 ) */
+
+#if ( ipconfigUSE_TCP != 0 )
```
paulbartell, line 282: These indented blocks could be combined.

FreeRTOS_Sockets.c
```diff
@@ -171,63 +174,162 @@ static BaseType_t prvDetermineSocketSize( BaseType_t xDomain,
...
+/** @brief Handle the socket option FREERTOS_SO_STOP_RX. */
+    static BaseType_t prvSetOptionStopRX( FreeRTOS_Socket_t * pxSocket,
+                                          const void * pvOptionValue );
+
+#endif /* ( ipconfigUSE_TCP != 0 ) */
+
+/** @brief Handle the socket options FREERTOS_SO_RCVTIMEO and
+ *         FREERTOS_SO_SNDTIMEO.
+ */
+static int32_t prvSetOptionTimeout( FreeRTOS_Socket_t * pxSocket,
```
paulbartell, line 312: The return type for prvSetOptionTimeout does not match the other socket option handlers.

FreeRTOS_Sockets.c
```diff
@@ -2958,14 +3267,13 @@ uint32_t FreeRTOS_inet_addr( const char * pcIPAddress )
...
 
-
         ( void ) memset( pucTarget, 0, ipSIZE_OF_IPv6_ADDRESS );
         xTargetIndex = 0;
         xColon = -1;
         xResult = 0;
 
         /* Leading :: requires some special handling. */
-        if( ( pcIterator[ 0 ] == ':' ) && ( pcIterator[ 1 ] == ':' ) && ( pcIterator[ 2 ] == '\0' ) )
+        if( memcmp( pcIterator, "::", 3 ) == 0 )
```
paulbartell, line 3276: MISRA Rule 21.14 requires that memcmp should not be used on null terminated strings.

FreeRTOS_Sockets.c
```diff
@@ -2933,6 +3199,49 @@ uint32_t FreeRTOS_inet_addr( const char * pcIPAddress )
...
+    {
+        cValue -= ( char ) 'A';
+        ucNew = ( uint8_t ) cValue;
+        ucNew += ( uint8_t ) 10;
+    }
+    else
+    {
+        /* The character does not represent a valid hex number, return 255.
+         */
+        ucNew = ( uint8_t ) socketINVALID_HEX_CHAR;
```
paulbartell, line 3237: socketINVALID_HEX_CHAR could be cast to a uint8_t when it is defined.

FreeRTOS_Sockets.c
```diff
@@ -3647,155 +3935,214 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
...
-                        {
-                            /* The no-memory error has priority above the non-connected error.
-                             * Both are fatal and will lead to closing the socket. */
-                            xByteCount = -pdFREERTOS_ERRNO_ENOMEM;
-                        }
-                        else
+                if( xRemainingTime == ( TickType_t ) 0 )
+                {
+                    #if ( ipconfigSUPPORT_SIGNALS != 0 )
                         {
```
paulbartell, line 4049: Extra scope brackets.

FreeRTOS_Sockets.c
```diff
@@ -3647,155 +3935,214 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
...
+                xByteCount = -pdFREERTOS_ERRNO_ENOTCONN;
+
+                if( pxSocket->u.xTCP.bits.bMallocError != pdFALSE_UNSIGNED )
+                {
+                    /* The no-memory error has priority above the non-connected error.
+                     * Both are fatal and will lead to closing the socket. */
+                    xByteCount = -pdFREERTOS_ERRNO_ENOMEM;
+                }
+
+                break;
```
paulbartell, line 4038: This usage does not fall under the MISRA Rule 15.4 exception.

FreeRTOS_Sockets.c
```diff
@@ -3647,155 +3935,214 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
...
-                    case eTIME_WAIT:
-                    default:
-                        /* Nothing. */
-                        break;
+                    #endif /* ipconfigSUPPORT_SIGNALS */
+                    break;
                 }
 
-                if( xByteCount < 0 )
+                if( ( ( uint32_t ) xFlags & ( uint32_t ) FREERTOS_MSG_DONTWAIT ) != 0U )
```
paulbartell, line 4058: Perhaps xFlags should be a uint32_t.

FreeRTOS_Sockets.c
htibosch: Do you propose to :
~~~c
IPTCPSocket_t * pxTCP = &( xTCP pxSocket->u.xTCP );
~~~
Yes good, I will do so.

FreeRTOS_Sockets.c
htibosch: As these value are set the user, I would be more forgiving; log the error and return `-EINVAL`.

FreeRTOS_Sockets.c
htibosch: Always confusing to me: send and sent. Changed.

FreeRTOS_Sockets.c
htibosch: This PR was already created 19 days ago and in a meanwhile I worked on both MISRA and the complexity of other functions.
~~~c
-    if( memcmp( pcIterator, "::", 3 ) == 0 )
+    if( strcmp( pcIterator, "::" ) == 0 )
~~~
Let us first finish and merge this PR236, and then I will create a new PR to get all complexity scores below 9.

htibosch (commented):
Thanks again Paul for this review. As I wrote I have a new version of FreeRTOS_Sockets.c in queue. I just wanted to do it in two steps because it is a complex job to make code simpler :-)

FreeRTOS_Sockets.c
htibosch: Here is a question about the "misra-c-2012-rule-15-4" rule on [stackoverflow](https://stackoverflow.com/questions/66369536/misra-c-2012-rule-15-4-and-replacing-gotos-with-breaks).
It says that "But it is just an **advisory** one and there's nothing wrong with using multiple breaks like you do as long as the code is readable and easy to follow".

FreeRTOS_Sockets.c
htibosch: MISRA C 2012 Rule 15.4 seems to be advisory. See my other comments in [PR #236](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/236#discussion_r636754135)

FreeRTOS_Sockets.c
htibosch: Like I said, this is a FreeRTOS habit. Maybe the reason is to limit the scope of any local variable that is introduced between the ifdef's.

FreeRTOS_Sockets.c
htibosch: Have done so.

FreeRTOS_Sockets.c
htibosch: And also: MISRA C 2012 rule 15.4 is advisory only.

FreeRTOS_Sockets.c
htibosch: I have put them together to show the difference in how they're treated.

FreeRTOS_Sockets.c
htibosch: Changed thee return type type void.

FreeRTOS_Sockets.c
htibosch: I would like to leave it this way, just to for the viewer.

FreeRTOS_Sockets.c
htibosch: No, it may be any value between 0 ( non-blocking ) up to `portMAX_DELAY` ( blocking until success ).

FreeRTOS_Sockets.c
htibosch: We have this habit now: declare all macros as either signed or unsigned:
~~~c
    #define NO_ERROR    0
    #define MAX_COUNT   1000U
~~~
Later you can decide what actual cast is needed, if any.

And also, I am not sure that after this declaration:
~~~c
#define VALUE ( ( uint8_t ) 255U )
~~~
you can use `VALUE` in an `#ifdef` expression, like e.g. `#if( VALUE > 512U )`.

FreeRTOS_Sockets.c
htibosch: In the original BSD declarations, `flags` were declared as `int`.
As you know, MISRA doesn't like to do bit-testing in a signed integer.
So therefore the use of a cast.

AniruddhaKanhere (approved):

FreeRTOS_Sockets.c
```diff
@@ -913,6 +978,121 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
...
 
+/**
+ * @brief : called from FreeRTOS_recvfrom(). This function waits for an incoming
+ *          UDP packet, or until a time-out occurs.
+ * @param[in] pxSocket : The socket that receives UDP packets.
+ * @param[in] xFlags : The flags as passed to FreeRTOS_recvfrom().
+ * @param[in,out] pxEventBits : The last even received in this function,
+ *                              either eSOCKET_INTR or eSOCKET_RECEIVE.
+ */
+static NetworkBufferDescriptor_t * prvRecvFromWaitPacket( FreeRTOS_Socket_t const * pxSocket,
```
AniruddhaKanhere, line 989: The name is descriptive but not as clear as I would like I think. (I had to go to the place from where this is called and then I understood what this does :) )
How does the name `prvRecvFromWaitForPacket` sound? I just added a "For" in there.

FreeRTOS_Sockets.c
```diff
@@ -913,6 +978,121 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
...
+            }
+
+            /* To ensure this part only executes once. */
+            xTimed = pdTRUE;
+
+            /* Fetch the current time. */
+            vTaskSetTimeOutState( &xTimeOut );
+        }
+
+        /* Wait for arrival of data.  While waiting, the IP-task may set the
```
AniruddhaKanhere, line 1034:
```suggestion
        /* Wait for arrival of data.  While waiting, the IP-task will set the
```

AniruddhaKanhere (changes requested):

FreeRTOS_Sockets.c
htibosch: Yes I agree, "wait for packet" is clearer than "wait packet",

FreeRTOS_Sockets.c
htibosch: Not sure about this one. "the IP-task will set" sounds like the `eSOCKET_RECEIVE` bit will be set for sure, but a time-out may come earlier.

AniruddhaKanhere (approved):

n9wxu (approved):

PR #237, opened 2021-05-04
Fix the IP header checksum of the ICMP_ECHO_RESPONSE

bernd-edlinger:
Description
-----------

Since the IP header field usFragmentOffset is modified,
update the checksum in the IP header.
While I'm already there, simplify the computation of
the ICMP header checksum as well.

Test Steps
-----------
If ping request are sent with DONT FRAGMENT and
the response is sent without that bit, the IP header checksum is worng.
Linux ignores the ping response.

Ping works with this patch applied.


htibosch:
The addition of `ipconfigFORCE_IP_DONT_FRAGMENT` was done in [February this year](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/179).
You are right, the calculation is incorrect in case the new macro is being used. Thanks for noticing this.

the handling of ICMP is already very old. I wonder if it would be more elegant to just call `usGenerateProtocolChecksum()`?

It would look like e.g. :
~~~c
#if( ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM != 0 )
{
    /* Many EMAC peripherals will only calculate the ICMP checksum
     * correctly if the field is nulled beforehand. */
    pxICMPHeader->usChecksum = 0U;
}
#else
{
    pxICMPHeader->usChecksum = usGenerateProtocolChecksum( pucEthernetBuffer, uxBufferLength, pdTRUE )
}
#endif

~~~
What is you perception of this?

Note that all this only applies when the checksum is not offloaded by the network interface.

bernd-edlinger:
Hmm,

the ICMP header checksum is okay, just the math is awkward.
only the IP header checksum is broken, since the ucFragmenOffset is in the IP header,
that needs adjustment.
OTOH, the `usGenerateProtocolChecksum` does not seem to have a shortcut when the
hardware does the protocol cheksum.

BTW: is the IP and/or protocol checksum checked in software when the hardware does not implement that?

htibosch:
The ICMP handling was already implemented in the UDP-only library, about 10 years ago. Many EMACs didn't have CRC offloading and a full CRC calculation was relatively expensive.

Of course for UDP and TCP, both checksums are calculated in hardware or in software, depending on `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM`.

> OTOH, the `usGenerateProtocolChecksum()` does not seem to have a shortcut when the hardware does the protocol checksum.

How do you mean a shortcut?
When the hardware can calculate it, the above function is normally not called.

> BTW: is the IP and/or protocol checksum checked in software when the hardware does not implement that?

Yes. Normally both checksums will be calculated in software, IP and protocol. ICMP was exceptional: the IP checksum was left untouched, and the protocol checksum got a minor correction.

I vote for a solution in which the ICMP checksums are handled just like UDP and TCP.

bernd-edlinger:
Okay, changed to suggested different approach.

htibosch:
> Okay, changed to suggested different approach.

You're a quick learner! Tonight I will look at it in more detail. Thanks so far.

bernd-edlinger:
Unfortunately there is a problem in the CI check "CBMC Proofs"

```
Loop unwinding failures (loop index)

    [trace] Loop usGenerateChecksum.0 at line 2926 in file FreeRTOS_IP.c
```

I don't see what is wrong in the `usGenerateCheksum` function.
The only thing that is odd in the test harnesh is that
`pxIPPacket` should alias `pxNetworkBuffer->pucEthernetBuffer` but it isn't.
https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/ad8a42929905a1842e0d57bef905e0f8ad7865d3/test/cbmc/proofs/parsing/ProcessIPPacket/ProcessIPPacket_harness.c#L35-L42

FreeRTOS_IP.c
```diff
@@ -2323,22 +2325,17 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
...
-            pxICMPHeader->usChecksum = pxICMPHeader->usChecksum + FreeRTOS_htons( usRequest + 1U );
-        }
-        else
-        {
-            pxICMPHeader->usChecksum = pxICMPHeader->usChecksum + FreeRTOS_htons( usRequest );
-        }
+                /* calculate the ICMP checksum for an outgoing packet. */
+                ( void ) usGenerateProtocolChecksum( ( uint8_t * ) pxICMPPacket, pxNetworkBuffer->xDataLength, pdTRUE );
+            }
+        #endif /* if ( ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM == 0 ) */
```
htibosch, line 2338: It would be desirable to clear the ICMP checksum 
~~~c
#else
{
    /* Many EMAC peripherals will only calculate the ICMP checksum
     * correctly if the field is nulled beforehand. */
    pxICMPHeader->usChecksum = 0U;
}
~~~

htibosch (approved):

FreeRTOS_IP.c
bernd-edlinger: But the IP header checksum is different?
I don't see the TCP checksum nulled either?

bernd-edlinger:
FYI the problem in the CI check "CBMC Proofs" is solved.

FreeRTOS_IP.c
htibosch: We should ask the chip designers why the ICMP checksum must be zero, whereas the IP- and other protocol-checksums may have a random value. Strange indeed.

bernd-edlinger: okay, so I will zero out the ICMP header checksum, but leave the IP header checsum unchanged.
right?

bernd-edlinger: > Update: the above avoids code like this

Oh, I see.  Should I remove this code in this PR ?

htibosch: Talking about this code in NetworkInterface.c:
~~~c
if( pxPacket->xICMPPacket.xIPHeader.ucProtocol == ( uint8_t ) ipPROTOCOL_ICMP )
{
    pxPacket->xICMPPacket.xICMPHeader.usChecksum = ( uint16_t ) 0u;
}
~~~
As I am not sure if users always have the same version of FreeRTOS_IP.c and NetworkInterface.c, it is maybe safer to keep it for now.

bernd-edlinger:
Hmm, I tried what happns if I send an option in the IP header of an ICMP packet.
`ping -R ` adds an RR option:
and the packets are returned but without the RR option
```C
13:58:20.098701 IP 10.20.240.22 > 169.254.237.190: ICMP echo request, id 39, seq 18, length 64
	0x0000:  7490 50c0 3131 704d 7b71 12c2 0800 4f00
	0x0010:  007c 9271 4000 4001 e9ed 0a14 f016 a9fe
	0x0020:  edbe 0107 2708 0a14 f016 0000 0000 0000
	0x0030:  0000 0000 0000 0000 0000 0000 0000 0000
	0x0040:  0000 0000 0000 0000 0000 0800 5438 0027
	0x0050:  0012 dcd9 9360 0000 0000 7381 0100 0000
	0x0060:  0000 1011 1213 1415 1617 1819 1a1b 1c1d
	0x0070:  1e1f 2021 2223 2425 2627 2829 2a2b 2c2d
	0x0080:  2e2f 3031 3233 3435 3637
13:58:20.099119 IP truncated-ip - 40 bytes missing! 169.254.237.190 > 10.20.240.22: ICMP echo reply, id 39, seq 18, length 104
	0x0000:  704d 7b71 12c2 7490 50c0 3131 0800 4500
	0x0010:  007c 9271 0000 4001 5628 a9fe edbe 0a14
	0x0020:  f016 0000 5438 0027 0012 dcd9 9360 0000
	0x0030:  0000 7381 0100 0000 0000 1011 1213 1415
	0x0040:  1617 1819 1a1b 1c1d 1e1f 2021 2223 2425
	0x0050:  2627 2829 2a2b 2c2d 2e2f 3031 3233 3435
	0x0060:  3637
`` 
the response has a correct ip header checksum, BUT an invalid `pxIPHeader->usLength`
I am not sure if I should drop those ICMP packets with options, or at least set the IP.usLength field
to something sensible.

**Comment by bernd-edlinger** [comment]
Okay, force-pushed a brand new version of the patch.
- fixed the ICMP packet with RR option.  I found that the response can be returned just without the unhandled options.
- fixed a bogus log message from `usGenerateProtocolChecksum()`, which was caused by returning raw CRC values instead of the documented return values.
- fixed a bogus tcp header checksum 0xFFFF which is apparently in conflict with  RFC 1624.

**Comment by bernd-edlinger** [comment]
BTW: this PR's number #237 is my favourite number (note 237 = 0xED my initials) :smile: :smile:

File: FreeRTOS_IP.c [hunk 2]
```diff
@@ -2356,7 +2360,8 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
  * @return eReleaseBuffer when the message buffer should be released, or eReturnEthernetFrame
  *                        when the packet should be returned.
  */
-    static eFrameProcessingResult_t prvProcessICMPPacket( ICMPPacket_t * const pxICMPPacket )
+    static eFrameProcessingResult_t prvProcessICMPPacket( ICMPPacket_t * const pxICMPPacket,
+                                                          NetworkBufferDescriptor_t * const pxNetworkBuffer )
     {
         eFrameProcessingResult_t eReturn = eReleaseBuffer;
```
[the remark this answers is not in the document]
bernd-edlinger: actually that could be an assertion, since the data length should be checked in
`usGenerateProtocolChecksum()` or `xCheckSizeFields()` dependent on
`ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM`
but when I look at the code in ipv6 branch you mentioned,
I miss the check for the destination address to be the local ip address.
Looks like it will answer broadcast pings?
Will it use the broadcast address in the response to such pings?

FreeRTOS_IP.c
```diff
@@ -2366,7 +2371,7 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
         {
             case ipICMP_ECHO_REQUEST:
                 #if ( ipconfigREPLY_TO_INCOMING_PINGS == 1 )
```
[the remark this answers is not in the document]
bernd-edlinger: done

FreeRTOS_IP.c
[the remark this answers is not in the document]
bernd-edlinger: the assertion uncovered the problem that I suspected previously:
https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/237#issuecomment-832679396
> The only thing that is odd in the test harnesh is that
pxIPPacket should alias pxNetworkBuffer->pucEthernetBuffer but it isn't.

fixed the test harnesh, and it works now.
so the assertion gives the cbmc something interesting to prove :-)

FreeRTOS_IP.c
```diff
@@ -2136,21 +2138,12 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
                          * went wrong because it will not be able to validate what it
                          * receives. */
                         #if ( ipconfigREPLY_TO_INCOMING_PINGS == 1 ) || ( ipconfigSUPPORT_OUTGOING_PINGS == 1 )
-                            if( pxNetworkBuffer->xDataLength >= sizeof( ICMPPacket_t ) )
                             {
-                                /* Map the buffer onto a ICMP-Packet struct to easily access the
-                                 * fields of ICMP packet. */
-                                ICMPPacket_t * pxICMPPacket = ipCAST_PTR_TO_TYPE_PTR( ICMPPacket_t, pxNetworkBuffer->pucEthernetBuffer );
-
                                 if( pxIPHeader->ulDestinationIPAddress == *ipLOCAL_IP_ADDRESS_POINTER )
```
bernd-edlinger, line 2142: @htibosch I think you'll need this in your ipv6 branch

FreeRTOS_IP.c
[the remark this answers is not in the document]
bernd-edlinger: okay looks now almost like in the ipv6 branch, but I left the assertion, because I like it double-safe...

FreeRTOS_IP.c
htibosch: `ipLOCAL_IP_ADDRESS_POINTER` only exists in the IPv4 branch.

bernd-edlinger: If I see it right both unicast and broadcast packets arrive here.
But I doubt that broadcast pings (`ping -b 255.255.255.255`) are handled well.

FreeRTOS_IP.c
```diff
@@ -2356,29 +2350,42 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
...
  */
-    static eFrameProcessingResult_t prvProcessICMPPacket( ICMPPacket_t * const pxICMPPacket )
+    static eFrameProcessingResult_t prvProcessICMPPacket( NetworkBufferDescriptor_t * const pxNetworkBuffer )
     {
         eFrameProcessingResult_t eReturn = eReleaseBuffer;
 
         iptraceICMP_PACKET_RECEIVED();
 
-        switch( pxICMPPacket->xICMPHeader.ucTypeOfMessage )
+        configASSERT( pxNetworkBuffer->xDataLength >= sizeof( ICMPPacket_t ) );
```
AniruddhaKanhere, line 2359: I don't like this particular `configASSERT` - because config asserts are supposed to test conditions which should ideally never happen. But this can indeed happen with a malformed packet - and that should not cause the TCP stack to give up - we should not process the said packet and move on.

Or did I miss something?

AniruddhaKanhere: Your thoughts @htibosch?

test/cbmc/proofs/parsing/ProcessIPPacket/Makefile.json
```diff
@@ -12,6 +12,7 @@
   ],
   "DEF":
   [
+    "ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM=1",
```
AniruddhaKanhere, line 15: This would reduce the coverage. I shall look at this. I would ideally keep this in two different configs and test both.
Will update shortly.

AniruddhaKanhere (commented):
Hello @bernd-edlinger, thanks for this PR.
I have a question/comment.  It would be great if you can answer :).

FreeRTOS_IP.c
bernd-edlinger: too short packets should be rejected in `prvAllowIPPacket` and/or `xCheckSizeFields`
where we have code like this
```C
            else if( ( ucProtocol == ( uint8_t ) ipPROTOCOL_ICMP ) ||
                     ( ucProtocol == ( uint8_t ) ipPROTOCOL_IGMP ) )
            {
                uxMinimumLength = uxIPHeaderLength + ipSIZE_OF_ETH_HEADER + ipSIZE_OF_ICMP_HEADER;
            }
            else
            {
                /* Unhandled protocol, other than ICMP, IGMP, UDP, or TCP. */
                DEBUG_SET_TRACE_VARIABLE( xLocation, 5 );
                break;
            }

            if( uxBufferLength < uxMinimumLength )
            {
                DEBUG_SET_TRACE_VARIABLE( xLocation, 6 );
                break;
            }
```

test/cbmc/proofs/parsing/ProcessIPPacket/Makefile.json
bernd-edlinger: I without this line, the CBMC proofs took forever, in the function that checks the checksum per software.
I tried to raise the loop limit to 100, but that did unfortunately never complete.

htibosch: Today I tested your branch "[fix_icmp_response_checksum](https://github.com/bernd-edlinger/FreeRTOS-Plus-TCP/tree/fix_icmp_response_checksum)", and played with the definition of `ipconfigFORCE_IP_DONT_FRAGMENT`. When it is defined as `1`, there will be a compilation error:
`ipFRAGMENT_FLAGS_DONT_FRAGMENT` undeclared, because my `ipconfigETHERNET_DRIVER_FILTERS_PACKETS` is non-zero, and so 
`ipFRAGMENT_FLAGS_DONT_FRAGMENT` is not defined, see [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/2134672c4876e12a978feb69ab98bf8dbcb81917/include/FreeRTOS_IP_Private.h#L358).

I would propose to remove the `#if` in FreeRTOS_IP_Private.h :
~~~
-#if ( ipconfigETHERNET_DRIVER_FILTERS_PACKETS == 0 )
     #if ( ipconfigBYTE_ORDER == pdFREERTOS_LITTLE_ENDIAN )
         /* The bits in the two byte IP header field that make up the fragment offset value. */
         #define ipFRAGMENT_OFFSET_BIT_MASK         ( ( uint16_t ) 0xff1fU )
         <snip>
         #define ipFRAGMENT_FLAGS_MORE_FRAGMENTS    ( ( uint16_t ) 0x2000U )
     #endif /* ipconfigBYTE_ORDER */
-#endif /* ipconfigETHERNET_DRIVER_FILTERS_PACKETS */
~~~
Ping the device went perfectly in both directions, also a heavy flood-ping from a Linux device.

Thanks

AniruddhaKanhere:
@bernd-edlinger, I have added configurations to the CBMC proof. These will allow the coverage to remain the same. To get around the checksum problem, I just stubbed them out since the actual checksum value is not important.

bernd-edlinger:
Ping?

htibosch:
Hello @bernd-edlinger , good that you send a ping.I think that the PR is all good. I am not sure what it is waiting for, but I think that @AniruddhaKanhere knows the answer to this question.

AniruddhaKanhere (approved):

paulbartell (approved):

PR #239, opened 2021-05-07
Orphans should not wakeup select

bernd-edlinger:
Fix select with ipconfigTCP_HANG_PROTECTION == 0:

The problem is that a select will not wake up when the
child socket is accepted, but only when some data is received.
However if the server wants to send data immediately after
accpet that will not work.

Also remove the code in prvTCPSocketCopy() which makes the
child socket inherit the socket set of the parent.
This is most likely just a work-around that makes the select
wake up when the first user data is received.

Select should not wake up on a child socket before it is
accepted.

After the socket is accepted the application may want to use
a different socket set.

Description
-----------

This is a bug-fix extracted from #229 

-----------


AniruddhaKanhere:
Hello @bernd-edlinger,

For now, I think I will close this PR. We can always reopen this if required and copy over your suggested changes.

Thanks,
Aniruddha

PR #256, opened 2021-05-20
Add port for ATSAME5x (Atmel SAM E5x)

ma-lalonde:
Atmel SAM E5x Network Interface

Description
-----------
Add a port (NetworkInterface.c) for ATSAME5x devices (Atmel SAM E5x). The port is based on Atmel's ASF4 GMAC driver, which remains unmodified. This port is NOT zero-copy and is generally meant to prioritize small size over performance.

Currently, configuration is done through Atmel START in Microchip Studio 7.0 and is stored in auto-generated file "hpl_gmac_config.h". 

Test Steps
-----------
See test project: https://github.com/ma-lalonde/ATSAME54N20A_FreeRTOS_Plus_TCP
This is developed for a custom ATSAME54N20A board, where there is a blinking LED for quick debugging (if the application freezes / crashes, the LED stops blinking).

Completed tests include DHCP, ICMP Echo (ping) send and receive (validated with WireShark, including CRC check), LLMNR (using the MCU's unique ID as network name), and iPerf3 testing using @htibosch's iperf3 server. Speed result is 14-15 Mbps with both BufferAllocation_1.c and BufferAllocation_2.c with CPU clock at 96 MHz, which makes me believe the true bottleneck is not being zero-copy. 

Related Issue
-----------
I needed the driver, so I made it, with some help: https://forums.freertos.org/t/re-freertos-tcp-on-sam-e5x


htibosch:
Hello [Marc-Antoine](https://github.com/ma-lalonde), thanks for this driver!

Unfortunately I do not have a ATSAME5x on my desk, otherwise I would be happy to see it working.
Does it work well enough for you?

> Speed result is 14-15 Mbps with both BufferAllocation_1.c and BufferAllocation_2.c with CPU clock at 96 MHz

That is indeed a bit low. I will be glad to help implementing zero-copy in two directions.

~~~c
mac_async_write( &ETH_MAC, pxDescriptor->pucEthernetBuffer, pxDescriptor->xDataLength );
~~~

but I'm afraid that the above function must be adapted.
You will find an example in ["NetworkInterface/DriverSAM"](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/portable/NetworkInterface/DriverSAM/NetworkInterface.c).

Note that you can implement zero-copy in two independent steps, incoming and outgoing.

When the throughput is low, it is also interesting to look at a PCAP that shows an IPERF3 session.
The IPERF3 server can be tuned with these defines:
~~~c
#define ipconfigIPERF_TX_BUFSIZE    ( 4 * ipconfigTCP_MSS )
#define ipconfigIPERF_TX_WINSIZE    ( 3 )
#define ipconfigIPERF_RX_BUFSIZE    ( 4 * ipconfigTCP_MSS )
#define ipconfigIPERF_RX_WINSIZE    ( 3 )
~~~


I just read your code again and it looks all good to me, thanks so far!

ma-lalonde:
Hi Hein,

I have updated the test project with the  iperf3 server. I had already roughly tuned the server to obtain better performance. I may have missed an important setting, or not.

I get a very consistent 14.1 Mbps in send and 18 Mbps in receive, until I hit a limit with the window size in receive. Then, WireShark starts reporting [ TCP Window Full ] and the receive bandwidth drops to 3.5 Mbps, with both BufferAllocation schemes. This seems to be a simple configuration issue, but I don't know how to solve it. There are no other error or warning messages from WireShark. Funny enough, the behavior (speed and consistency) is different depending on whether I use a switch between the ATSAME54 and the router or not.

I will not be putting more efforts into making the driver zero-copy at this point, as I need to move forward with the rest of the application. My use case is a very simple MQTT sensor app that reports at intervals >= 1s, so I am happy with the result and all the stuff I learned so far and hope it helps someone else get his/her project going.

AniruddhaKanhere (approved):
It looks good to me. If we need some changes, we can make them later on. Thank you @ma-lalonde for making this port and sharing that with all of us! You contribution is greatly appreciated.

aggarg (approved):

PR #257, opened 2021-05-21
Modify doxygen check to use common CI actions repository

AniruddhaKanhere:
Description
-----------
Fix Doxygen check. Now it requires an additional library to be installed.
Instead of doing that here, the repo switched to using common actions - there is a PR in the common actions repository: https://github.com/FreeRTOS/CI-CD-Github-Actions/pull/25

Test Steps
-----------

Related Issue
-----------


.github/workflows/ci.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       - name: Install Doxygen
         run: |
           wget -qO- "http://doxygen.nl/files/doxygen-1.9.1.linux.bin.tar.gz" | sudo tar --strip-components=1 -xz -C /usr/local
-          sudo apt-get install -y libclang-9-dev
+          sudo apt-get install -y libclang-9-dev libclang-cpp9 graphviz
```
aggarw13, line 97: You can just update this repository to use the shared action instead!

AniruddhaKanhere: Sure, I was thinking of fixing this for now as PRs are failing this check. I can follow up with a cleaned up ci.yml using the common github actions repository. Would that be okay?

aggarw13 (dismissed):

aggarw13 (approved):

paulbartell (approved):

PR #258, opened 2021-05-24
FreeRTOS_IP: remove CRC swap message

twpedersen:
Swapping the UDP CRC to 0xffff from 0x0000 is expected
behaivor as per RFC 768[1]. Remove the associated debug
message since it is not interesting to the user.

[1]: https://datatracker.ietf.org/doc/html/rfc768


htibosch:
I put it once there to test the logic of it, and forgot to remove the logging.
So yes, that is a good clean-up, thanks!

AniruddhaKanhere (approved):

n9wxu (approved):

PR #259, opened 2021-05-26
Refactor: Decouple dns cache, networking, parsing ,dns main

alfred2g:
Refactor: Decouple dns cache, networking, parsing ,dns main

Description
-----------
Refactor: Decouple dns cache, networking, parsing ,dns main


htibosch:
Alfred, thanks a lot for this refactoring.

I see many advantages of splitting up the DNS module. It must have been hard to understand and maintain it, if you're not used to it.

> Refactor: Decouple dns cache, networking, parsing, dns main

What I miss is a description at the top of this PR, a summary of the changes, and tell what will be the consequences for the user.

I would have wished to simplify DNS in the past, but FreeRTOS always had one promise to the users: every software update behaves like **drop-in replacements**. You can use any version of the kernel (> 8.0) in combination with any version of FreeRTOS+TCP: it will compile and work properly.
The kernel has 7 source files, +TCP has 9 source files, excluding the portable files.
I though you'd only decouple DNS_Cache? Now you are proposing to split-up FreeRTOS_DNS.c into 5 sub-modules.

DNS is one of the medium-size modules, there are four big source files:

    128 KB FreeRTOS_IP
    156 KB FreeRTOS_Sockets
    153 KB FreeRTOS_TCP_IP
     83 KB FreeRTOS_TCP_WIN

I am not sure if we want a distribution of let's say 30 essential source files.

FreeRTOS-Plus-TCP.alfred/FreeRTOS_IP.c: In function 'prvCheckNetworkTimers':
FreeRTOS-Plus-TCP.alfred/FreeRTOS_IP.c:796:17: error: implicit declaration of function 'vDNSCheckCallBack' [-Werror=implicit-function-declaration]
                 vDNSCheckCallBack( NULL );
                 ^~~~~~~~~~~~~~~~~

It would solved by including "DNS/DNS_Callback.h"?

Adding source files breaks every build, and so does putting header files at unusual locations. Do we require to add an -I ? Shouldn't this be documented in a "how to build"?

/tools/tcp_utilities/plus_tcp_demo_cli.c: In function 'pxDNSLookup':
/tools/tcp_utilities/plus_tcp_demo_cli.c:713:4: error: implicit declaration of function 'FreeRTOS_dnsclear'; did you mean 'FreeRTOS_select'? [-Werror=implicit-function-declaration]

Why don't you include "DNS/DNS_Cache.h" from FreeRTOS_DNS.h ?

You can not require to include new header files in order to get existing code running.

What is this file doing? 

    /private/FreeRTOS_DNS_Cache.h

I see the following includes:

    /* FreeRTOS+TCP includes. */
    #include "FreeRTOS_IP.h"

    /* Standard includes. */
    #include <stdint.h>

Should not be necessary to includ e"stdint.h" when "FreeRTOS.h" is already included. I am not sure if every compiler has "stdint.h"?

Finally one remark:

You probably know that `gethostbyname()` is already deprecated, and replaced by the more generic [`getaddrinfo()`](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/50688ce665b27f5556326d3c00ac16a4f1e597dd/FreeRTOS_DNS.c#L802). 

I think it would be good to implement `getaddrinfo()` for the IPv4 branch as well. It will return multiple results, and it helps to make the gap smaller to migrate from IPv4 to IPv6.

htibosch:
Hello @alfred2g, I have spoken to various people involved and every body agrees that we should get on with this refactoring. [Aniruddha](https://github.com/AniruddhaKanhere) convinced me that writing and maintaining the unit-tests will become a lot easier.

My only worry is the developers who will complain about incompatibility. We'll have to give clear information and examples.

So let's get on with this PR!

Thanks, Hein

alfred2g:
Hi @htibosch,

Thanks for the thorough  review, even though this initial change was just a proof of concept to demonstrate how easy it would be to unit test.
I will go ahead and implement/change the code to conform to your comments, if you have anything else that you would like to change/add , please let me know, I will happily do it.

Alfred

PR #261, opened 2021-06-01
Revising tcp_mem_stats.c

htibosch:
Description
-----------
In my previous [PR #260](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/260), I defined a macro in `FreeRTOSIPConfigDefaults.h`:
~~~c
#define ipconfigTCP_MEM_STATS_MAX_ALLOCATION   0
~~~
just to prevent a pragma/message showing up, saying: "`ipconfigTCP_MEM_STATS_MAX_ALLOCATION undefined?`"

This macro defines the length of an array, and so it should never be defined as zero. I removed the define, and added some comments in tcp_mem_stats.c:
~~~c
    /* If you don't want to see this pragma message, you can either
     * remove it or define 'ipconfigTCP_MEM_STATS_MAX_ALLOCATION' in
     * your freeRTOSIPConfig.h. */
~~~
When `ipconfigTCP_MEM_STATS_MAX_ALLOCATION` was zero, I discovered an issue: `vTaskSuspendAll()` would be called without ever calling `xTaskResumeAll()`!
I solved this by giving a single-point-of-exit to the function `vAddAllocation()`.

Related Issue
-----------
`vTaskSuspendAll()` is called without a corresponding `xTaskResumeAll()` when tcp_mem_stats.c is used.

Test Steps
-----------
Define `ipconfigTCP_MEM_STATS_MAX_ALLOCATION` as e.g. 10 or 0, and you will see that every call to `xTaskResumeAll()` will match with a call to `xTaskResumeAll()`.

### DIFF
--- a/.github/lexicon.txt
+++ b/.github/lexicon.txt
@@ -636,6 +636,7 @@ ppucdata
 ppucrecvdata
 ppxnetworkbuffer
 pr
+pragma
 pre
 printf
 printfs
--- a/include/FreeRTOSIPConfigDefaults.h
+++ b/include/FreeRTOSIPConfigDefaults.h
@@ -639,11 +639,6 @@
     #define ipconfigSELECT_USES_NOTIFY    0
 #endif
 
-#ifndef ipconfigTCP_MEM_STATS_MAX_ALLOCATION
-    /* See "/tools/tcp_utilities/tcp_mem_stats.c". */
-    #define ipconfigTCP_MEM_STATS_MAX_ALLOCATION    0
-#endif
-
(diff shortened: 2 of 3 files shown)


htibosch:
About the last change "Do check macros when the module is disabled" :

It makes no sense to warn the user about macro's if the entire modules is disabled, i.e. when `ipconfigUSE_TCP_MEM_STATS` equals zero.

AniruddhaKanhere (approved):

dan4thewin (approved):

PR #265, opened 2021-06-03
BufferAllocation_2: respect 0 length buffer request

twpedersen:
The pxGetNetworkBufferWithDescriptor() documentation states:

	If xRequestedSizeBytes is zero then the returned network
	buffer descriptor will not reference an Ethernet buffer
	(the reference is set to NULL).

But this did not match the implementation. 2 bytes were
unconditionally being added to the requested size, and the
returned NetworkBufferDescriptor_t would have a nonzero
pucEthernetBuffer.

Move the code performing length alignment as is into the
branch checking whether size is nonzero. This lets the
caller request a bare NetworkBufferDescriptor_t to eg.
implement zero copy RX.

### DIFF
--- a/portable/BufferManagement/BufferAllocation_2.c
+++ b/portable/BufferManagement/BufferAllocation_2.c
@@ -228,22 +228,6 @@ NetworkBufferDescriptor_t * pxGetNetworkBufferWithDescriptor( size_t xRequestedS
 
     if( xNetworkBufferSemaphore != NULL )
     {
-        if( ( xRequestedSizeBytes != 0U ) && ( xRequestedSizeBytes < ( size_t ) baMINIMAL_BUFFER_SIZE ) )
-        {
-            /* ARP packets can replace application packets, so the storage must be
-             * at least large enough to hold an ARP. */
-            xRequestedSizeBytes = baMINIMAL_BUFFER_SIZE;
-        }
-
-        /* Add 2 bytes to xRequestedSizeBytes and round up xRequestedSizeBytes
-         * to the nearest multiple of N bytes, where N equals 'sizeof( size_t )'. */
-        xRequestedSizeBytes += 2U;
-
-        if( ( xRequestedSizeBytes & ( sizeof( size_t ) - 1U ) ) != 0U )
-        {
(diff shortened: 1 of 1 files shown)


twpedersen:
`FAIL: ./portable/BufferManagement/BufferAllocation_2.c (File size changed from 15804 to 15986)` ?

yourslab:
Hi @twpedersen. Please use uncrustify to format `BufferAllocation_2.c` with the following config: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/tools/uncrustify.cfg

twpedersen:
thanks @yourslab. I ran `BufferAllocation_2.c` through uncrustify and repushed.

twpedersen:
Thanks @htibosch. Good catch. I think it might be better to initialize `xDataLength` along with `pucEthernetBuffer` while freeing the descriptor in`vReleaseNetworkBufferAndDescriptor()`?

htibosch:
> I think it might be better to initialize `xDataLength` along with `pucEthernetBuffer` while freeing the descriptor in`vReleaseNetworkBufferAndDescriptor()`?

Do you mean the following location?
~~~c
 void vReleaseNetworkBufferAndDescriptor( NetworkBufferDescriptor_t * const pxNetworkBuffer )
 {
     BaseType_t xListItemAlreadyInFreeList;

     /* Ensure the buffer is returned to the list of free buffers before the
     * counting semaphore is 'given' to say a buffer is available.  Release the
     * storage allocated to the buffer payload.  THIS FILE SHOULD NOT BE USED
     * IF THE PROJECT INCLUDES A MEMORY ALLOCATOR THAT WILL FRAGMENT THE HEAP
     * MEMORY.  For example, heap_2 must not be used, heap_4 can be used. */
     vReleaseNetworkBuffer( pxNetworkBuffer->pucEthernetBuffer );
     pxNetworkBuffer->pucEthernetBuffer = NULL;
+    pxNetworkBuffer->xDataLength = 0U;

     taskENTER_CRITICAL();
~~~
I also think that is semantically more correct, `pucEthernetBuffer` and `xDataLength` are kept together.

twpedersen:
Exactly. I updated the PR with that change.

n9wxu (approved):

PR #266, opened 2021-06-11
Update kernel version and manifest information

AniruddhaKanhere:
Description
-----------
This PR updates the Kernel submodule version and updates the manifest to reflect the same.

Test Steps
-----------

Related Issue
-----------


gitwallit (dismissed):

manifest.yml
```diff
@@ -5,7 +5,7 @@ description: |-
 
 dependencies:
   - name: "FreeRTOS-Kernel"
-    version: "V10.4.3"
+    version: "V10.4.4"
     repository:
       type: "git"
       url: "https://github.com/FreeRTOS/FreeRTOS-Kernel"
```
paulbartell, line 11: Could you add the "path" attribute here to specify the relative path of the submodule?

paulbartell (commented):
If we have a manifest in this repo, should we also have a cmake include file similar to the other libraries?

manifest.yml
AniruddhaKanhere: Done.

paulbartell (approved):

PR #269, opened 2021-06-22
Refactor the TCP_IP module

AniruddhaKanhere:
Description
-----------

Test Steps
-----------

Related Issue
-----------


htibosch:
@AniruddhaKanhere , thanks again for this refactoring!

I have tried the TCP protocol by running FTP sessions. That uses many features of the TCP library.
I tried both **active** and **passive** TCP connections.
There were also zero-byte connections, in case an empty file is being copied with FTP.

Also interesting is sending the last bytes along with a FIN flag for a quick closure using the socket option [FREERTOS_SO_CLOSE_AFTER_SEND](https://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/API/setsockopt.html).

It passed all tests and with the same great performance as usual.

There were 3 problems while compiling:
~~~
xTCPCheckNewClient not declared in FreeRTOS_IP.c
Maybe "FreeRTOS_TCP_IP.h" must be included?

eTCP_LISTEN is undefined in FreeRTOS_Sockets.c
Maybe "FreeRTOS_TCP_IP.h" must be included?

vTCPStateChange() is undefined in FreeRTOS_Sockets.c
Shouldn't "FreeRTOS_TCP_IP_utils.h" also be included?
~~~
As you know, I adapted my Makefile(s) as follows to include the new files, only when present:
~~~make
ifeq (,$(wildcard $(PLUS_TCP_PATH)/FreeRTOS_TCP_IP_Reception.c))
    # old sources before split-up
else
	C_SRCS += \
		$(PLUS_TCP_PATH)/FreeRTOS_TCP_IP_Reception.c \
		$(PLUS_TCP_PATH)/FreeRTOS_TCP_IP_StateHandling.c \
		$(PLUS_TCP_PATH)/FreeRTOS_TCP_IP_TimerWork.c \
		$(PLUS_TCP_PATH)/FreeRTOS_TCP_IP_utils.c
endif
~~~
That makes it possible to easily switch between software releases, only `$(PLUS_TCP_PATH)` will change.

htibosch:
Hi Aniruddha,

Could you add to FreeRTOS_IP.c:
~~~c
#if( ipconfigUSE_TCP == 1 )
    #include "FreeRTOS_TCP_IP.h"
#endif
~~~
and from FreeRTOS_Sockets.c :
~~~c
#if( ipconfigUSE_TCP == 1 )
    #include "FreeRTOS_TCP_IP.h"
    #include "FreeRTOS_TCP_IP_utils.h"
#endif
~~~
so that the two modules can compile?

AniruddhaKanhere:
> Hi Aniruddha,
> 
> Could you add to FreeRTOS_IP.c:
> 
> ```c
> #if( ipconfigUSE_TCP == 1 )
>     #include "FreeRTOS_TCP_IP.h"
> #endif
> ```
> 
> and from FreeRTOS_Sockets.c :
> 
> ```c
> #if( ipconfigUSE_TCP == 1 )
>     #include "FreeRTOS_TCP_IP.h"
>     #include "FreeRTOS_TCP_IP_utils.h"
> #endif
> ```
> 
> so that the two modules can compile?

Hello Hein!
Of course. This is still changing and being modified with unit-tests etc. But yes, I'll add the header files to the source files you mentioned. Thank you for figuring that out! 

Also, we need to come up with a folder structure which would be beneficial to our user/customer. I am looking at other TCP stacks to figure out a good way to arrange all these new files in a good manner.

PR #278, opened 2021-07-21
Make sure that SYN packets sent to a broadcast IP address are discarded

AniruddhaKanhere:
Description
-----------
RFC 1122 section 3.2.1.3 and 4.2.3.10 point that a broadcast packet (255.255.255.255) must be discarded silently by the device. Make sure that is taken care of.
> A TCP implementation MUST silently discard an incoming SYN
> segment that is addressed to a broadcast or multicast
> address.

Test Steps
-----------

Related Issue
-----------


FreeRTOS_TCP_IP.c
```diff
@@ -3725,12 +3725,22 @@
...
 
-        /* Assume that a new Initial Sequence Number will be required. Request
-         * it now in order to fail out if necessary. */
-        ulInitialSequenceNumber = ulApplicationGetNextSequenceNumber( *ipLOCAL_IP_ADDRESS_POINTER,
-                                                                      pxSocket->usLocalPort,
-                                                                      pxTCPPacket->xIPHeader.ulSourceIPAddress,
-                                                                      pxTCPPacket->xTCPHeader.usSourcePort );
+        /* Silently discard a SYN packet which was sent to the broadcast address
+         * 255.255.255.255. */
+        if( pxTCPPacket->xIPHeader.ulDestinationIPAddress != 0xFFFFFFFF )
```
htibosch, line 3730: Aniruddha, I think that we should also test for a local broadcast address "x.x.x.255", and for multicast addresses. Maybe in a separate function:
~~~c
BaseType_t xIsUnicastAddress( uint32_t ulIPAddress )
{
    BaseType_t xResult = pdTRUE;

    if( ( FreeRTOS_ntohl( ulIPAddress ) & 0xffU ) == 0xffU )
    {
        /* This is a broadcast address x.x.x.255. */
        xResult = pdFALSE;
    }
    else if( xIsIPv4Multicast( ulIPAddress ) != pdFALSE )
    {
        /* This is a multicast address. */
        xResult = pdFALSE;
    }

    return xResult;
}
~~~
And so make this change:
~~~c
-    if( pxTCPPacket->xIPHeader.ulDestinationIPAddress != 0xFFFFFFFF )
+    if( xIsUnicastAddress( pxTCPPacket->xIPHeader.ulDestinationIPAddress ) != pdFALSE )
~~~

htibosch (dismissed):
Thank you for this PR. Broadcast and multicast addresses can indeed reach the TCP handlers, and we should ignore them.

I propose to make `xIsUnicastAddress()` a global function so that it can be called from the network interfaces for early address filtering.

FreeRTOS_TCP_IP.c
AniruddhaKanhere: I agree, added.
Thank you.

FreeRTOS_TCP_IP.c
```diff
@@ -3727,12 +3763,22 @@
...
+             * it now in order to fail out if necessary. */
+            ulInitialSequenceNumber = ulApplicationGetNextSequenceNumber( *ipLOCAL_IP_ADDRESS_POINTER,
+                                                                          pxSocket->usLocalPort,
+                                                                          pxTCPPacket->xIPHeader.ulSourceIPAddress,
+                                                                          pxTCPPacket->xTCPHeader.usSourcePort );
+        }
+        else
+        {
+            /* Set the sequence number to 0 to avoid further processing. */
+            ulInitialSequenceNumber = 0;
```
htibosch, line 3780: A tiny remark:
~~~c
-    ulInitialSequenceNumber = 0;
+    ulInitialSequenceNumber = 0U;
~~~

FreeRTOS_TCP_IP.c
```diff
@@ -423,6 +423,42 @@
...
 
+/**
+ * @brief Check whether the address is unicast of multicast/broadcast.
+ *
+ * @param[in] ulIPAddress: The IP address to be checked (in 32-bit format).
+ *
+ * @return pdTRUE if the IP address is either multicast or broadcast. pdFALSE
+ *         otherwise.
+ */
+    BaseType_t xIsUnicastAddress( uint32_t ulIPAddress )
```
htibosch, line 434: Although I came up with the above function signature, maybe it is better to pass a socket address:
~~~c
    BaseType_t xIsUnicastAddress( struct freertos_sockaddr * pxAddress )
    {
        uint32_t ulIPAddress = pxAddress->sin_addr;
        ...
    }
~~~
The advantage is that this will be compatible with IPv6. Would you mind to change it?

htibosch (approved):
Thanks a lot, I approve this PR.
Hein

FreeRTOS_TCP_IP.c
```diff
@@ -418,6 +418,47 @@
...
+            xResult = pdFALSE;
+        }
+        else if( xIsIPv4Multicast( ulIPAddress ) != pdFALSE )
+        {
+            /* This is a multicast address. */
+            xResult = pdFALSE;
+        }
+        else if( ulIPAddress == 0U )
+        {
+            /* This is a non-standard broadcast address. */
```
htibosch, line 450: You can also use the value `FREERTOS_INADDR_ANY`, instead of `0U`.
Comment is not correct, it is a "an invalid, unknown or non-applicable target".
Quoted text borrowed from Wikipedia: https://en.wikipedia.org/wiki/0.0.0.0
( that is why I wanted `inet_pton()` to clear the IP-address ).

FreeRTOS_TCP_IP.c
```diff
@@ -3711,13 +3752,27 @@
...
-        ulInitialSequenceNumber = ulApplicationGetNextSequenceNumber( *ipLOCAL_IP_ADDRESS_POINTER,
-                                                                      pxSocket->usLocalPort,
-                                                                      pxTCPPacket->xIPHeader.ulSourceIPAddress,
-                                                                      pxTCPPacket->xTCPHeader.usSourcePort );
+        /* Put the IP address in the address-structure. */
+        xAddress.sin_addr = pxTCPPacket->xIPHeader.ulDestinationIPAddress;
+
+        /* Silently discard a SYN packet which was sent to the broadcast address
+         * (x.x.x.255) or a multicast address. */
+        if( xIsUnicastAddress( &xAddress ) != pdFALSE )
```
htibosch, line 3762: When we started this PR, we said that broadcast and multicast packets carrying a SYN flag should be silently discarded.
We already filter out other IP-addresses, only the device's IP address can be targeted.... and multi- and broadcast packets.

Now what if we don't call `xIsUnicastAddress()` but in stead a test like this?
~~~
if( pxTCPPacket->xIPHeader.ulDestinationIPAddress == *ipLOCAL_IP_ADDRESS_POINTER )
{
    ulInitialSequenceNumber = ulApplicationGetNextSequenceNumber( *ipLOCAL_IP_ADDRESS_POINTER );
}
else
{
    ulInitialSequenceNumber = 0U;
}
~~~
That makes the patch much smaller and more effective.

htibosch (commented):
I'm sorry that I propose a change, overruling my earlier proposal. But I think that the simpler solution is more effective, also non-cast packets with the wrong target  will be filtered out.

FreeRTOS_TCP_IP.c
AniruddhaKanhere: I made this change but want to note that earlier change would have been more compatible with IPv6/multi.

htibosch: You're right about that, we could use the same code in both /single and in /multi

n9wxu (approved):

PR #279, opened 2021-07-24
When storing the last TCP packet, clear the TCP flags.

htibosch:
Description
-----------
In response to a report on the [FreeRTOS forum](https://forums.freertos.org/t/freertos-tcp-corehttp-connection-reset-from-server-side): "FreeRTOS+TCP coreHTTP connection reset from server side".

A FreeRTOS+TCP device sends a SYN, which is answered twice with an identical SYN+ACK.

[Janis Barscausks](https://forums.freertos.org/u/janis.barscausks) wrote that he has no idea why his Apache server on Ubuntu replied with 2 identical SYN+ACK packets.

Both packets received a correct reply: a pure ACK, which is all good.

But as a side-effect, the second packet carrying SYN+ACK got stored (in `xTCP.xPacket.u.ucLastPacket`), along with the SYN flag. When the device sends the first data, it will also send the SYN flag, which is incorrect.

Test Steps
-----------
I tested by making a hack: as soon as a SYN+TCP is received, I presented the packet twice to the TCP code. The first data packet contained the SYN flag.

When the SYN flag is not stored, this will not occur anymore.

I tested the change by making thousands of connections, both passive and active.

### DIFF
--- a/FreeRTOS_TCP_IP.c
+++ b/FreeRTOS_TCP_IP.c
@@ -3623,6 +3623,7 @@
                     }
                     else
                     {
+                        ProtocolHeaders_t * pxProtocolHeaders;
                         /* Update the copy of the TCP header only (skipping eth and IP
                          * headers).  It might be used later on, whenever data must be sent
                          * to the peer. */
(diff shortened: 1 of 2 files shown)


FreeRTOS_TCP_IP.c
```diff
@@ -3623,13 +3623,19 @@
                     }
                     else
                     {
+                        ProtocolHeaders_t * pxProtocolHeaders;
```
AniruddhaKanhere, line 3626: Just a thought: We might want to avoid mixed code and declaration. I know this makes code more readable - but we still might want to refrain from doing this.

AniruddhaKanhere (approved):

FreeRTOS_TCP_IP.c
htibosch: > Just a thought: We might want to avoid mixed code and declaration. I know this makes code more readable - but we still might want to refrain from doing this.

I agree that mixing code with declarations decreases the quality, and mistakes will become more likely.  But `pxProtocolHeaders` was declared at the beginning of a compound block.  There are many places in the +TCP code where local variables are declared at the beginning of a compound block, thus limiting their scope, which I think is good.

However, I didn't like the long declaration neither, and want to clear the TCP flags with an indexed assignment:
~~~
    ucLastPacket[ uxOffset + ipTCP_FLAGS_OFFSET ] = tcpTCP_FLAG_ACK;
~~~
where `ipTCP_FLAGS_OFFSET` is defined as 13U. This will also work for IPv6.

htibosch: PS. I just retested the solution, by running it twice: with and without clearing the TCP flags in `ucLastPacket`. It worked as expected.

AniruddhaKanhere (approved):

paulbartell (approved):

PR #288, opened 2021-07-30
Remove extra MSS field from TCP sockets

AniruddhaKanhere:
Description
-----------
This PR removes an extra MSS field (`usCurMSS`) in the TCP sockets since it was always identical to another field (`usInitMSS`). Both of these fields are then replaced by a new field (`usMSS`).

Test Steps
-----------

Related Issue
-----------


htibosch:
Thanks for this clean-up. Originally the two fields had a different meaning, and after some changes, they became the same.
Thanks for this PR, it has my approval.

cobusve (dismissed):

paulbartell (approved):

PR #289, opened 2021-07-30
Remove unused macro

AniruddhaKanhere:
Description
-----------
The macro `ipconfigTCP_ACK_EARLIER_PACKET` was added to allow users to determine if they want to acknowledge earlier packets or not in a TCP connection. But, this is not used except in testing - meaning, that the macro is defined as 1.

This PR removes the macro and the corresponding code where the macro value was `0`.

Not sure why did uncrustify decide to modify the `define` s.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/FreeRTOS_TCP_IP.c
+++ b/FreeRTOS_TCP_IP.c
@@ -77,19 +77,19 @@
 /*
  * A few values of the TCP options:
  */
-    #define tcpTCP_OPT_END                        0U /**< End of TCP options list. */
-    #define tcpTCP_OPT_NOOP                       1U /**< "No-operation" TCP option. */
-    #define tcpTCP_OPT_MSS                        2U /**< Maximum segment size TCP option. */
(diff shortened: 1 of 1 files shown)


htibosch:
Thanks Aniruddha, for this PR, it looks good to me.

I will add some comments for those who wonder what this is all about: this code decides if an ACK should be delayed. When the device is receiving a continuous stream of data, it will only want to send an ACK when these conditions are all true:

● In case the peer has enough space in his TCP TX window
● Only when the TCP connection is established
● Only when a FIN has not yet been sent
● Our TX queue is empty
● We only need to send a pure ACK ( i.e. no other flags )

The removed macro was called `ipconfigTCP_ACK_EARLIER_PACKET`, meaning that when defined, an ACK will be sent earlier, as soon as the space drops below 2 x MSS (`u.xTCP.usCurMSS`).

cobusve (approved):

PR #294, opened 2021-08-03
IPv6/Multi: Remove extra MSS field from TCP sockets

htibosch:
Description
-----------
Just like we did recently in #288, this PR ( based on [IPv6/multi](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi) ) will replace two fields with a single one:
~~~c
-    uint16_t usCurMSS;        /**< Current Maximum Segment Size */
-    uint16_t usInitMSS;       /**< Initial maximum segment Size */
+    uint16_t usMSS;           /**< The Maximum Segment Size for the current connection. */
~~~
The fields are entirely private to the library, just like all contents of a socket structure.
They can be joined because in fact they always have the same value.

Test Steps
-----------
I tested the changes by running IPERF with a varying value for `ipconfigNETWORK_MTU`, which influence the MSS ( Maximum Segment Size ).

Related Issue
-----------
This PR doesn't have the intention to solve an issue.

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -640,8 +640,7 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
                         {
                             /* StreamSize is expressed in number of bytes */
                             /* Round up buffer sizes to nearest multiple of MSS */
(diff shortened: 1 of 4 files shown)


htibosch:
There are conflicts with 3 files because of other PR's which got merged. I think it's the easiest to recreade this PR, based on the current sources.

PR #299, opened 2021-08-03
IPv6/Multi: Run uncrustify before applying complexity

htibosch:
Description
-----------
This is a copy of PR #270, in order to be merged with the [IPv6/Multi](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi) branch.

In short: the [GNU tool complexity](https://www.gnu.org/software/complexity) has problems with sources files that have indented functions. It just doesn't analyse/score those functions.

For instance, in FreeRTOS_TCP.c, there is a global `#if`:
~~~c
#if ipconfigUSE_TCP == 1
    /* Here all code, indented 4 spaces. */
#endif
~~~
The same is seen in the kernel, e.g. timers.c:
~~~c
#if ( configUSE_TIMERS == 1 )
    /* All code compiled conditionally. */
#endif
~~~
The complexity tool can not handle this. This PR will change [ci.yml](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/labs/ipv6_multi/.github/workflows/ci.yml), so that it will first unindent the functions in a temporary source file.


htibosch:
The PR's that reduce the complexity of the IPv6/multi branch are not yet merged. 

Therefore, the test for complexity will fail.

Here is the current top-5:
~~~txt
   77     407     284   ./temp/FreeRTOS_TCP_IP.c(747): prvTCPReturnPacket
  104     407     268   ./temp/FreeRTOS_DHCP.c(343): vDHCPProcessEndPoint
  191     281     224   ./temp/FreeRTOS_DHCPv6.c(826): prvDHCPv6Analyse
  269     287     184   ./temp/FreeRTOS_DHCP.c(879): prvProcessDHCPReplies
  507     534     389   ./temp/FreeRTOS_DNS.c(1866): prvParseDNSReply
~~~

.github/workflows/ci.yml
```diff
@@ -123,10 +123,30 @@ jobs:
       - uses: actions/checkout@v2
       - name: Setup
         run: sudo apt-get install complexity
+      - name: Install Uncrustify
+        run: sudo apt-get install uncrustify
       - name: Complexity
         run: |
-          find . -iname '*.c' |\
-          xargs complexity --scores --threshold=0 --horrid-threshold=55
+          COMPLEXITY_PARAMS="--scores --threshold=10 --horrid-threshold=55"
```
AniruddhaKanhere, line 130: You might want to add a different horrid threshold... This threshold is not enough after this change

AniruddhaKanhere (changes requested):
Thank you for this PR. Would you mind making the change(s) requested below?

PR #301, opened 2021-08-08
Check return value of inet_pton4

AniruddhaKanhere:
Description
-----------
This PR adds check for the return value of `FreeRTOS_inet_pton4` in `FreeRTOS_inet_addr` such that if the address translation failed, then 0 should be returned as IP address. 

Test Steps
-----------

Related Issue
-----------


FreeRTOS_Sockets.c
```diff
@@ -2551,7 +2551,11 @@ uint32_t FreeRTOS_inet_addr( const char * pcIPAddress )
     uint32_t ulReturn = 0UL;
 
     /* inet_pton AF_INET target is a 4-byte 'struct in_addr'. */
-    ( void ) FreeRTOS_inet_pton4( pcIPAddress, &( ulReturn ) );
+    if( pdFAIL == FreeRTOS_inet_pton4( pcIPAddress, &( ulReturn ) ) )
+    {
+        /* Return 0 if translation failed. */
+        ulReturn = 0;
```
htibosch, line 2557: It is good to test the result of `FreeRTOS_inet_pton4()`, although it must be said the `ulReturn ` will be cleared in case of an error.

PS. MISRA likes an unsigned const : `ulReturn = 0U;`

htibosch (approved):
Like I said, the PR doesn't change the behaviour, but yet it is preferable to test the return value and clear the IPv4 address returned.
I approve this PR, thanks

FreeRTOS_Sockets.c
AniruddhaKanhere: Thank you for the review @htibosch. I have update the value as you mentioned above.

aggarg (approved):

PR #302, opened 2021-08-08
Check hostname length even without DNS cache 

AniruddhaKanhere:
Description
-----------
This PR adds a check on DNS hostname length when DNS cache is not enabled such that the maximum length is limited to 255 (`0xFF`) according to [RFC 1035](https://www.ietf.org/rfc/rfc1035.txt) section 3.1.

Test Steps
-----------

Related Issue
-----------


htibosch:
@AniruddhaKanhere, thank you for this patch, it looks good and I approve it.

Hein

n9wxu (approved):

PR #306, opened 2021-08-14
IPv6/multi: Fix FreeRTOS_recvfrom function

AniruddhaKanhere:
Description
-----------
This PR fixes the implementation of `FreeRTOS_recvfrom` to return a proper length of data and also update the sender's address info to include the 'family' (IPv4/IPv6) information and return an IPv6 address in case of IPv6 packet.

Test Steps
-----------

Related Issue
-----------


FreeRTOS_Sockets.c
```diff
@@ -1263,9 +1263,16 @@ int32_t FreeRTOS_recvfrom( Socket_t xSocket,
                     {
                         uxPayloadOffset = ipUDP_PAYLOAD_OFFSET_IPv6;
                         xAddressLength = sizeof( struct freertos_sockaddr6 );
+                        pxSourceAddress->sin_family = ( uint8_t ) FREERTOS_AF_INET6;
```
htibosch, line 1266: `pxSourceAddress` might be NULL, it should be tested.

The following expression is correct:
~~~c
uxPayloadLength = pxNetworkBuffer->xDataLength - ( ipSIZE_OF_ETH_HEADER + uxIPHeaderSizePacket( pxNetworkBuffer ) + sizeof( UDPHeader_t ) );
~~~
but we had `uxPayloadLength` already, which carries the same value:
~~~c
uxPayloadLength = pxNetworkBuffer->xDataLength - uxPayloadOffset;
~~~
Earlier in the chain, it was proven that `xDataLength` is larger than or equal to `uxPayloadOffset`.

I played a bit with the code and propose this block:
~~~c
if( pxNetworkBuffer != NULL )
{
    #if ( ipconfigUSE_IPv6 != 0 )
        {
            UDPPacket_t * pxUDPPacket = ipCAST_PTR_TO_TYPE_PTR( UDPPacket_t, pxNetworkBuffer->pucEthernetBuffer );

            #if ( ipconfigUSE_IPv6 != 0 )
                if( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv6_FRAME_TYPE )
                {
                    if( pxSourceAddress != NULL )
                    {
                        sockaddr6_t * pxSourceAddressV6 = ipCAST_PTR_TO_TYPE_PTR( sockaddr6_t, pxSourceAddress );
                        UDPPacket_IPv6_t * pxUDPPacket = ipCAST_PTR_TO_TYPE_PTR( UDPPacket_IPv6_t, pxNetworkBuffer->pucEthernetBuffer );
                        memcpy( ( void * ) pxSourceAddressV6->sin_addrv6.ucBytes,
                                ( void * ) pxUDPPacket->xIPHeader.xSourceAddress.ucBytes,
                                ipSIZE_OF_IPv6_ADDRESS );
                        pxSourceAddress->sin_family = ( uint8_t ) FREERTOS_AF_INET6;
                        pxSourceAddress->sin_addr = 0U;
                        pxSourceAddress->sin_port = pxNetworkBuffer->usPort;
                    }

                    uxPayloadOffset = ipUDP_PAYLOAD_OFFSET_IPv6;
                    xAddressLength = sizeof( struct freertos_sockaddr6 );
                }
                else
            #endif /* ( ipconfigUSE_IPv6 != 0 ) */

            {
                uxPayloadOffset = ipUDP_PAYLOAD_OFFSET_IPv4;

                if( pxSourceAddress != NULL )
                {
                    pxSourceAddress->sin_family = ( uint8_t ) FREERTOS_AF_INET;
                    pxSourceAddress->sin_addr = pxNetworkBuffer->ulIPAddress;
                    pxSourceAddress->sin_port = pxNetworkBuffer->usPort;
                }
            }
        }
    #endif /* if ( ipconfigUSE_IPv6 != 0 ) */

    if( pxSourceAddressLength != NULL )
    {
        *pxSourceAddressLength = xAddressLength;
    }

    /* The returned value is the length of the payload data, which is
     * calculated at the total packet size minus the headers.
     * The validity of `xDataLength` prvProcessIPPacket has been confirmed
     * in 'prvProcessIPPacket()'. */
    uxPayloadLength = pxNetworkBuffer->xDataLength - uxPayloadOffset;
    lReturn = ( int32_t ) uxPayloadLength;

    lReturn = prvRecvFrom_CopyPacket( &( pxNetworkBuffer->pucEthernetBuffer[ uxPayloadOffset ] ), pvBuffer, uxBufferLength, xFlags, lReturn );

    if( ( ( UBaseType_t ) xFlags & ( ( ( UBaseType_t ) FREERTOS_MSG_PEEK ) | ( ( UBaseType_t ) FREERTOS_ZERO_COPY ) ) ) == 0U )
    {
        vReleaseNetworkBufferAndDescriptor( pxNetworkBuffer );
    }
}
~~~

I tested everything with a Windows/Linux host. It worked well, and the contents of `pxSourceAddress` looks good.

htibosch (commented):
@AniruddhaKanhere, thanks for this PR again!

I tested it in real hardware, and it looks good, both with IPv4 and IPv6 packets.
It will need some changes like testing for `NULL`, and a few other changes to please MISRA.
For instance to pass character arrays tot `memcpy()` in stead of the address of a struct.
Hein

AniruddhaKanhere:
Thank you @htibosch for the review.
Yes, I agree with the clean up you suggested. I have applied your changes (after some modification) to my PR.

The modifications I made were to make the cod look cleaner by rearranging and removing some assignments. I also removed one extra #if ipconfigUSE_IPv6 since after the suggested changes, if we disable IPv6, nothing would work. And some other minor changes.

I think this should work well.

FreeRTOS_Sockets.c
AniruddhaKanhere: Done so, please see my comment. Thank you.

FreeRTOS_Sockets.c
```diff
@@ -1254,18 +1256,39 @@ int32_t FreeRTOS_recvfrom( Socket_t xSocket,
 
         if( pxNetworkBuffer != NULL )
         {
-            uxPayloadOffset = ipUDP_PAYLOAD_OFFSET_IPv4;
             #if ( ipconfigUSE_IPv6 != 0 )
-                {
-                    UDPPacket_t * pxUDPPacket = ipCAST_PTR_TO_TYPE_PTR( UDPPacket_t, pxNetworkBuffer->pucEthernetBuffer );
+                UDPPacket_IPv6_t * pxUDPPacketV6 = ipCAST_PTR_TO_TYPE_PTR( UDPPacket_IPv6_t, , pxNetworkBuffer->pucEthernetBuffer );
```
htibosch, line 1260: Yes that is a good change, thank you

htibosch (approved):
I also think it's all good now, thank you!

**EDIT** : I retested the functions and the results are as expected.

cobusve (approved):

PR #308, opened 2021-08-18
IPv6/multi: Update IP header length after removing IP-options

AniruddhaKanhere:
Description
-----------
FreeRTOS+TCP doesn't support IP-options. While skipping over (and removing) the IP options, the IP-header fields are updated to reflect the same. However, the total length was not being updated. This PR fixes that issue.

Test Steps
-----------
Send a packet with IP-options (best to send an ICMP packet).
Receive the ICMP response which will have incorrect IP packet length.

Related Issue
-----------


FreeRTOS_IP.c
```diff
@@ -2802,6 +2802,8 @@ static eFrameProcessingResult_t prvCheckIP4HeaderOptions( NetworkBufferDescripto
             /* Rewrite the Version/IHL byte to indicate that this packet has no IP options. */
             pxIPHeader->ucVersionHeaderLength = ( pxIPHeader->ucVersionHeaderLength & 0xF0U ) | /* High nibble is the version. */
                                                 ( ( ipSIZE_OF_IPv4_HEADER >> 2 ) & 0x0FU );
+            pxIPHeader->usLength = FreeRTOS_htons( FreeRTOS_ntohs( pxIPHeader->usLength ) - optlen );
```
htibosch, line 2805: Thanks for this PR. It looks all good. I tested it with real ICMP packets on a little-endian platform. I tried all possible option sizes from 0 up to the maximum op 40 bytes ( when `ucVersionHeaderLength = 0x4f` ).

The expression to adapt `usLength` is correct, but I'd rather see multiple steps:
~~~c
    uint16_t usLength;

    usLength = FreeRTOS_ntohs( xICMPPacket->xIPHeader.usLength );
    usLength = usLength + ( uint16_t ) uxOptionsLength;
    xICMPPacket->xIPHeader.usLength = FreeRTOS_htons( usLength );
~~~
This is easier to read/debug, and MISRA will also be happier about this.

htibosch (approved):
@AniruddhaKanhere , thanks a lot for this PR. Would you mind to split up the expression in 3 expressions?
Thanks

FreeRTOS_IP.c
htibosch: Sorry, wrong code snippet. I showed the code that I used for sending IP-options.
This is the code for `prvCheckIP4HeaderOptions()`:
~~~c
    uint16_t usLength;

    usLength = FreeRTOS_ntohs( pxIPHeader->usLength );
    usLength = usLength - ( uint16_t ) optlen;
    pxIPHeader->usLength = FreeRTOS_ntohs( usLength );
~~~
And I wouldn't mind to rename and comment `optlen`:
~~~c
-    const size_t optlen = xxx
+    /* The number of bytes contained in the IP-options. */
+    const size_t xOptionsLength = xxx
~~~

htibosch (approved):

htibosch (approved):
It looks all good to me, approved, thanks!

PR #309, opened 2021-08-19
IPv6/multi: Update the TTL field of ICMP return packet

AniruddhaKanhere:
Description
-----------
According to RFC 1122 section 3.2.1.7 - any device should not send any packet with TTL 0. FreeRTOS+TCP updates the TTL of TCP and UDP packets. This PR adds such ability to the ICMP packet processing as well.

Test Steps
-----------
Send a ping with TTL 0 to the device. FreeRTOS+TCP will return the packet as is (i.e. with packet TTL unchanged - set to 0)

Related Issue
-----------


include/FreeRTOSIPConfigDefaults.h
```diff
@@ -343,6 +343,11 @@
     #define ipconfigTCP_TIME_TO_LIVE    128
 #endif
 
+#ifndef ipconfigICMP_TIME_TO_LIVE
+    /* Set the default value suggested in RFC 1700. */
+    #define ipconfigICMP_TIME_TO_LIVE    64
```
htibosch, line 348: Yes indeed, when replying to an ICMP ping request, the stack would leave the TTL field `ucTimeToLive` untouched.
Thanks for this change.

htibosch (approved):
I approve this PR, thanks.

dan4thewin (approved):

PR #310, opened 2021-08-20
IPv6/multi: Ignore link level only broadcast

AniruddhaKanhere:
Description
-----------
According to RFC 1122 section 3.3.6, link level only broadcast packets should be discarded. This PR adds functionality to check and discard such packets.

Link level only broadcast means that the Ethernet address is set to broadcast (`FF:FF:FF:FF:FF:FF`) whereas the IP address is not a broadcast address (of type `x.x.x.FF`) - it is a directed address.

Test Steps
-----------

Related Issue
-----------


FreeRTOS_IP.c
```diff
@@ -2646,6 +2646,15 @@ static eFrameProcessingResult_t prvAllowIPPacketIPv4( const IPPacket_t * const p
                 /* Packet is not for this node, release it */
                 eReturn = eReleaseBuffer;
             }
+            else if( ( memcmp( ( void * ) xBroadcastMACAddress.ucBytes,
+                               ( void * ) ( pxIPPacket->xEthernetHeader.xDestinationAddress.ucBytes ),
+                               sizeof( MACAddress_t ) ) == 0 ) &&
+                     ( ( ulSourceIPAddress & 0xff ) != 0xff ) )
```
htibosch, line 2652: Here `ulSourceIPAddress` is in network byte order, which means that you should apply ntohl:
~~~c
    ( ( FreeRTOS_ntohl( ulSourceIPAddress ) & 0xff ) != 0xff ) )
~~~
I will check first this to be sure.

htibosch: I think you wanted to test the destination address in stead of `ulSourceIPAddress`.
The line below works fine for me:
~~~c
    ( ( FreeRTOS_ntohl( ulDestinationIPAddress ) & 0xff ) != 0xff ) )
~~~
I tested this on a little-endian platform.

htibosch (approved):
If you change the code as suggested I will approve the change.
Thanks!

htibosch (approved):
Thank you Aniruddha for this PR. I retested it and it looks good.
I approve it.

PR #312, opened 2021-08-24
IPv6/multi: Adding the mDNS protocol

htibosch:
Description
-----------
FreeRTOS+TCP has implemented 3 types of name service clients/servers:
- DNS ( Domain Name System )
- LLMNR ( Link-Local Multicast Name Resolution )
- NBNS ( NetBios Name Server )

The latter two are now considered vulnerable. In all major operating systems, LLMNR and NBNS are now disabled **by default**.

This PR adds the mDNS protocol! It allows to lookup device names that end with `.local`. Also the device can be found by its name. For any incoming request, the following function is called:
~~~c
    BaseType_t xApplicationDNSQueryHook( const char *pcName );
~~~


htibosch:
Closing this PR because I created a new version: ["Adding mDNS and more changes to DNS lookups"](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/337). In that version, DNS will become independent from the use of IPv4.

PR #313, opened 2021-08-24
IPv6/multi: recognise and treat IPv6 extension headers

htibosch:
Description
-----------
When we were doing the protocol testing, it was noticed that the IPv6 branch doesn't recognise and support so-called extension headers ( see [RFC 8200](https://datatracker.ietf.org/doc/html/rfc8200) ).

This PR adds a function `eHandleIPv6ExtensionHeaders()` that removes all extension headers, so that the packets can be processed normally.

Test Steps
-----------
I tested this PR by sending packets with hop-by-hop options to the IP-stack.

**To-do**: I could only test this PR while allowing checksum offloading. The function `usGenerateProtocolChecksum()` will still drop all packets containing hop-by-hop options.


FreeRTOS_IP.c
```diff
@@ -2961,6 +3034,23 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
             eReturn = prvCheckIP4HeaderOptions( pxNetworkBuffer );
         }
 
+        #if ( ipconfigUSE_IPv6 != 0 )
+            if( ( pxIPPacket->xEthernetHeader.usFrameType == ipIPv6_FRAME_TYPE ) &&
+                ( ucProtocol == ipPROTOCOL_EXT_HEADER ) )
```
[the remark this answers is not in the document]
htibosch: Yes you are right, there are several option types that can start a chain of extension headers. And also the options come in a particular order, which must be checked.
Both are implemented in the commit "Completed the list of possible header extensions", which is now based on RFC8200 (in stead of RFC2460).

FreeRTOS_IP.c
```diff
@@ -2961,6 +3134,19 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
             eReturn = prvCheckIP4HeaderOptions( pxNetworkBuffer );
         }
 
+        #if ( ipconfigUSE_IPv6 != 0 )
+            if( ( pxIPPacket->xEthernetHeader.usFrameType == ipIPv6_FRAME_TYPE ) &&
+                ( xGetExtensionOrder( ucProtocol, 0U ) > 0 ) )
+            {
+                eReturn = eHandleIPv6ExtensionHeaders( pxNetworkBuffer, pdTRUE );
```
AniruddhaKanhere, line 3256: Maybe we need a config parameter since we are dealing with `memmove` etc. here.
It is similar to IP options which we may or may not accept based on the config value. What do you think about `ipconfigALLOW_IPv6_WITH_EXTENSIONS_HEADERS`?

include/FreeRTOS_IP.h
```diff
@@ -126,6 +126,15 @@
     #define ipICMP_NEIGHBOR_SOLICITATION_IPv6          ( ( uint8_t ) 135U )
     #define ipICMP_NEIGHBOR_ADVERTISEMENT_IPv6         ( ( uint8_t ) 136U )
 
+    #define ipIPv6_EXT_HEADER_HOP_BY_HOP               0U
```
AniruddhaKanhere, line 129: NIT: Can we arrange them in ascending order? ( 0 ->43->44->50->51->60).
I am fine with this as well as this is arranged in the order the headers should be received.

include/FreeRTOS_IP.h
```diff
@@ -126,6 +126,15 @@
...
+    #define ipIPv6_EXT_HEADER_DESTINATION_OPTIONS      60U
+    #define ipIPv6_EXT_HEADER_ROUTING_HEADER           43U
+    #define ipIPv6_EXT_HEADER_FRAGMENT_HEADER          44U
+    #define ipIPv6_EXT_HEADER_AUTHEN_HEADER            51U
+    #define ipIPv6_EXT_HEADER_SECURE_PAYLOAD           50U
+    /* Destination options may follow here in case there are no routing options. */
+    #define ipIPv6_EXT_HEADER_MOBILITY_HEADER          135U
+
```
AniruddhaKanhere, line 137: I did find a couple more options here - we might want to add them as well for completeness.
`Host Identity Protocol - 139`
`Shim6 Protocol - 140`.

What do you think @htibosch?

FreeRTOS_IP.c
```diff
@@ -3011,6 +3017,173 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
+                ( void ) memmove( pucTo, pucFrom, xMoveLen );
+                pxNetworkBuffer->xDataLength -= uxRemovedBytes;
+
+                usPayloadLength -= uxRemovedBytes;
+                pxIPPacket_IPv6->xIPHeader.usPayloadLength = FreeRTOS_htons( usPayloadLength );
+                eResult = eProcessBuffer;
+            }
+        }
+
+        FreeRTOS_printf( ( "Hop-by-hop option : %s Truncated %u bytes. Removed %u, Payload %u xDataLength now %u\n",
```
AniruddhaKanhere, line 3175:
```suggestion
        FreeRTOS_printf( ( "Extension headers: %s Truncated %u bytes. Removed %u, Payload %u xDataLength now %u\n",
```

FreeRTOS_IP.c
```diff
@@ -3011,6 +3017,173 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
+                pucTo = &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER ] );
+                pucFrom = &( pxNetworkBuffer->pucEthernetBuffer[ uxIndex ] );
+                xMoveLen = uxMaxLength - uxIndex;
+                ( void ) memmove( pucTo, pucFrom, xMoveLen );
+                pxNetworkBuffer->xDataLength -= uxRemovedBytes;
+
+                usPayloadLength -= uxRemovedBytes;
+                pxIPPacket_IPv6->xIPHeader.usPayloadLength = FreeRTOS_htons( usPayloadLength );
+                eResult = eProcessBuffer;
+            }
```
AniruddhaKanhere, line 3172: There should be an `else` clause here to be in line with guidance of MISRA. (Ref. rule 15.7)

FreeRTOS_IP.c
```diff
@@ -3011,6 +3017,173 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
+                break;
+            }
+
+            uxIndex = uxIndex + uxHopSize;
+
+            if( ( ucNextHeader == ipPROTOCOL_TCP ) ||
+                ( ucNextHeader == ipPROTOCOL_UDP ) ||
+                ( ucNextHeader == ipPROTOCOL_ICMP_IPv6 ) )
+            {
+                FreeRTOS_printf( ( "Stop at header %u\n", ucNextHeader ) );
```
AniruddhaKanhere, line 3119: I would prefer if all these are _debug_ printfs.

FreeRTOS_IP.c
```diff
@@ -2896,6 +2902,173 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
+            }
+
+            xNextOrder = xGetExtensionOrder( ucNextHeader, pucSource[ uxIndex ] );
+
+/*
+ * [IP-Task         ] Going from header  0 (1) to 60 (2)
+ * [IP-Task         ] Going from header 60 (2) to 43 (3)
+ * [IP-Task         ] Going from header 43 (3) to 44 (4)
+ * [IP-Task         ] Stop at header 58
+ */
```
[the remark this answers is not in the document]
htibosch: The comment showed some logging that I used while developing.
I removed the comments, and change `FreeRTOS_printf()` to `FreeRTOS_debug_printf()`.

FreeRTOS_IP.c
htibosch: Yes you are right. In the first draft I mistakenly though that an extension header always starts with a hop-to-hop option.
I change the logging as suggested.

FreeRTOS_IP.c
htibosch: In an earlier version there was such a config macro, but then I realised that we must process packets that carry an extension header.

include/FreeRTOS_IP.h
htibosch: I declared the options them in the order in which they appear in an extension header. I can see that in the new function `xGetExtensionOrder()`.

FreeRTOS_IP.c
htibosch: Replaced.

FreeRTOS_IP.c
htibosch: Added an else with a comment

include/FreeRTOS_IP.h
htibosch: I had also seen them but let's stick to RFC 8200 for the moment.

FreeRTOS_IP.c
```diff
@@ -2896,6 +2902,173 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
+                break;
+
+            case ipIPv6_EXT_HEADER_SECURE_PAYLOAD:
+                xReturn = 6;
+                break;
+
+            /* Destination options may follow here in case there are no routing options. */
+            case ipIPv6_EXT_HEADER_MOBILITY_HEADER:
+                xReturn = 8;
+                break;
```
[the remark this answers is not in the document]
htibosch: Added a default statement.

htibosch (commented):
Thank you for this review, @AniruddhaKanhere, good points.

AniruddhaKanhere (dismissed):

AniruddhaKanhere (approved):

aggarg (approved):

PR #316, opened 2021-08-30
Allow FreeRTOS_send being called with a NULL pointer

htibosch:
Description
-----------
The TCP function `FreeRTOS_send()` takes a pointer `pvBuffer`, which points to the data to be send.
However, when this pointer is NULL, the idea was to update only the head of the transmission stream buffer, i.e. the **zero-copy method**.
~~~c
    size_t uxLength, uxRead;
    uint8_t * pucBuffer = FreeRTOS_get_tx_head( xSocket, &( uxLength ) );
    if( pucBuffer != NULL )
    {
        uxRead = read_from_disk( pucBuffer, uxLength );
        if( uxRead > 0U )
        {
            FreeRTOS_send( xSocket, NULL, uxLength, 0 );
        }
    }
~~~
The above methods used to work well, however after a certain PR, the function would treat NULL as an error.

Test Steps
-----------
The [FTP server](https://github.com/htibosch/freertos_plus_projects/tree/master/plus/Framework/FreeRTOS-Plus-TCP/protocols) that I wrote uses this zero-copy transmissions when `ipconfigFTP_TX_ZERO_COPY` is defined as `1` in the configuration file "FreeRTOSFATConfig.h".

Related Issue
-----------
The function `FreeRTOS_send()` returned `-pdFREERTOS_ERRNO_EINVAL` when it was called with a NULL pointer. That happens in [this code](https://github.com/htibosch/freertos_plus_projects/blob/8e3ca77db6c99828ce3dddbd38a170e6a6d0ef0e/plus/Framework/FreeRTOS-Plus-TCP/protocols/FTP/FreeRTOS_FTP_server.c#L1821).

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -3382,7 +3382,9 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
  *        the socket gets connected.
  *
  * @param[in] xSocket: The socket owning the connection.
- * @param[in] pvBuffer: The buffer containing the data.
+ * @param[in] pvBuffer: The buffer containing the data. The value of this pointer
+ *                      may be NULL in case zero-copy transmissions are used.
+ *                      It is used in combination with 'FreeRTOS_get_tx_head()'.
  * @param[in] uxDataLength: The length of the data to be added.
  * @param[in] xFlags: This parameter is not used. (zero or FREERTOS_MSG_DONTWAIT).
  *
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -3394,7 +3394,7 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                               size_t uxDataLength,
                               BaseType_t xFlags )
     {
-        BaseType_t xByteCount = -pdFREERTOS_ERRNO_EINVAL;
+        BaseType_t xByteCount;
         BaseType_t xBytesLeft;
         FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
         TickType_t xRemainingTime;
@@ -3407,10 +3407,7 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
          * may be used in future versions. */
         ( void ) xFlags;
 
-        if( pvBuffer != NULL )
-        {
-            xByteCount = ( BaseType_t ) prvTCPSendCheck( pxSocket, uxDataLength );
-        }
+        xByteCount = ( BaseType_t ) prvTCPSendCheck( pxSocket, uxDataLength );
 
         if( xByteCount > 0 )
         {
@@ -3482,8 +3479,9 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
 
                     xBytesLeft -= xByteCount;
 
-                    if( xBytesLeft == 0 )
+                    if( ( xBytesLeft == 0 ) || ( pvBuffer == NULL ) )
                     {
+                        /* pvBuffer can be NULL in case TCP zero-copy transmissions are used. */
                         break;
                     }
 


FreeRTOS_Sockets.c
```diff
@@ -3482,8 +3479,9 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
 
                     xBytesLeft -= xByteCount;
 
-                    if( xBytesLeft == 0 )
+                    if( ( xBytesLeft == 0 ) || ( pvBuffer == NULL ) )
```
AniruddhaKanhere, line 3482: If this is to be only used with zero copy, why should we not enclose this in an `#if` block? Because, if we are not using zero copy method, then this should fail(?).

FreeRTOS_Sockets.c
```diff
@@ -3407,10 +3407,7 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
          * may be used in future versions. */
         ( void ) xFlags;
 
-        if( pvBuffer != NULL )
```
AniruddhaKanhere, line 3410: If this is only for the zero copy method, should we not try to make this in line with the underlying implementation of zero copy? Maybe we should enclose this in an #if block?

FreeRTOS_Sockets.c
```diff
@@ -3394,7 +3394,7 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                               size_t uxDataLength,
                               BaseType_t xFlags )
     {
-        BaseType_t xByteCount = -pdFREERTOS_ERRNO_EINVAL;
+        BaseType_t xByteCount;
```
AniruddhaKanhere, line 3397: The comment in line 3385 should also be updated to reflect that the buffer can be NULL or non-NULL and the conditions can be added in a `@note` section.

AniruddhaKanhere (commented):
Thank you @htibosch for this PR. I found the PR which you mentioned in the description: https://github.com/FreeRTOS/FreeRTOS/pull/104.

I do have some questions comment. Please see them below.

FreeRTOS_Sockets.c
htibosch: We could have introduced a new flag, but I'm afraid it is too late now. Some developers have been using this feature for many years. If we add a macro `ipconfig_TCP_SEND_ALLOWS_ZERO_COPY` it should be **enabled** by default, in order not to break existing code.

Background: as you know, TCP sockets have 2 stream buffers: one for incoming and one for outgoing. This has advantages, but it makes it impossible to create a zero-copy feature like UDP sockets have. Zero-copy UDP sockets use network buffers, zero-copy TCP sockets use pointers to the stream buffers.

The zero-copy feature in TCP has always been classified as "for advanced users only".

FreeRTOS_Sockets.c
htibosch: Here the same answer as above: developers are already using it.

FreeRTOS_Sockets.c
htibosch: Yes, that we can do, I am proposing:
~~~c
/*
 * @param[in] pvBuffer: The buffer containing the data. The value of this pointer
 *                      may be NULL in case zero-copy transmissions are used.
 *                      It is used in combination with 'FreeRTOS_get_tx_head()'.
 */
~~~

AniruddhaKanhere: Yes, that would be very helpful, thank you.

htibosch: > Yes, that would be very helpful, thank you.

I added the comment, and also in the IPv6/multi version of this PR: #317

AniruddhaKanhere: Hello! do you mind pushing the changes so that we can get this merged into the repository?

Thanks :)

htibosch: It got stuck in an unpushed commit. Now it it there.

AniruddhaKanhere (approved):

n9wxu (approved):

PR #318, opened 2021-08-30
Send ARP requests for unknown IP address requests.

AniruddhaKanhere:
Description
-----------
The IP stack should not populate its ARP cache with received datagrams and should only use ARP to create a connection between a MAC address and an IP-address.

Ref. RFC 1122 sec 2.3.3: _"Address translation from Internet addresses to link-layer addresses on Ethernet and IEEE 802 networks MUST be managed by the Address Resolution Protocol (ARP)."_

Test Steps
-----------

Related Issue
-----------


FreeRTOS_IP.c
```diff
@@ -3097,7 +3097,24 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
...
+
+                    if( ( pxIPHeader->ulSourceIPAddress & pxIPv4Settings->ulNetMask ) == ( pxIPv4Settings->ulIPAddress & pxIPv4Settings->ulNetMask ) )
+                    {
+                        /* If the IP is on the same subnet and we do not have an ARP entry already,
+                         * then we should send out ARP for finding the MAC address. */
+                        if( xIsIPInARPCache( pxIPHeader->ulSourceIPAddress ) == pdFALSE )
+                        {
+                            FreeRTOS_OutputARPRequest( pxIPHeader->ulSourceIPAddress );
+                            /* Drop this packet. */
+                            eReturn = eReleaseBuffer;
```
htibosch, line 3110: Here either a TCP or an ICMP packet was received. It will be dropped because the sender's IPv4 address does not occur in the ARP cache yet. Which is probably good and correct. I would like to test if it will happen often that packets are dropped here, and how it affects existing projects.

Note that every entry in the ARP cache is checked regularly by the function `vARPAgeCache()`. Entries that are about to expire will send an ARP request.

htibosch (commented):
Hi Aniruddha, thanks for this PR. The way ARP is handled and updated has a long history.
I think that we can adapt to without disturbing existing projects, but I will test that first.
Note that we should also do the same with the UDP packets. Shall I write a proposal for this?

AniruddhaKanhere:
@htibosch,
Thank you for the review.
I am ready with the UDP changes as well. I haven't pushed them yet. I shall commit the changes soon.

Yes, I was also concerned about the unnecessary ARP Packets being sent. I am curious to see the results you get from your experiment.

Thanks

htibosch:
In the PR you set `eReturn = eReleaseBuffer`, but you forget to test `eReturn` later on. After I changed that I saw nasty time-outs:
~~~
ping 192.168.2.114

Pinging 192.168.2.114 with 32 bytes of data:
Request timed out.
Reply from 192.168.2.114: bytes=32 time<1ms TTL=64
~~~
and I saw the same while making a TCP connection: the first SYN times out and retries after 1 second.
Once my laptop's IP-address is stored in the ARP table, all is fine.

While testing your PR I had to repair a problem in FreeRTOS_ARP.c, which caused double entries (see #319).

I will continue find a solution for the above time-out problems.

FreeRTOS_ARP.c
```diff
@@ -305,9 +308,42 @@ static void vARPProcessPacketReply( ARPPacket_t * pxARPFrame,
...
+
+            /* Clear the buffer. */
+            pxARPWaitingNetworkBuffer = NULL;
+            uxCount = 0;
+        }
+        else
+        {
+            uxCount++;
+
+            if( uxCount > 5 )
```
AniruddhaKanhere, line 339: The value `5` has been chosen arbitrarily. Should make this into a define.

FreeRTOS_ARP.c
```diff
@@ -305,9 +308,42 @@ static void vARPProcessPacketReply( ARPPacket_t * pxARPFrame,
...
+
+        if( ulSenderProtocolAddress == pxIPHeader->ulSourceIPAddress )
+        {
+            IPStackEvent_t xEventMessage;
+            const TickType_t xDontBlock = ( TickType_t ) 0;
+
+            xEventMessage.eEventType = eNetworkRxEvent;
+            xEventMessage.pvData = ( void * ) pxARPWaitingNetworkBuffer;
+
+            xSendEventStructToIPTask( &xEventMessage, xDontBlock );
```
htibosch, line 329: That is a great solution! The IP-task will keep the network buffer until a matching ARP request comes in. At that moment, the message will be processed.

I tested it again, and the PCAP looks like expected. Here is a PCAP: [icmp_test.zip](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/7088483/icmp_test.zip). The delayed reply comes immediately after the ARP messages.

This is very simple:
~~~c
    uxCount++;
    if( uxCount > 5 )
    {
        /* Release the network buffer. */
    }
~~~
Maybe an alarm clock would be nicer? There is a simple light-weight timer `IPTimer_t`, like e.g. :
~~~c
    /** @brief ARP timer, to check its table entries. */
    static IPTimer_t xARPTimer;

    /* Reload the timer. */
    prvIPTimerReload( &xARPTimer, pdMS_TO_TICKS( ipARP_TIMER_PERIOD_MS ) );

   /* Check if it expires. */
    if( prvIPTimerCheck( &xARPTimer ) != pdFALSE )
    {
        ( void ) xSendEventToIPTask( eARPTimerEvent );
    }

    /* See how long the IP-task may sleep. */
    if( xARPTimer.bActive != pdFALSE_UNSIGNED )
    {
        if( xARPTimer.ulRemainingTime < xMaximumSleepTime )
        {
            xMaximumSleepTime = xARPTimer.ulRemainingTime;
        }
    }
~~~

AniruddhaKanhere: Yes, I was thinking along the same lines. I was going to add this along with a timer or create a new one which would be checked by the IP task.
I shall add more commits soon (tomorrow)

FreeRTOS_ARP.c
```diff
@@ -308,6 +311,29 @@ static void vARPProcessPacketReply( ARPPacket_t * pxARPFrame,
...
```
htibosch, line 328: Oops, the return value should be checked:
~~~c
-    xSendEventStructToIPTask( &xEventMessage, xDontBlock )
+    if( xSendEventStructToIPTask( &xEventMessage, xDontBlock ) != pdPASS )
     {
         /* Failed to send the message, so release the network buffer. */
         vReleaseNetworkBufferAndDescriptor( pxARPWaitingNetworkBuffer );
     }

    /* Clear the buffer. */
    pxARPWaitingNetworkBuffer = NULL;
~~~

FreeRTOS_IP.c
```diff
@@ -1009,6 +1025,18 @@ static void prvCheckNetworkTimers( void )
         ( void ) xSendEventToIPTask( eARPTimerEvent );
     }
 
+    /* Is the ARP resolution timer expired? */
+    if( prvIPTimerCheck( &xARPResolutionTimer ) != pdFALSE )
+    {
+        if( pxARPWaitingNetworkBuffer != NULL )
```
htibosch, line 1031: These timers are auto-repeating. Let's switch it off when it expires:
~~~c
     /* Is the ARP resolution timer expired? */
     if( prvIPTimerCheck( &xARPResolutionTimer ) != pdFALSE )
     {
+        vIPSetARPResolutionTimerEnableState( pdFALSE );
+
         if( pxARPWaitingNetworkBuffer != NULL )
~~~

htibosch (approved):
I approve this PR, thanks a lot.
I tested it again with real hardware. I also simulated a time-out on the ARP request. The network buffer `pxARPWaitingNetworkBuffer` got released in `prvCheckNetworkTimers()` after exactly 2 seconds.
Note that there is only one IP-task, so there are no race conditions among all these actions.

FreeRTOS_IP.c
AniruddhaKanhere: Yes, done. Thank you.

FreeRTOS_ARP.c
AniruddhaKanhere: Oh yes! Nice catch!
Thank you.

FreeRTOS_ARP.c
AniruddhaKanhere: I have added more commits and added UDP processing as well. Thank you.

htibosch (dismissed):
Beside 2 suggestions, it all looks good to me now, thank you!

FreeRTOS_ARP.c
```diff
@@ -62,6 +62,9 @@
     #define arpGRATUITOUS_ARP_PERIOD    ( pdMS_TO_TICKS( 20000U ) )
 #endif
 
+/** @brief The pointer to buffer with packet waiting for ARP resolution. */
+extern NetworkBufferDescriptor_t * pxARPWaitingNetworkBuffer;
```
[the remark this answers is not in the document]
htibosch: PS. I have tested your branch AniruddhaKanhere:SendARPForUnknownIPs in real hardware. It all still worked as expected.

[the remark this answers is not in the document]
AniruddhaKanhere: Done, thank you.

FreeRTOS_IP.c
```diff
@@ -1009,6 +1025,23 @@ static void prvCheckNetworkTimers( void )
...
+        {
+            /* Disable the ARP resolution timer. */
+            vIPSetARPResolutionTimerEnableState( pdFALSE );
+
```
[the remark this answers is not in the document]
AniruddhaKanhere: I have added the macros to apt locations. Thanks.

AniruddhaKanhere (commented):
Thanks for the review @htibosch. I have added your suggested changes to the PR.

n9wxu (approved):

PR #338, opened 2021-09-17
Send ARP packets for unknown IP addresses

AniruddhaKanhere:
Description
-----------
The IP stack should not populate its ARP cache with received datagrams and should only use ARP to create a connection between a MAC address and an IP-address.

Ref. RFC 1122 sec 2.3.3: "Address translation from Internet addresses to link-layer addresses on Ethernet and IEEE 802 networks MUST be managed by the Address Resolution Protocol (ARP)."

This is based on an earlier PR https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/318.

Test Steps
-----------
This can be tested by sending an ICMP ping to the FreeRTOS+TCP device from a device on local subnet. After this change, the unknown IP address will be queried through ARP first and then once resolved - the ICMP reply will be sent by the FreeRTOS+TCP stack.

Related Issue
-----------


htibosch:
~~~
Description
Test Steps
Related Issue
~~~
It would be nice to read some comments in these headers. At least you could indicate that this PR is based on an earlier PR #318, and copy some of the earlier comments?

AniruddhaKanhere:
> ```
> Description
> Test Steps
> Related Issue
> ```
> 
> It would be nice to read some comments in these headers. At least you could indicate that this PR is based on an earlier PR #318, and copy some of the earlier comments?

Of course. A lot of tests/checks were broken - I thought I'll fix them and update the description later.
But, yes, I can add description etc. Thanks.

cobusve (approved):

PR #343, opened 2021-09-20
IPv4/single: Adding EUI48_pton and EUI48_ntop for MAC address conversion

htibosch:
Description
-----------
It is a copy of #342, except that this PR is aiming at the IPv4 main branch.It adds two functions:
~~~c
/** @brief This funtion converts a 48-bit MAC address to a human readable string. */

    void FreeRTOS_EUI48_ntop( const uint8_t * pucSource,
                              char * pcTarget,
                              char cTen,
                              char cSeparator );

/** @brief This funtion converts a human readable string, representing an 48-bit MAC address,
 * into a 6-byte address. Valid inputs are e.g. "62:48:5:83:A0:b2" and "0-12-34-fe-dc-ba". */

    BaseType_t FreeRTOS_EUI48_pton( const char * pcSource,
                                    uint8_t * pucTarget );
~~~

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -178,6 +178,8 @@ static BaseType_t prvDetermineSocketSize( BaseType_t xDomain,
                                           BaseType_t xProtocol,
                                           size_t * pxSocketSize );
 
+static uint8_t ucASCIIToHex( char cChar );
+
 #if ( ipconfigUSE_TCP == 1 )
 
 /*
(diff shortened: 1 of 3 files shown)


FreeRTOS_Sockets.c
```diff
@@ -2429,6 +2436,183 @@ const char * FreeRTOS_inet_ntop4( const void * pvSource,
...
+        }
+    }
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief This function converts a human readable string, representing an 48-bit MAC address,
+ *        into a 6-byte address. Valid inputs are e.g. "62:48:5:83:A0:b2" and "0-12-34-fe-dc-ba".
+ *
+ * @param[in] pcSource: The string to be parsed.
```
AniruddhaKanhere, line 2549:
```suggestion
 * @param[in] pcSource: The NULL terminated string to be parsed.
```

FreeRTOS_Sockets.c
```diff
@@ -2429,6 +2436,183 @@ const char * FreeRTOS_inet_ntop4( const void * pvSource,
...
+                /* A hex value was too big. */
+                break;
+            }
+
+            uxSum |= ucHex;
+            uxNibbleCount++;
+        }
+        else
+        {
+            if( uxNibbleCount < 1U )
```
AniruddhaKanhere, line 2587: Maybe we should make this more strict? Such that all bytes in the output must constitute of 2 characters.

I am suggesting this as this function will give out `uint8_t MAC = {1,2,3,4,5,6}` when it is passed a string as `"1:2:3:4:5:6"`.
But when the MAC[] is fed to the ntop function, it will give a string as `"01:02:03:04:05:06"` which is not same as the original string.

AniruddhaKanhere (commented):
Hello @htibosch,
Thank you for this PR. Would you mind taking a look at the suggestions/comments that I have?

FreeRTOS_Sockets.c
htibosch: I had the same thoughts when writing it. I just found a [Wiki page](https://en.wikipedia.org/wiki/MAC_address) saying: "As typically represented, MAC addresses are recognizable as six groups of two hexadecimal digits, separated by hyphens, colons, or without a separator.".

So yes, it would be OK to force the use of 2 hex digits. And this would also make sure that a string can be converted to an address, and vv without changing the contents.

htibosch (commented):
Thanks for this review. I will make the same changes in the IPv6/multi PR.

AniruddhaKanhere (approved):

n9wxu (approved):

PR #344, opened 2021-09-21
Add Checks in ARP processing to filter out invalid packets

AniruddhaKanhere:
Description
-----------
This PR adds checks in ARP processing function to filter out ARP packets with incorrect IP/MAC addresses.
Ref: RFC 1812 section 3.3.2.

Test Steps
-----------

Related Issue
-----------


htibosch (commented):
Thank you for these ARP changes. I have had a chance to test it yet, I'll come back to it after the tests.

FreeRTOS_ARP.c
```diff
@@ -119,27 +118,75 @@ eFrameProcessingResult_t eARPProcessPacket( ARPPacket_t * const pxARPFrame )
...
+            break;
+        }
+
+        /* Check whether there is a clash with another device for this IP address. */
+        if( ( ulSenderProtocolAddress == *ipLOCAL_IP_ADDRESS_POINTER ) &&
+            ( *ipLOCAL_IP_ADDRESS_POINTER != 0UL ) )
+        {
+            /* There is a clash with another device. Send out ARP request to show
+             * that there is a clash. */
+            FreeRTOS_OutputARPRequest( *ipLOCAL_IP_ADDRESS_POINTER );
```
htibosch, line 177: Can this lead to a continuous bouncing of ARP requests between this device and the other device with the same IP-address?

htibosch (approved):
I have one last question about sending an ARP request in case of an IP-conflict.
Otherwise it looks all good to me, thanks a lot.

FreeRTOS_ARP.c
```diff
@@ -152,8 +245,12 @@ eFrameProcessingResult_t eARPProcessPacket( ARPPacket_t * const pxARPFrame )
...
 
                     /* The packet contained an ARP request.  Was it for the IP
-                     * address of the node running this code? */
-                    if( ulTargetProtocolAddress == *ipLOCAL_IP_ADDRESS_POINTER )
+                     * address of the node running this code? And does the MAC
+                     * address claim that it is coming from this device itself? */
+                    if( ( ulTargetProtocolAddress == *ipLOCAL_IP_ADDRESS_POINTER ) &&
+                        ( memcmp( ( void * ) ipLOCAL_MAC_ADDRESS,
+                                  ( void * ) ( pxARPHeader->xSenderHardwareAddress.ucBytes ),
+                                  sizeof( MACAddress_t ) ) != 0 ) )
```
yourslab, line 253: It's an interesting case because `ucBytes` is the only member in the struct so it's unlikely to have [alignment issues](https://stackoverflow.com/questions/3585033/size-of-struct-with-a-single-element). However, it is not necessarily guaranteed, so feel like it would be best to use `sizeof(pxARPHeader->xSenderHardwareAddress.ucBytes)` instead.

FreeRTOS_ARP.c
```diff
@@ -119,27 +136,103 @@ eFrameProcessingResult_t eARPProcessPacket( ARPPacket_t * const pxARPFrame )
...
+            if( uxARPClashCounter < arpIP_CLASH_MAX_RETRIES )
+            {
+                /* Increment the counter. */
+                uxARPClashCounter++;
+
+                /* Send out a defensive ARP request. */
+                FreeRTOS_OutputARPRequest( *ipLOCAL_IP_ADDRESS_POINTER );
+
+                /* Since an ARP Request for this IP was just sent, do not send a gratuitous
+                 * APR for arpGRATUITOUS_ARP_PERIOD. */
```
yourslab, line 216:
```suggestion
                 * ARP for arpGRATUITOUS_ARP_PERIOD. */
```

FreeRTOS_ARP.c
```diff
@@ -61,10 +61,19 @@
...
 
+/** @brief When there is another device which has the same IP address as the IP address
+ * of this device, a defensive ARP request should be sent out. However, according to
+ * RFC 5227 section 1.1, there must be a minimum interval of 10 seconds between
+ * consecutive defensive ARP packets. */
+#define arpIP_CLASH_RESET_TIMEOUT    10000
+
+/** @brief Maximum number of defensive ARPs to be sent for an ARP clash per
+ * arpIP_CLASH_RESET_TIMEOUT period. */
+#define arpIP_CLASH_MAX_RETRIES      1
```
yourslab, line 72: Should both of these macros be wrapped in `#ifndef`..`#endif` so that they can be configured? Why is 1 the best default?

FreeRTOS_ARP.c
```diff
@@ -85,7 +94,7 @@ _static ARPCacheRow_t xARPCache[ ipconfigARP_CACHE_ENTRIES ];
 
 /** @brief  The time at which the last gratuitous ARP was sent.  Gratuitous ARPs are used
  * to ensure ARP tables are up to date and to detect IP address conflicts. */
-static TickType_t xLastGratuitousARPTime = ( TickType_t ) 0;
+static TickType_t xLastGratuitousARPTime = 0U;
```
yourslab, line 97: Why did you remove the cast?

FreeRTOS_ARP.c
htibosch: In most cases in the library, `ipMAC_ADDRESS_LENGTH_BYTES` is used, which is defined as 6.
All struct's that describe network data are packed, so `MACAddress_t` and `MACAddress_t::ucBytes` are both 6 bytes long.
When some compiler doesn't pack the struct's, the library won't work at all.

FreeRTOS_ARP.c
htibosch: Indeed, that would be nice.
As you may have seen, there are two types of macros:
- E.g. `ipconfigSOCK_DEFAULT_RECEIVE_BLOCK_TIME` which is a global macro, it gets a default value in FreeRTOSIPConfigDefaults.h.
- E.g. `arpGRATUITOUS_ARP_PERIOD`, which is local to FreeRTOS_ARP.c. it is more an expert option but can be overwritten as well.

So I also vote for `#ifndef` `#endif`.

@yourslab wrote:
> Why is 1 the best default?

That used to be the default, I think?

FreeRTOS_ARP.c
htibosch: That was my suggestion. `TickType_t` is an unsigned type, so it is better to use `0U`. When written as `0U` no further cast is needed, neither for MISRA, nor for a compiler.
I did the same thing when working on MISRA compliance.

htibosch (commented):
Although I was not the composer of this PR, thank you @yourslab for your review.
I will write my own review later today.

FreeRTOS_ARP.c
AniruddhaKanhere: I agree that we should wrap these in `#ifndef`...`#endif`.

1 is the values suggested by `RFC 5227 section 2.4 part (b)` [here](https://datatracker.ietf.org/doc/html/rfc5227#section-2.4). I shall update my comment to reflect why was this value chosen.

It also suggests that if there are more conflicts in the DEFEND_INTERVAL, we should cease to use the IP address immediately - however, that is quite complex to do - considering we might have ongoing TCP connections and we don't support ICMP packets except request and response.

FreeRTOS_ARP.c
AniruddhaKanhere: Done, thanks!

FreeRTOS_ARP.c
AniruddhaKanhere: Yes, as Hein said, if a compiler doesn't pack the structures - the whole library will not work.
Also, I shall modify the number of bytes to `ipMAC_ADDRESS_LENGTH_BYTES`.

FreeRTOS_ARP.c
```diff
@@ -119,27 +141,103 @@ eFrameProcessingResult_t eARPProcessPacket( ARPPacket_t * const pxARPFrame )
...
-        pvCopyDest = &ulSenderProtocolAddress;
-        ( void ) memcpy( pvCopyDest, pvCopySource, sizeof( ulSenderProtocolAddress ) );
-        /* The field ulTargetProtocolAddress is well-aligned, a 32-bits copy. */
-        ulTargetProtocolAddress = pxARPHeader->ulTargetProtocolAddress;
+        /* Next defensive request must not be sent for arpIP_CLASH_RESET_TIMEOUT_MS
+         * period. */
+        TickType_t xARPClashTimeout = pdMS_TO_TICKS( arpIP_CLASH_RESET_TIMEOUT_MS );
+
+        /* Has the timeout been reached? */
+        if( xTaskCheckForTimeOut( &ARPClashTimeOut, &xARPClashTimeout ) == pdTRUE )
```
dan4thewin, line 164: Could xTaskCheckForTimeOut() be called before vTaskSetTimeOutState()?

FreeRTOS_ARP.c
```diff
@@ -119,27 +141,103 @@ eFrameProcessingResult_t eARPProcessPacket( ARPPacket_t * const pxARPFrame )
...
+            break;
+        }
+
+        /* Check whether the lowest bit of the highest byte is 1 to check for
+         * multicast address or even a broadcast address (FF:FF:FF:FF:FF:FF). */
+        if( ( pxARPHeader->xSenderHardwareAddress.ucBytes[ 0 ] & 0x01U ) == 0x01U )
+        {
+            /* Senders address is a multicast OR broadcast address which is not
+             * allowed for an ARP packet. Drop the packet. See RFC 1812 section
+             * 3.3.2. */
```
dan4thewin, line 194: Should we have the trace call here too? (and in the following bad packet cases?)

FreeRTOS_ARP.c
```diff
@@ -119,27 +141,103 @@ eFrameProcessingResult_t eARPProcessPacket( ARPPacket_t * const pxARPFrame )
...
+                uxARPClashCounter++;
+
+                /* Send out a defensive ARP request. */
+                FreeRTOS_OutputARPRequest( *ipLOCAL_IP_ADDRESS_POINTER );
+
+                /* Since an ARP Request for this IP was just sent, do not send a gratuitous
+                 * ARP for arpGRATUITOUS_ARP_PERIOD. */
+                xLastGratuitousARPTime = xTaskGetTickCount();
+
+                FreeRTOS_printf( ( "First one at %u\n", xLastGratuitousARPTime ) );
```
dan4thewin, line 224: Is the customer supposed to see and understand this message?

FreeRTOS_ARP.c
AniruddhaKanhere: I don't think so. However that condition will occur since `uxARPClashCounter` will never be non-zero unless `vTaskSetTimeOutState` has been already called.
Also, since this is in the same function - there is no race condition.

FreeRTOS_ARP.c
AniruddhaKanhere: Ah! Thanks for catching this. It is a debugging message I added for my aid - I shall remove that.

FreeRTOS_ARP.c
AniruddhaKanhere: Yes, added. Thanks.

AniruddhaKanhere (commented):
Thanks @dan4thewin for the review, I have made the changes you suggested and answered your questions/concerns below.

yourslab (dismissed):

FreeRTOS_ARP.c
```diff
@@ -119,27 +145,102 @@ eFrameProcessingResult_t eARPProcessPacket( ARPPacket_t * const pxARPFrame )
...
+
+                /* Since an ARP Request for this IP was just sent, do not send a gratuitous
+                 * ARP for arpGRATUITOUS_ARP_PERIOD. */
+                xLastGratuitousARPTime = xTaskGetTickCount();
+
+                /* Note the time at which this request was sent. */
+                vTaskSetTimeOutState( &xARPClashTimeOut );
+
+                /* Reset the time-out period to the given value. */
+                xARPClashTimeoutPeriod = pdMS_TO_TICKS( arpIP_CLASH_RESET_TIMEOUT_MS );
```
yourslab, line 230: Shouldn't `xARPClashTimeOut` be set to `0` so that its members are also reset?

AniruddhaKanhere: Not actually, it should be set to the `arpIP_CLASH_RESET_TIMEOUT_MS` value so that up in line 164, the value can be checked. Whenever the value is checked in `xTaskCheckForTimeOut`, the value of `xARPClashTimeoutPeriod` gets changed (reduced).
What reset means in this context is to set the value back to its original value.

FreeRTOS_ARP.c
```diff
@@ -99,6 +113,18 @@ static TickType_t xLastGratuitousARPTime = ( TickType_t ) 0;
...
+ * also to limit the requests to arpIP_CLASH_MAX_RETRIES per arpIP_CLASH_RESET_TIMEOUT_MS
+ * period. */
+static UBaseType_t uxARPClashCounter = 0;
+
+/** @brief The time at which the last ARP clash was sent. */
+static TimeOut_t xARPClashTimeOut;
+
+/** @brief Next defensive request must not be sent for arpIP_CLASH_RESET_TIMEOUT_MS
+ * period. */
+static TickType_t xARPClashTimeoutPeriod = pdMS_TO_TICKS( arpIP_CLASH_RESET_TIMEOUT_MS );
```
htibosch, line 126: `xARPClashTimeoutPeriod` is missing a `U` **prefix** because `TickType_t` is always an unsigned number.

FreeRTOS_ARP.c
```diff
@@ -99,6 +113,18 @@ static TickType_t xLastGratuitousARPTime = ( TickType_t ) 0;
     MACAddress_t xARPClashMacAddress;
 #endif /* ipconfigARP_USE_CLASH_DETECTION */
 
+/** @brief This local variable is used to keep track of number of ARP requests sent and
+ * also to limit the requests to arpIP_CLASH_MAX_RETRIES per arpIP_CLASH_RESET_TIMEOUT_MS
+ * period. */
+static UBaseType_t uxARPClashCounter = 0;
```
htibosch, line 119: Maybe write zero with a `U` suffix?
The unsigned suffix is missing in 3 locations related to `uxARPClashCounter`. Also I would suggest to define `arpIP_CLASH_MAX_RETRIES` as `1U` to avoid an extra cast in
    `if( uxARPClashCounter < arpIP_CLASH_MAX_RETRIES )`.

PR #347, opened 2021-09-25
IPv6/multi: disable counter interrupts in the STM32Hxx driver

htibosch:
Description
-----------
This PR is a copy of #346, except that is aiming at the [IPv6/multi branch](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi).

The STM32Hxx EMAC has 10 counters that generate an interrupt when they are half-full or full. These interrupts are enabled by default.

Test Steps
-----------
With the following instructions, all counters are given a value of almost half-full:
~~~c
    ETH->MMCCR &= ~( ETH_MMCCR_CNTPRSTLVL );
    ETH->MMCCR |= ETH_MMCCR_CNTPRST;
~~~
Ping the device and one counter ( `ETH->MMCRUPGR` or `ETH_RX_UNICAST_PACKETS_GOOD` ) will trigger an interrupt that is not reset, unless this PR is applied.

Related Issue
-----------

### DIFF
--- a/portable/NetworkInterface/STM32Hxx/stm32hxx_hal_eth.c
+++ b/portable/NetworkInterface/STM32Hxx/stm32hxx_hal_eth.c
@@ -429,6 +429,26 @@ extern SemaphoreHandle_t xTXDescriptorSemaphore;
             heth->gState = HAL_ETH_STATE_READY;
             heth->RxState = HAL_ETH_STATE_READY;
 
+            /*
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (approved):
Thank you Hein for these PRs (this and 346).

n9wxu (approved):

PR #348, opened 2021-09-26
IPv4/single: applied MISRA to FreeRTOS_DHCP.c

htibosch:
Description
-----------

This PR presents some related change for FreeRTOS_DHCP.c.


htibosch:
Now closing "IPv4_single_MISRA_FreeeRTOS_DHCP.c" because it is reviewed.
The sources will be merged from the joined PR #357 "IPv4/single: keep all sources MISRA compliant".

PR #349, opened 2021-09-26
IPv4/single: applied MISRA to FreeRTOS_DNS.c

htibosch:

This PR presents some related change for FreeRTOS_DNS.c.


AniruddhaKanhere (approved):
Thank you for this too! I think these look good.

You can close these PRs - since you'd like to create just one PR - which I agree with.

htibosch:
Now closing "IPv4_single_MISRA_FreeeRTOS_DNS.c" because it is reviewed.
The sources will be merged from the joined PR #357 "IPv4/single: keep all sources MISRA compliant".

PR #350, opened 2021-09-27
IPv4/single: applied MISRA to FreeRTOS_UDP_IP.c

htibosch:

This PR presents some related change for FreeRTOS_UDP_IP.c.

### DIFF
--- a/include/FreeRTOS_IP.h
+++ b/include/FreeRTOS_IP.h
@@ -70,8 +70,8 @@
                                                         uint16_t usDestinationPort );
 
 /* The number of octets in the MAC and IP addresses respectively. */
-    #define ipMAC_ADDRESS_LENGTH_BYTES                 ( 6 )
-    #define ipIP_ADDRESS_LENGTH_BYTES                  ( 4 )
+    #define ipMAC_ADDRESS_LENGTH_BYTES                 ( 6U )
+    #define ipIP_ADDRESS_LENGTH_BYTES                  ( 4U )
 
 /* IP protocol definitions. */
     #define ipPROTOCOL_ICMP                            ( 1U )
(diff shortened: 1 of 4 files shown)


include/FreeRTOS_IP_Private.h
```diff
@@ -48,16 +48,18 @@
 /*-----------------------------------------------------------*/
 /* Utility macros for marking casts as recognized during     */
 /* static analysis.                                          */
+/* Note _HT_ Changed 'vCastConstPointerTo' to the shorter    */
+/* vCastConstPtrTo to limit the length of the function name. */
```
AniruddhaKanhere, line 52: Why is name length required to be limited?

include/FreeRTOS_IP_Private.h
```diff
@@ -48,16 +48,18 @@
...
 /*-----------------------------------------------------------*/
     #define ipCAST_PTR_TO_TYPE_PTR( TYPE, pointer )                ( vCastPointerTo_ ## TYPE( ( void * ) ( pointer ) ) )
-    #define ipCAST_CONST_PTR_TO_CONST_TYPE_PTR( TYPE, pointer )    ( vCastConstPointerTo_ ## TYPE( ( const void * ) ( pointer ) ) )
+    #define ipCAST_CONST_PTR_TO_CONST_TYPE_PTR( TYPE, pointer )    ( vCastConstPtrTo_ ## TYPE( ( const void * ) ( pointer ) ) )
```
AniruddhaKanhere, line 55: Same question.

include/FreeRTOS_IP.h
```diff
@@ -350,6 +361,11 @@
  * be defined in a user module. */
     extern BaseType_t xApplicationGetRandomNumber( uint32_t * pulNumber );
```
AniruddhaKanhere, line 362: I don't think extern is required
```suggestion
    BaseType_t xApplicationGetRandomNumber( uint32_t * pulNumber );
```

include/FreeRTOS_IP_Private.h
htibosch: MISRA likes to make sure that all identifiers are still unique if only the first 32 characters are used. This is true for both macros, variables and function names.

One example is `vCastConstPointerTo_ICMPRouterAdvertisement_IPv6_t()` with 50 characters.

AniruddhaKanhere (commented):
Thank you @htibosch for this PR.
If these changes are also merged in the single all encompassing PR, then I think we can close this.

htibosch:
The changes proposed in this PR also appear in PR #357, and they will be merged to main from there.
Closing this PR now.

PR #352, opened 2021-09-27
IPv4/single: applied MISRA to FreeRTOS_IP.c

htibosch:

This PR presents some related change for FreeRTOS_IP.c.

### DIFF
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -69,8 +69,8 @@
 #define ipICMP_ECHO_REPLY                   ( ( uint8_t ) 0 ) /**< ICMP echo reply. */
 
 /* IPv4 multi-cast addresses range from 224.0.0.0.0 to 240.0.0.0. */
-#define ipFIRST_MULTI_CAST_IPv4             0xE0000000UL /**< Lower bound of the IPv4 multicast address. */
-#define ipLAST_MULTI_CAST_IPv4              0xF0000000UL /**< Higher bound of the IPv4 multicast address. */
+#define ipFIRST_MULTI_CAST_IPv4             0xE0000000U /**< Lower bound of the IPv4 multicast address. */
+#define ipLAST_MULTI_CAST_IPv4              0xF0000000U /**< Higher bound of the IPv4 multicast address. */
 
 /* The first byte in the IPv4 header combines the IP version (4) with
  * with the length of the IP header. */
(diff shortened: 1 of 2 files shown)


FreeRTOS_IP.c
```diff
@@ -1212,14 +1202,13 @@ BaseType_t FreeRTOS_IPInit( const uint8_t ucIPAddress[ ipIP_ADDRESS_LENGTH_BYTES
     configASSERT( xNetworkEventQueue == NULL );
     configASSERT( xIPTaskHandle == NULL );
 
-    if( sizeof( uintptr_t ) == 8 )
     {
```
AniruddhaKanhere, line 1205: Why not get rid of these curly braces?

FreeRTOS_IP.c
```diff
@@ -3677,6 +3610,19 @@ uint32_t FreeRTOS_max_uint32( uint32_t a,
 }
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Get the highest value of two variables of type size_t.
+ * @param[in] a: the first value.
+ * @param[in] b: the second value.
+ * @return The highest of the two values.
+ */
+size_t FreeRTOS_max_size_t( size_t a,
```
AniruddhaKanhere, line 3619: maybe we should make these inline?

FreeRTOS_IP.c
```diff
@@ -3703,6 +3649,19 @@ uint32_t FreeRTOS_min_uint32( uint32_t a,
 }
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Get the lowest value of two variables of type size_t.
+ * @param[in] a: the first value.
+ * @param[in] b: the second value.
+ * @return The lowest of the two values.
+ */
+size_t FreeRTOS_min_size_t( size_t a,
```
AniruddhaKanhere, line 3658: maybe we should make these inline?

FreeRTOS_IP.c
htibosch: Indeed, they don't serve a purpose, I removed them in #357.

FreeRTOS_IP.c
htibosch: These functions used to have both a macro and an inline version, depending on whether `portINLINE` is defined. When we started working on MISRA compliance, we decided to make all these small helper functions global.

FreeRTOS_IP.c
htibosch: See my earlier comment.

htibosch:
The changes proposed in this PR also appear in PR #357, and they will be merged to main from there.
Closing this PR now.

PR #361, opened 2021-10-13
IPv4/single: updating NetworkInterface sources

htibosch:
Description
-----------
Some of the latest updates of the NetworkInterface sources:

- Updated copy right notices and release versions (V2.3.4)
- DriverSAM: variable `uxCurrentSemCount` was not defined
- STM32Fxx : added support for STM32F1 and STM32F2
- phyHandling : added support for 2 more PHY's (DP83TC811S and MV88E6071).
- Removed definition of `ipFRAGMENT_OFFSET_BIT_MASK` from STM32Fxx/NetworkInterface.c

Test Steps
-----------
The changes were tested on an STM32F4x and a on a SAME70 board.
Also users [on the FreeRTOS forum](https://forums.freertos.org/t/i-have-heavy-ping-lose-rate-on-my-stm32f107-board-with-freertos-tcp/11178) helped with testing.


htibosch:
I just noticed that I have 2 open PR's that affect the STM32Fxx driver.
The other PR is #370, it removed the definition of `ipFRAGMENT_OFFSET_BIT_MASK` in STM32Fxx/NetworkInterface.c, because it is already defined in FreeRTOS_IP_Private.h.
I will merge the other PR with this one so it can be closed without merging it.

htibosch:
Aniruddha, reviewers: I have 2 open PR's that affect the STM32Fxx driver.

#361 IPv4/single: updating NetworkInterface sources
#372 IPv4/single: Added text to readme.txt for the STM32Fxx driver

I would recommend to merge them in their order of creation: first 361 and then 372.
Thanks

AniruddhaKanhere (approved):

portable/NetworkInterface/Common/phyHandling.c
```diff
@@ -1,4 +1,30 @@
...
+ *
+ * http://aws.amazon.com/freertos
+ * http://www.FreeRTOS.org
+ */
+
+/**
+ * @brief
  * Handling of Ethernet PHY's
  * PHY's communicate with an EMAC either through
  * a Media-Independent Interface (MII), or a Reduced Media-Independent Interface (RMII).
```
paulbartell, line 30: May want to mention that this is for phys with an MDIO/MDC bus connection (where there are 32 addresses). Typically the RGMII/RMII/MII bus is point to point.

paulbartell (approved):

PR #365, opened 2021-10-19
Add checks to reject IP packets with broadcast source IP/MAC address

AniruddhaKanhere:
Description
-----------
This PR adds checks to filter out IP packets with broadcast/multicast source MAC/IP addresses. These kind of packets are supposed to be rejected in conformance with [RFC 1112 section 3.2.1.3](https://datatracker.ietf.org/doc/html/rfc1122#section-3.2.1.3).

Test Steps
-----------
Send IP packets with source MAC set as broadcast address (`FF:FF:FF:FF:FF:FF`) OR source IP address set as a multicast address OR both to the device running FreeRTOS+TCP stack. Currently, those packets will be accepted and processed. 
After the change mentioned in this PR, the above packet will be dropped and not processed further.

Related Issue
-----------


FreeRTOS_IP.c
```diff
@@ -2025,6 +2026,18 @@ static eFrameProcessingResult_t prvAllowIPPacket( const IPPacket_t * const pxIPP
...
                 eReturn = eReleaseBuffer;
             }
+            else if( memcmp( ( void * ) &xBroadcastMACAddress,
+                             ( void * ) &( pxIPPacket->xEthernetHeader.xSourceAddress ),
+                             sizeof( MACAddress_t ) ) == 0 )
+            {
+                /* Ethernet source is a broadcast address. */
+                eReturn = eReleaseBuffer;
+            }
+            else if( ( ipFIRST_MULTI_CAST_IPv4 <= ulHostEndianSrcIPAddr ) && ( ulHostEndianSrcIPAddr < ipLAST_MULTI_CAST_IPv4 ) )
```
htibosch, line 2036: What about this change:
~~~c
- else if( ( ipFIRST_MULTI_CAST_IPv4 <= ulHostEndianSrcIPAddr ) && ( ulHostEndianSrcIPAddr < ipLAST_MULTI_CAST_IPv4 ) )
+ else if( xIsIPv4Multicast( ulSourceIPAddress ) == pdTRUE )
~~~
and you won't need the help variable `ulHostEndianSrcIPAddr` anymore.

AniruddhaKanhere: Done.

PR #369, opened 2021-10-27
Add required define to enable st driver

rpcme:
Description
-----------
When adding the port for the STM32F7 driver, you need to define that the module is enabled.  The define might be automatically set when using the CUBE IDE but it is not if you're using the SDK directly, like from https://github.com/STMicroelectronics/STM32CubeF7

Test Steps
-----------
Try to add port, and you will get zero symbols out of the driver, in which case NetworkInterface.o cannot find required HAL symbols.

Related Issue
-----------
None


portable/NetworkInterface/STM32Fxx/readme.txt
```diff
@@ -1,5 +1,21 @@
...
 
+
+BUILDING
+========
+
+When you are building outside of the IDE, for example building an
+application with the STM32 Cube SDK directly, you must define
+HAL_ETH_MODULE_ENABLED to enable the STM ethernet driver, i.e.
+
+gcc ... -DHAL_ETH_MODULE_ENABLED=1
```
htibosch, line 11: In order to run a TCP application, it is indeed necessary to define the macro `HAL_ETH_MODULE_ENABLED`.

But in stead of passing the macro to GCC, it is custom in STM projects to define the configuration in the standard ST configuration file `stm32f4xx_hal_conf.h` ( where `f4xx` may vary in other projects, depending on the platform. ). It is also used in the FreeRTOS+TCP makefile projects.

Mentioned header file will be included automatically in all places where it is needed.

htibosch (approved):
@rpcme, thank you for this PR.

portable/NetworkInterface/STM32Fxx/readme.txt
rpcme: Hello @htibosch I reviewed the file and ... I feel schooled :) I am not very familiar with the STM ecosystem so I missed that even though I have it in inc/!  I think my problem is I didn't have my app's inc first in the list.

I agree then that the content I provided doesn't make sense now.  But I think noting in the readme that the config should be included prior to the HAL includes probably should be there.  I don't use the CUBE IDE - I use the SDK direct since ... I am one of those Emacs people :D So I didn't realize my include order was wrong.

Can we leave the PR open and I work on reworking it to explain this file should be included prior to the HAL includes?

htibosch: Sure we can leave it open.

Some more information:
The STM32Fxx driver now works for F1, F2, F4, and F7. The H series (STM32Hxx) has [its own driver](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/main/portable/NetworkInterface/STM32Hxx).

The very latest STM32Fxx network driver can only be found in [this branch](https://github.com/htibosch/FreeRTOS-Plus-TCP/blob/IPv4_updating_network_interfaces/portable/NetworkInterface/STM32Fxx). It will be merged with main in [PR #361](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/361).

The driver needs to know the CPU family, which is one out of: 
~~~c
STM32F1xx 
STM32F2xx
STM32F4xx
STM32F7xx
~~~
I normally define this macro in Makefile, and pass it with `-D` to the compiler. The CPU family must indeed be known before including header files.

In my STM32F4xx project, I have these modules enabled in `stm32f4xx_hal_conf.h`:
~~~c
#define HAL_MODULE_ENABLED  
#define HAL_ETH_MODULE_ENABLED
#define HAL_SRAM_MODULE_ENABLED
#define HAL_RNG_MODULE_ENABLED   
#define HAL_RTC_MODULE_ENABLED
#define HAL_SD_MODULE_ENABLED
#define HAL_TIM_MODULE_ENABLED   
#define HAL_UART_MODULE_ENABLED
#define HAL_GPIO_MODULE_ENABLED
#define HAL_DMA_MODULE_ENABLED
#define HAL_RCC_MODULE_ENABLED
#define HAL_FLASH_MODULE_ENABLED
#define HAL_PWR_MODULE_ENABLED
#define HAL_CORTEX_MODULE_ENABLED
~~~

Your wrote:

> I don't use the CUBE IDE - I use the SDK direct since ... I am one of those Emacs people :D So I didn't realize my include order was wrong.

Same for me,
When I have a new CPU, I use Cube only once to create the configuration files ( clocks, RAM, etc ). After that I continue with Make, Eclipse, and my favourite editor (PN).

AniruddhaKanhere (approved):

paulbartell (approved):

htibosch:
@rpcme wrote:

> Can we leave the PR open and I work on reworking it to explain this file should be included prior to the HAL includes?

To @AniruddhaKanhere, @paulbartell:

That means that this PR was on hold until the text is ready. I know it is only about a README file, but the text is not in style with ST's custom of configuring a project.

Can a merge be undone? Or shall I propose a new text? What do you think, @rpcme?

Hein

htibosch:
I just wrote a new proposal for the readme.txt, see PR #372.

PR #370, opened 2021-10-29
Remove a define from STM32Fxx network interface.

htibosch:
Description
-----------
The network driver for STM32Fxx defines a macro `ipFRAGMENT_OFFSET_BIT_MASK`. It is used for filtering out fragmented packets. However, since [PR #179](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/179), the macro is defined in `FreeRTOS_IP_Private.h`, in a slightly different way. That gives a conflict as reported by [Richard Elberger](https://forums.freertos.org/u/rpcme) in [this forum post](https://forums.freertos.org/t/freertos-tcp-and-stm32fx-redefined-definition/13773).

This PR will simply remove the `#define`.

Related Issue
-----------
A compiler conflict for `STM32Fxx\NetworkInterface.c`.

### DIFF
--- a/portable/NetworkInterface/STM32Fxx/NetworkInterface.c
+++ b/portable/NetworkInterface/STM32Fxx/NetworkInterface.c
@@ -801,7 +801,7 @@ static BaseType_t xMayAcceptPacket( uint8_t * pcBuffer )
 
             /* Ensure that the incoming packet is not fragmented (only outgoing packets
              * can be fragmented) as these are the only handled IP frames currently. */
-            if( ( pxIPHeader->usFragmentOffset & FreeRTOS_ntohs( ipFRAGMENT_OFFSET_BIT_MASK ) ) != 0U )
+            if( ( pxIPHeader->usFragmentOffset & ipFRAGMENT_OFFSET_BIT_MASK ) != 0U )
             {
                 return pdFALSE;
             }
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (dismissed):

htibosch:
Hello  @rpcme,
 
> Thank you @htibosch - we will def want this before a relatively large conference in about 5 weeks :)

Which conference will that be? Las Vegas?

When we moved `ipFRAGMENT_OFFSET_BIT_MASK` to `FreeRTOS_IP_Private.h`, it was also changed to become network-endian.
~~~c
    #if ( ipconfigBYTE_ORDER == pdFREERTOS_LITTLE_ENDIAN )

/* The bits in the two byte IP header field that make up the fragment offset value. */
        #define ipFRAGMENT_OFFSET_BIT_MASK         ( ( uint16_t ) 0xff1fU )
    #else /* if ( ipconfigBYTE_ORDER == pdFREERTOS_LITTLE_ENDIAN ) */
/* The bits in the two byte IP header field that make up the fragment offset value. */
        #define ipFRAGMENT_OFFSET_BIT_MASK         ( ( uint16_t ) 0x1fffU )
    #endif
~~~
Therefore, in NetworkInterface.c, it is no more necessary to call `ntohs()`.
I just committed that change:

~~~c
-if( ( usFragmentOffset & FreeRTOS_ntohs( ipFRAGMENT_OFFSET_BIT_MASK ) ) != 0U )
+if( ( usFragmentOffset & ipFRAGMENT_OFFSET_BIT_MASK ) != 0U )
~~~

htibosch:
I will close this PR because I copied the tiny change to #361.

PR #371, opened 2021-10-29
Fix warnings in FreeRTOS_DHCP.c

AniruddhaKanhere:
Description
-----------
This PR fixes some warnings in the code as pointed out by [this](https://forums.freertos.org/t/freertos-tcp-dhcp-type-qualifier-warnings/13784) forum post by @rpcme.

The warnings should not cause any unusual behavior.

Test Steps
-----------

Related Issue
-----------


tomstorey:
Hi,

I have another warning coming up when compiling.

In FreeRTOS_DHCP.c, line 1160:

```c
pvCopyDest = &pucUDPPayloadBuffer[ dhcpFIRST_OPTION_BYTE_OFFSET + dhcpREQUESTED_IP_ADDRESS_OFFSET ];
```

This produces the following warning:

```
warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
```

It is fixed by modifying the line to be:

```c
pvCopyDest = (void *)&pucUDPPayloadBuffer[ dhcpFIRST_OPTION_BYTE_OFFSET + dhcpREQUESTED_IP_ADDRESS_OFFSET ];
```

I wonder if you wouldnt mind slipping that one into your PR as well? :-)

FreeRTOS_DHCP.c
```diff
@@ -1162,7 +1162,7 @@
     static BaseType_t prvSendDHCPDiscover( void )
     {
         BaseType_t xResult = pdFAIL;
-        uint8_t const * pucUDPPayloadBuffer;
+        uint8_t * pucUDPPayloadBuffer;
```
htibosch, line 1165: I am not sure why you want to make `pucUDPPayloadBuffer` a const pointer? A few lines further down there is this cast:
~~~c
void * pvCopyDest = &pucUDPPayloadBuffer[ dhcpFIRST_OPTION_BYTE_OFFSET + dhcpREQUESTED_IP_ADDRESS_OFFSET ];
~~~
and data will be written to the buffer.

htibosch: I think that the changes are all good, thank you.

htibosch (approved):

FreeRTOS_DHCP.c
tomstorey: Hm yup, you can ignore my message too, removing const does seem better, and removes the warning I got. Protip, dont write (read?) code late at night. :-P

AniruddhaKanhere: @tomstorey, I was about to type us a reply saying that I had already removed the `const` (It is late here 🙂 - your theory about not reading code at night seems correct). But then I saw this message (by accident) and stopped. :)

paulbartell (approved):

PR #372, opened 2021-11-05
IPv4/single: Added text to readme.txt for the STM32Fxx driver

htibosch:
Description
-----------

Recently, in PR #369, some text was added to the readme.txt file of the STM32Fxx driver. The recommendation was OK, but it was not in line with the way in which the STM32 HAL should be configured.

This PR will explain how to use and configure the HAL properly, i.e. by using the HAL configuration file, e.g. `stm32f4xx_hal_conf.h`.

**EDIT** : also I learned that a PR can be put into DRAFT mode, which will prevent to from getting merged prematurely :-)


rpcme:
@htibosch I like the changes and I think the modifications are very useful to customers using the port.

htibosch:
Thanks @rpcme for reviewing this PR. I have now put it in normal mode again, so it can be merged.

AniruddhaKanhere (approved):

PR #387, opened 2021-11-26
IPv4/single: TCP minimum time for retransmissions

htibosch:
Description
-----------
When TCP sends a packet, it will wait some time for an ACK. It will also measure how long it normally takes before a peer responds. That time is called Smoothed Round Trip Time (SRTT).
In FreeRTOS_TCP_WIN.c the value will capped at `winSRTT_CAP_mS`, which had a value of 50 ms. That can be OK when talking to other real-time peers, but a host may often delay sending an ACK for 200 ms.

A developer [nlangenb](https://forums.freertos.org/u/nlangenb) saw unnecessary [resends of packets](https://forums.freertos.org/t/freertos-tcp-retransmission-timer-incorrect). That can be solved by waiting at lease 1000 ms. THis small PR makes that possible.

Actually, a good minimum would be 1000, but as we never want to disrupt existing applications, the default will still be 50 ms.

### DIFF
--- a/FreeRTOS_TCP_WIN.c
+++ b/FreeRTOS_TCP_WIN.c
@@ -50,11 +50,12 @@
 #if ( ipconfigUSE_TCP == 1 )
 
 /* Constants used for Smoothed Round Trip Time (SRTT). */
-    #define winSRTT_INCREMENT_NEW        2  /**< New increment for the smoothed RTT. */
-    #define winSRTT_INCREMENT_CURRENT    6  /**< Current increment for the smoothed RTT. */
-    #define winSRTT_DECREMENT_NEW        1  /**< New decrement for the smoothed RTT. */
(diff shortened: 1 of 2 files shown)


RAc-GIT:
Hi there Hein, I was wondering if the SRTT is the starting point for the exponential backoff RTT, implying that if the second retransmission doesn't get acknowledged, subsequent retransmissions will be delayed even further? Thanks Ruediger

htibosch:
@RAc-GIT wrote:
> implying that if the second retransmission doesn't get acknowledged, subsequent retransmissions will be delayed even further?

That is correct, after every time-out the waiting time will be multiply with 2.

Only when e.g. packets 1, 2, 4 are acknowledged, the missing packet 3 will get a fast retransmit.

AniruddhaKanhere (approved):

cobusve (approved):

PR #391, opened 2021-12-09
IPv4/single: added comments to FreeRTOSIPConfigDefaults.h

htibosch:
Description
-----------
Added many inline comments to the header file 'FreeRTOSIPConfigDefaults.h'

More complete documentation can be found on [freertos.org](https://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/TCP_IP_Configuration.html), which will be updated soon.

I left the definitions exactly as they were, not changing any value. I did reorder some items, to keep similar macros together as a group: ARP, ICMP, DHCP, DNS etc.

### DIFF
--- a/.github/lexicon.txt
+++ b/.github/lexicon.txt
@@ -192,6 +192,7 @@ eapplicationprocesscustomframehook
 earpcachehit
 earpcachemiss
 earpgetcacheentry
+earpgetcacheentrybymac
 earptimerevent
 ec
 ecantsendpacket
@@ -595,6 +596,7 @@ pcdata
 pcdestination
 pce
 pcf
+pcformat
 pcfrom
 pchostname
 pcipaddress
@@ -640,6 +642,7 @@ pme
 pmecr
 pmt
 pointee
+portinline
 posix
 pparam
 ppkt
@@ -1256,6 +1259,7 @@ uxenoughspace
 uxentrycount
 uxflags
 uxfront
+uxgetminimumipqueuespace
 uxhead
 uxheaderlength
 uxidentifier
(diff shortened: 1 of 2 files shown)


AniruddhaKanhere:
Hello Hein,

Thank you for this PR. I see that the kernel submodule pointer has also been updated in this. I don't think that was intended - was it?
Can you take a look and please remove the changes to the kernel submodule pointer?

Thanks

htibosch:
Thank you Aniruddha for your comments, I applied all suggestions. Although I still have doubts about:
~~~c
/* This file should be included only after the configuration files FreeRTOSConfig.h and
 * FreeRTOSIPConfig.h have been included already.
~~~
When the comment says **should**, the user might think that action is required. I thought more of a confirmation:
~~~c
/* This file is included after the configuration files FreeRTOSConfig.h and
 * FreeRTOSIPConfig.h have been included already.
~~~
What do you think?

AniruddhaKanhere:
> Thank you Aniruddha for your comments, I applied all suggestions. Although I still have doubts about:
> 
> ```c
> /* This file should be included only after the configuration files FreeRTOSConfig.h and
>  * FreeRTOSIPConfig.h have been included already.
> ```
> 
> When the comment says **should**, the user might think that action is required. I thought more of a confirmation:
> 
> ```c
> /* This file is included after the configuration files FreeRTOSConfig.h and
>  * FreeRTOSIPConfig.h have been included already.
> ```
> 
> What do you think?

Maybe it should say "must"? That is: `This file must be included ...`

htibosch:
About the commit "A minor change: should becomes must": that was the last thing open. Thanks for reviewing.

AniruddhaKanhere (approved):

n9wxu (approved):

PR #392, opened 2021-12-10
IPv4/single: TCP handles partial segments.

htibosch:
Description
-----------
When using UDP, it is possible to receive data without first copying it. In that case it will return a pointer to a network buffer.

This is the way it works:

~~~c
    socklen_t uxAddrLen = sizeof xAddress;
    uint8_t *pucBuffer = NULL;
    BaseType_t xLength = FreeRTOS_recvfrom( xSocket, &( pucBuffer ), 0, FREERTOS_ZERO_COPY, &( xAddress ), &( uxAddrLen ) );
    if( xLength > 0 )
    {
        /* Use the contents of 'pucBuffer'. */
        /* Release the network buffer that contains 'pucBuffer'. */
        FreeRTOS_ReleaseUDPPayloadBuffer( pucBuffer );
    }
~~~

For TCP sockets, it is possible to peek into the circular reception buffer without copying it. When done, the pointer in the reception buffer must be advanced to mark the data as "read".

Recently, a user Matt Boytim on the [FreeRTOS forum](https://forums.freertos.org/t/releaseudppayloadbuffer-but-for-tcp/13890) asked for a function similar to 'FreeRTOS_ReleaseUDPPayloadBuffer()' for TCP. We wrote this function and named it, not surprisingly 'FreeRTOS_ReleaseTCPPayloadBuffer()'. 

~~~c
    uint8_t *pucBuffer = NULL;
    BaseType_t xLength = FreeRTOS_recv( xSocket, &( pucBuffer ), 0, FREERTOS_ZERO_COPY );
    if( xLength > 0 )
    {
        /* Use the contents of 'pucBuffer'. */
        /* Release the space that was occupied by received data. */
        FreeRTOS_ReleaseTCPPayloadBuffer( xSocket, pucBuffer, xLength );
    }
~~~

This technique of calling FreeRTOS_recv() with the zero-copy flag always seemed to work well. Except, we found out, when using it in an echo server.

~~~c
    BaseType_t xLength = FreeRTOS_recv( xSocket, &( pucBuffer ), 0, FREERTOS_ZERO_COPY );
    if( xLength > 0 )
    {
        FreeRTOS_send( xSocket, pucBuffer, xLength, 0 );
        FreeRTOS_ReleaseTCPPayloadBuffer( xSocket, pucBuffer, xLength );
    }
~~~

When calling FreeRTOS_send(), the driver starts returning the data to the echo client, without sending an ACK for the received data.

The client then sees that the earlier packet must have been dropped and quickly repeats it. The echo-server will send an ACK only after returning the data. These two packets may cross: the repeated data and the ACK.

Still not a big problem, except when the echo-client sends the packet again, but now with more bytes:

image

Wireshark marks the repeated packet as "**TCP out of order**", and FreeRTOS+TCP just drops the packet, thinking it is an error.

This is all not fatal, but it leads to time-outs and unnecessary Selective ACK's.

This PR repairs that: when receiving a TCP segment that was partially seen already, it will add the unseen bytes and acknowledge it properly. After this PR, it is OK to use zero-copy perception in an echo-server.

Test Steps
-----------
Create an echo server using zero-copy reception, as described above.

Related Issue
-----------
The communication will not run smoothly.

### DIFF
--- a/.github/lexicon.txt
+++ b/.github/lexicon.txt
@@ -1160,6 +1160,7 @@ ulrxwindowlength
 ulsenderprotocoladdress
 ulsequencenumber
 ulsize
+ulskipcount
 ulsourceipaddress
 ulspace
 ulsrtt
--- a/.github/lexicon.txt
+++ b/.github/lexicon.txt
@@ -743,6 +743,7 @@ pulipaddress
 pullups
 pulnetmask
 pulnumber
+pulskipcount
 pulvalue
 pvargument
 pvbuffer
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -1172,6 +1172,15 @@ void FreeRTOS_ReleaseUDPPayloadBuffer( void const * pvBuffer )
 /*-----------------------------------------------------------*/
 
 #if ( ipconfigUSE_TCP == 1 )
+
+/**
+ * @brief Release the memory that was previously obtained by calling FreeRTOS_recv()
+ *        with the flag 'FREERTOS_ZERO_COPY'.
+ *
+ * @param[in] xSocket: The socket that was read from.
+ * @param[in] pvBuffer: The buffer returned in the call to FreeRTOS_recv().
+ * @param[in] xByteCount: The number of bytes that have been used.
+ */
     void FreeRTOS_ReleaseTCPPayloadBuffer( Socket_t xSocket,
                                            void const * pvBuffer,
                                            BaseType_t xByteCount )
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -1171,20 +1171,20 @@ void FreeRTOS_ReleaseUDPPayloadBuffer( void const * pvBuffer )
 }
 /*-----------------------------------------------------------*/
 
-#if( ipconfigUSE_TCP == 1 )
+#if ( ipconfigUSE_TCP == 1 )
     void FreeRTOS_ReleaseTCPPayloadBuffer( Socket_t xSocket,
-    								       void const * pvBuffer,
-								           BaseType_t xByteCount )
+                                           void const * pvBuffer,
+                                           BaseType_t xByteCount )
     {
         BaseType_t xByteCountReleased;
-    	uint8_t * pucData;
-    	size_t uxBytesAvailable = uxStreamBufferGetPtr( xSocket->u.xTCP.rxStream, &( pucData ) );
+        uint8_t * pucData;
+        size_t uxBytesAvailable = uxStreamBufferGetPtr( xSocket->u.xTCP.rxStream, &( pucData ) );
 
-    	/* Make sure the pointer is correct. */
-    	configASSERT( pucData == ( uint8_t * ) pvBuffer );
+        /* Make sure the pointer is correct. */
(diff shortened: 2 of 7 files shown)


htibosch:
About the commit "Fine-tuning parameters for flow control" :

TCP-sockets in FreeRTOS+TCP have two parameters that express space in the input stream buffer in bytes:

    pxSocket->u.xTCP.uxLittleSpace
    pxSocket->u.xTCP.uxEnoughSpace

that determine the flow control of incoming data. These parameters may be set from the application with the socket option 'FREERTOS_SO_SET_LOW_HIGH_WATER'.

When the available space drops below 'uxLittleSpace', the IP-task will send a WIN update to the peer:

    pxSocket->u.xTCP.bits.bLowWater = pdTRUE;
    pxSocket->u.xTCP.bits.bWinChange = pdTRUE;

When the application reads from the socket, there will be more space in the buffer. As soon as the space is more than or equal to 'uxEnoughSpace', there will be another WIN update:

    pxSocket->u.xTCP.bits.bLowWater = pdFALSE;
    pxSocket->u.xTCP.bits.bWinChange = pdTRUE;

The earlier code was very simple:

    pxSocket->u.xTCP.uxLittleSpace = ( 20 * pxSocket->u.xTCP.uxRxStreamSize ) / 100;
    pxSocket->u.xTCP.uxEnoughSpace = ( 80 * pxSocket->u.xTCP.uxRxStreamSize ) / 100;

So the defaults were 20% and 80% of the size of the buffer. 

That works well when the reception buffer is big, like 5 or more MSS. But when I observed the behaviour with small buffers, things could be optimised.
The following scheme seems to work better than the default of 20%/80%:

~~~c
    { 0U,  100U }, /* 1 segment. */
    { 50U, 100U }, /* 2 segments. */
    { 34U, 100U }, /* 3 segments. */
    { 25U, 100U }, /* 4 segments. */
    { 20U,  80U }, /* 5 or more segments. */
~~~
Where "1 segment" means a buffer size of 1 x MSS, e.g. a buffer of 1460 bytes.

AniruddhaKanhere (approved):

htibosch:
@AniruddhaKanhere, thanks for approving this PR. Yesterday I compiled a project with:
~~~c
#define ipconfigUSE_TCP_WIN			( 0 )
~~~
and saw that we also need to change this:
~~~c
int32_t lTCPWindowRxCheck( TCPWindow_t * pxWindow,
                           uint32_t ulSequenceNumber,
                           uint32_t ulLength,
-                          uint32_t ulSpace )
+                          uint32_t ulSpace,
+                          uint32_t * pulSkipCount )
~~~
I will submit that later today.

htibosch:
This PR only needed a minor change:
~~~diff
#if( ipconfigUSE_TCP_WIN == 0 )
    int32_t lTCPWindowRxCheck( TCPWindow_t * pxWindow,
                               uint32_t ulSequenceNumber,
                               uint32_t ulLength,
-                              uint32_t ulSpace )
+                              uint32_t ulSpace,
+                              uint32_t * pulSkipCount )
#endif
~~~
and it is ready for review.

AniruddhaKanhere (dismissed):

n9wxu (approved):

n9wxu (approved):

PR #393, opened 2021-12-10
Extend the Zynq Ultrascale port for the Cortex R5

StefanBalt:
Description
-----------

Previously the Port only worked for the A53 in 64-bit mode. With these changes the stack is also working for the Cortex R5 as long as the nicUSE_UNCACHED_MEMORY option is off.

With nicUSE_UNCACHED_MEMORY enabled the R5 runs into an exception.

 ARP and ping work. I tested a lot of UDP steaming and everything seems to work fine. I did not test TCP.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/portable/NetworkInterface/xilinx_ultrascale/uncached_memory.c
+++ b/portable/NetworkInterface/xilinx_ultrascale/uncached_memory.c
@@ -84,13 +84,13 @@
         #define uncMEMORY_SIZE         0x100000U
         #define DDR_MEMORY_END         ( XPAR_PSU_R5_DDR_0_S_AXI_HIGHADDR )
         #define uncMEMORY_ATTRIBUTE    STRONG_ORDERD_SHARED | PRIV_RW_USER_RW
-    #else
+    #else /* if defined( __aarch64__ ) || defined( ARMA53_32 ) */
         #error Please define the specific target Zynq Ultrascale+ architecture
-    #endif
+    #endif /* if defined( __aarch64__ ) || defined( ARMA53_32 ) */
 #else /* if ( ipconfigULTRASCALE == 1 ) */
     #ifndef uncMEMORY_SIZE
         /* Reserve 1 MB of memory. */
-        #define uncMEMORY_SIZE         0x100000U
+        #define uncMEMORY_SIZE     0x100000U
     #endif
     #define DDR_MEMORY_END         ( XPAR_PS7_DDR_0_S_AXI_HIGHADDR + 1 )
     #define uncMEMORY_ATTRIBUTE    0x1C02
--- a/portable/NetworkInterface/xilinx_ultrascale/uncached_memory.c
+++ b/portable/NetworkInterface/xilinx_ultrascale/uncached_memory.c
@@ -69,13 +69,29 @@
 #include "uncached_memory.h"
 
 #if ( ipconfigULTRASCALE == 1 )
-    /* Reserve 2 MB of memory. */
-    #define uncMEMORY_SIZE         0x200000U
-    #define DDR_MEMORY_END         ( XPAR_PSU_DDR_0_S_AXI_HIGHADDR )
-    #define uncMEMORY_ATTRIBUTE    NORM_NONCACHE | INNER_SHAREABLE
-#else
-    /* Reserve 1 MB of memory. */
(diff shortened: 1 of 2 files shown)


htibosch:
Hi Stefan, thanks a lot for this PR. A small change that opens new possibilities!

Now for the Cortex R5, the network buffers will be located in the normal .data section, the normal RAM:
~~~c
#if ( nicUSE_UNCACHED_MEMORY == 0 )
    static uint8_t ucNetworkPackets[ ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS * niBUFFER_1_PACKET_SIZE ] __attribute__( ( aligned( 32 ) ) );
#endif
~~~

However, uncached_memory.c still reserves 1 MB, which will only be used to locate DMA descriptors:

    xemacpsif->rxSegments = ( struct xBD_TYPE * ) ( pucGetUncachedMemory( xRxSize ) );
    xemacpsif->txSegments = ( struct xBD_TYPE * ) ( pucGetUncachedMemory( xTxSize ) );

where xBD_TYPE holds an address and flags.

I wonder if you can not skip 'uncached_memory.c' entirely and declare the DMA descriptors statically:

    struct xBD_TYPE rxSegments[ ipconfigNIC_N_RX_DESC ];
    struct xBD_TYPE txSegments[ ipconfigNIC_N_TX_DESC ];

and thus save the MB otherwise declared in 'uncached_memory.c'?

You would have to rewrite `ucIsCachedMemory()` to have it always return false.

I would be curious to see if everything works fine when the descriptors are declared in normal memory.

Thanks

StefanBalt:
Hi Hein,
thanks for the fast feedback.
Static descriptors only work as long as the D-Cache is disabled. As soon as the (L1) D-Cache is active, the descriptors will not work between R5 and DMA without a lot of flush/invalidate.

I could make uncMEMORY_SIZE a lot smaller though.

StefanBalt:
> I could make uncMEMORY_SIZE a lot smaller though.

Well that is not true, I cannot make it smaller since the MPU only allows to set 1MB blocks.

portable/NetworkInterface/xilinx_ultrascale/uncached_memory.c
```diff
@@ -69,11 +69,18 @@
...
 
 #if ( ipconfigULTRASCALE == 1 )
-    /* Reserve 2 MB of memory. */
-    #define uncMEMORY_SIZE         0x200000U
-    #define DDR_MEMORY_END         ( XPAR_PSU_DDR_0_S_AXI_HIGHADDR )
-    #define uncMEMORY_ATTRIBUTE    NORM_NONCACHE | INNER_SHAREABLE
-#else
+    #if defined( __aarch64__ ) || defined( ARMA53_32 )
+        /* Reserve 2 MB of memory. */
+        #define uncMEMORY_SIZE         0x200000U
```
htibosch, line 74: It would be nice if it also works when `uncMEMORY_SIZE` is predefined, so make each of the three \#define's conditional:
~~~diff
-    #define uncMEMORY_SIZE         0x200000U
+    #ifndef uncMEMORY_SIZE
+        #define uncMEMORY_SIZE         0x200000U
+    #endif
~~~
The module `uncached_memory.c` is also used by other software, for instance for data caching of an SD-card. In that case `uncMEMORY_SIZE` may become bigger.

htibosch (approved):
Hello Stefan, sorry that this PR is sitting here so long.
I have 2 minor remarks, otherwise I approve of the changes.

htibosch (approved):
I am all happy with it, thank you Stefan!

AniruddhaKanhere:
Hello @StefanBalt,

Thank you for the PR, and apologies for the late response.

I just ran uncrustify to fix the formatting. For that, I commented `/bot run uncrustify` [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/393#issuecomment-1026416459).

I'll take a look at the PR ASAP.

Thanks

AniruddhaKanhere (approved):

htibosch:
@AniruddhaKanhere wrote:
> I just ran uncrustify to fix the formatting. For that, I commented /bot run uncrustify here.

@StefanBalt : do not forget to pull from your local repo before you do a new push.

PR #375, opened 2021-11-09
Refactor and unit-test FreeRTOS_IP.c

AniruddhaKanhere:
Description
-----------
This PR splits FreeRTOS_IP.c into multiple files to increase modularity and to increase ease of testing. Namely:
- FreeRTOS_IP.c - Core functionality and all `FreeRTOS_xxx` APIs.
- FreeRTOS_ICMP.c - Processing of ICMP packets.
- FreeRTOS_IP_Utils.c - Utility functions for IP. Such as `FreeRTOS_max_uint32` etc.
- FreeRTOS_IP_Timers.c - FreeRTOS_IP used to process timer events for ARP/DHCP/IP/TCP. This module now handles the timer related work.

In addition, this PR adds 100% unit-test coverage for FreeRTOS_IP.c and FreeRTOS_ICMP.c.
There will be a PR later which will add coverage for the remaining two files.

Test Steps
-----------

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #394, opened 2021-12-14
Unit-test IP_utils and IP_Timers

AniruddhaKanhere:
Description
-----------
This PR adds unit tests for IP_Utils and IP_Timers which were split up from FreeRTOS_IP.c
It is a follow up PR for https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/375

Test Steps
-----------

Related Issue
-----------


FreeRTOS_IP_Utils.c
```diff
@@ -1170,7 +1180,14 @@ size_t FreeRTOS_min_size_t( size_t a,
 uint32_t FreeRTOS_round_up( uint32_t a,
                             uint32_t d )
 {
-    return d * ( ( a + d - 1U ) / d );
+    uint32_t ulResult = 0;
+
+    if( d != 0 )
```
alfred2g, line 1185: if d is zero
zero will be returned, which is wrong 
is that the intended behavior?

test/unit-test/CMakeLists.txt
```diff
@@ -186,13 +186,15 @@ include( ${UNIT_TEST_DIR}/FreeRTOS_Sockets/ut.cmake )
...
+include( ${UNIT_TEST_DIR}/FreeRTOS_IP_Timers/ut.cmake )
 include( ${UNIT_TEST_DIR}/FreeRTOS_UDP_IP/ut.cmake )
 
 #  ==================================== Coverage Analysis configuration ========================================
 
 # Add a target for running coverage on tests.
 add_custom_target( coverage
     COMMAND ${CMAKE_COMMAND} -P ${MODULE_ROOT_DIR}/test/unit-test/cmock/coverage.cmake
-    DEPENDS cmock unity FreeRTOS_ARP_utest FreeRTOS_DHCP_utest FreeRTOS_UDP_IP_utest FreeRTOS_Stream_Buffer_utest FreeRTOS_Sockets_utest
+    DEPENDS cmock unity #FreeRTOS_ARP_utest FreeRTOS_DHCP_utest FreeRTOS_UDP_IP_utest FreeRTOS_Stream_Buffer_utest FreeRTOS_Sockets_utest
```
alfred2g, line 198: omitting all of these on purpose ?

FreeRTOS_IP_Utils.c
AniruddhaKanhere: I have added an assert to make sure that `d == 0` is caught in development phase.
However, since we cannot return an error with this function signature, I am returning `a` in round_up and `0` in round_down when assert is not defined.

test/unit-test/CMakeLists.txt
AniruddhaKanhere: Oops, fixed now.

alfred2g (dismissed):

n9wxu (approved):

PR #397, opened 2021-12-17
Decouple dns cache

alfred2g:
Decouple dns cache

Description
-----------
Refactor DNS and add Unit Test

Test Steps
-----------
All Unit Tests passes


htibosch:
Hello Alfred, thanks for this great job you are doing.
There is one thing I do not understand: there are 2 versions of FreeRTOS_DNS.c: one in the root directory, and one in DNS/.
I assume you will remove the one in the root directory?

This I like in `DNS/FreeRTOS_DNS.c`:
~~~c
#include "DNS/DNS_Cache.h"
#include "DNS/DNS_Parser.h"
#include "DNS/DNS_Networking.h"
#include "DNS/DNS_Callback.h"
~~~
You also include the sub directory `DNS/`, which saves me one -I option.

I hope that the applications does not have to change anything to the `#include` statements?

Thanks so far

PR #398, opened 2021-12-20
Fix constant comparison warnings

AniruddhaKanhere:
Description
-----------
There were some comparisons of constants with constants in configASSERTS which the compilers are not a big fan of. This PR updates the way the comparisons are done so as not to get any warnings from compilers. 

Test Steps
-----------

Related Issue
-----------


gitwallit (commented):
Is the volatile qualifier required? would the compiler still complain that the conditional expression is constant if volatile is not used?

alfred2g (approved):

AniruddhaKanhere:
@gitwallit,
> Is the volatile qualifier required? would the compiler still complain that the conditional expression is constant if volatile is not used?

I believe so. It makes sure that the compiler thinks that the variable can change out of context and it should be loaded from memory every time - i.e., it is not a constant.

n9wxu (approved):

PR #401, opened 2021-12-22
Minor changes to the GH action uncrustify

AniruddhaKanhere:
Description
-----------
This PR adds a comment to let the user know on what to do when uncrustify failed.
Also, it also adds changes to run uncrustify a second time to catch any deviations which were missed the first time.

Test Steps
-----------

Related Issue
-----------


.github/workflows/ci.yml
```diff
@@ -76,6 +76,11 @@ jobs:
         run: |
           uncrustify --version
           find . -not -path '.*catch_assert.h' -iname "*.[hc]" -exec uncrustify --check -c tools/uncrustify.cfg {} +
+          if [ "$?" = "0" ]; then
+            exit 0
+          else
+            echo "To have the code uncrustified for you, please comment '/bot run uncrustify' on the Pull Request."
```
kstribrnAmzn, line 82: Shouldn't this rather echo that the uncrustify job has failed? The `if` of the conditional checks for the uncrustify job status.

.github/workflows/uncrustify.yml
```diff
@@ -40,6 +40,7 @@ jobs:
         run: |
           uncrustify --version
           find . -not -path '.*catch_assert.h' -iname "*.[hc]" -exec uncrustify -c tools/uncrustify.cfg --no-backup --replace {} +
+          find . -not -path '.*catch_assert.h' -iname "*.[hc]" -exec uncrustify -c tools/uncrustify.cfg --no-backup --replace {} +
```
kstribrnAmzn, line 43: More of a suggestion for future work...It would be cool to run uncrustify as long as it is modifying files. I'd think this could be done in a loop by simply staging before every uncrustify operation and then after uncrustify-ing checking if any modified files are present.

AniruddhaKanhere: Agreed. I am also thinking of just uncrustify-ing the files which were actually changed by the PR. However, these optimizations can be added incrementally.

AniruddhaKanhere (commented):
@kstribrnAmzn, I have updated the PR. Added a comment which would be more helpful (and hopefully more visible) to the user along with returning a non-zero status so that the GH action will fail.

.github/workflows/ci.yml
AniruddhaKanhere: Updated :)

paulbartell (approved):

PR #405, opened 2022-01-14
Remove DNS Cache dependency from DNS code

### LINKED ISSUE #390: Domain name handling with out DNS cache enabled
I downloaded the latest multi/IPv6 branch and compiled it in default configuration. The DNS code produces compilations errors:

- pcName is disabled in ParseSet_t, even needed for dns operation [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/dbba1fad396eed22a2c418db9882eb6994d4a0a6/include/FreeRTOS_DNS.h#L139)
- prvReadNameField is disabled, even needed [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/dbba1fad396eed22a2c418db9882eb6994d4a0a6/FreeRTOS_DNS.c#L223-226)
- ucName is disabled in freertos_addrinfo, even needed for dns operations. [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/dbba1fad396eed22a2c418db9882eb6994d4a0a6/include/FreeRTOS_DNS.h#L94-L96)

go2sh:
Description
------------
Fix the DNS parsing code to not rely on the DNS cache option.

Test Steps
-----------
Compile the code with ipconfigUSE_DNS_CACHE = 0 and it works.

Related Issue
-----------
Fixes #390 

### DIFF
--- a/FreeRTOS_DNS.c
+++ b/FreeRTOS_DNS.c
@@ -73,7 +73,10 @@
 
 /** @brief If the top two bits in the first character of a name field are set then the
  * name field is an offset to the string, rather than the string itself. */
-    #define dnsNAME_IS_OFFSET    ( ( uint8_t ) 0xc0 )
+    #define dnsLABEL_IS_POINTER    ( ( uint8_t ) 0xc0 )
+
+/** @brief A zero length label equals the root domain. */
+    #define dnsLABEL_IS_ROOT       ( ( uint8_t ) 0x00 )
 
 /* NBNS flags. */
     #if ( ipconfigUSE_NBNS == 1 )
@@ -220,10 +223,8 @@
     #endif /* ipconfigUSE_NBNS */
 
 
-    #if ( ipconfigUSE_DNS_CACHE == 1 ) || ( ipconfigDNS_USE_CALLBACKS == 1 )
-        _static size_t prvReadNameField( ParseSet_t * pxSet,
-                                         size_t uxDestLen );
-    #endif /* ipconfigUSE_DNS_CACHE || ipconfigDNS_USE_CALLBACKS */
+    _static size_t prvReadNameField( ParseSet_t * pxSet,
+                                     size_t uxDestLen );
 
     #if ( ipconfigUSE_DNS_CACHE == 1 )
         /** @brief Copy DNS cache entries at xIndex to a linked struct addrinfo. */
@@ -979,8 +980,10 @@
             pxAddrInfo = ( struct freertos_addrinfo * ) pvBuffer;
 
             ( void ) memset( pxAddrInfo, 0, sizeof( *pxAddrInfo ) );
-            pxAddrInfo->ai_canonname = pxAddrInfo->xPrivateStorage.ucName;
-            ( void ) strncpy( pxAddrInfo->xPrivateStorage.ucName, pcName, sizeof( pxAddrInfo->xPrivateStorage.ucName ) );
+            #if ( ipconfigUSE_DNS_CACHE != 0 )
+                pxAddrInfo->ai_canonname = pxAddrInfo->xPrivateStorage.ucName;
+                ( void ) strncpy( pxAddrInfo->xPrivateStorage.ucName, pcName, sizeof( pxAddrInfo->xPrivateStorage.ucName ) );
+            #endif
 
             #if ( ipconfigUSE_IPv6 == 0 )
                 pxAddrInfo->ai_addr = &( pxAddrInfo->xPrivateStorage.sockaddr4 );
@@ -2037,8 +2040,6 @@
     }
 /*-----------------------------------------------------------*/
 
-    #if ( ipconfigUSE_DNS_CACHE == 1 ) || ( ipconfigDNS_USE_CALLBACKS == 1 )
-
 /**
  * @brief Read the Name field out of a DNS response packet.
  *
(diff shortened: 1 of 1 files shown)


FreeRTOS_DNS.c
```diff
@@ -2047,146 +2052,125 @@
...
+                /* The domain pointer is finished */
+                if( uxReturnIndex != 0 )
+                {
+                    uxIndex = uxReturnIndex;
+                }
 
-                    /* Determine if the name is the fully coded name, or an offset to the name
-                     * elsewhere in the message. */
-                    if( ( uxCount & dnsNAME_IS_OFFSET ) == dnsNAME_IS_OFFSET )
+                #if ( configDNS_CACHE_ENABE != 0 )
```
AniruddhaKanhere, line 2087: This should change from `configUSE_DNS_CACHE` -> `ipconfigUSE_DNS_CACHE`.

FreeRTOS_DNS.c
```diff
@@ -2047,146 +2052,125 @@
...
+                                     size_t uxDestLen )
+    {
+        size_t uxSourceLen = pxSet->uxSourceBytesRemaining;
+        size_t uxLabelLen;
+        size_t uxIndex = 0U;
+        size_t uxReturnIndex = 0U;
+        size_t uxLabelOffset;
+        const uint8_t * pucByte = pxSet->pucByte;
+
+        #if ( configDNS_CACHE_ENABE != 0 )
```
AniruddhaKanhere, line 2066: This should change from `configUSE_DNS_CACHE` -> `ipconfigUSE_DNS_CACHE`.

AniruddhaKanhere:
Hello @go2sh, I took the liberty to fix the formatting on your PR. To do so, I commented `/bot run uncrustify` on the PR.

For future reference, you can do that too to fix the formatting without needing to install uncrustify locally.

You might want to pull in the latest changes.

AniruddhaKanhere (commented):
Hello @go2sh,

There are some changes I would like to request you to make. Can you please take a look?

FreeRTOS_DNS.c
```diff
@@ -2047,146 +2053,125 @@
...
                         }
                     }
+                #endif /* if ( configDNS_CACHE_ENABE != 0 ) */
 
-                    if( ( uxIndex + uxCount ) > uxSourceLen )
-                    {
-                        uxIndex = 0U;
-                        break;
-                    }
+                /* Successfully paresed a name */
```
AniruddhaKanhere, line 2103: And maybe change this to 'parsed'?
Thanks :)

htibosch:
Hello [Christoph Seitz](https://github.com/go2sh), thank you for this PR. It is a bit difficult to understand it by reading the diff. The function `prvReadNameField()` was heavily changed.
I see a new macro `configDNS_CACHE_ENABE` which seems to have a similar meaning as `ipconfigUSE_DNS_CACHE`. Note that latter is already used in many existing projects, so:
~~~cmd
sed "s/configDNS_CACHE_ENABE/ipconfigUSE_DNS_CACHE/g"
~~~
I get a compilation error here at line 2153:
~~~c
    /* Move the label */
    if( ( uxNameLen + uxCount ) <= uxDestLen )
~~~
It is in a block that only compiles when `ipconfigUSE_DNS_CACHE` is defined.

I have just tested with:
~~~h
#define ipconfigUSE_DNS_CACHE      0
#define ipconfigUSE_IPv6           1
~~~
and success, it works!
I looked up both IPv4 and IPv6 addresses, also from hosts with multiple IP-addresses like google.x

What we should look at is this:
~~~
    #if ( ipconfigUSE_DNS_CACHE != 0 )
        char ucName[ ipconfigDNS_CACHE_NAME_LENGTH ];
    #endif
~~~
Maybe it should depend on the macro `ipconfigDNS_CACHE_NAME_LENGTH`:
~~~c
    #if ( ipconfigDNS_CACHE_NAME_LENGTH != 0 )
        char ucName[ ipconfigDNS_CACHE_NAME_LENGTH ];
    #endif
~~~
and it must also be checked where the field `ucName` is being used.

I will come back to you, these were my first impressions.

htibosch:
One question:
~~~diff
-    uxResult = prvReadNameField( pxSet,
-                                 sizeof( pxSet->pcName ) );
+    #if ( ipconfigUSE_DNS_CACHE != 0 )
+        uxResult = prvReadNameField( pxSet,
+                                     sizeof( pxSet->pcName ) );
+    #else
+        uxResult = prvReadNameField( pxSet, 0 );
+    #endif
~~~
Why wouldn't we include `ParseSet_t::pcName` and `freertos_addrinfo::ucName` unconditionally?
In fact those fields have nothing to do with the presence of a DNS cache table.

It would be good then to check the existence and validity of the macro `ipconfigDNS_CACHE_NAME_LENGTH`, something like it must be between 16 and 254 or so.

Can you tell why you reworked the function `prvReadNameField()`? Did you encounter problems with the existing implementation?

I think that when you want to make the current DNS module work while `ipconfigUSE_DNS_CACHE == 0`, you only need a very small PR. But please tell if you have different thoughts.

Thanks

htibosch:
Hi @go2sh , I just tested this version of FreeRTOS_DNS.c :
[FreeRTOS_DNS_without_cache.zip](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/8009468/FreeRTOS_DNS_without_cache.zip)

As you will see, it has very minimal changes compared to the [current version](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/labs/ipv6_multi/FreeRTOS_DNS.c).

Can you also give it a try?

I tested both IPv4 and IPv6 lookups, both with DNS, mDNS, and LLMNR.
All tests were done while `ipconfigUSE_DNS_CACHE = 0`.

Also I tried everything with and without the asynchronous calls (see `ipconfigDNS_USE_CALLBACKS`).

go2sh:
Hi thank you for your feedback.

The configUSE_DNS_CACHE thing is just a typo and ipconfigUSE_DNS_CACHE should be used.

Regarding: Why wouldn't we include ParseSet_t::pcName and freertos_addrinfo::ucName unconditionally?
I don't know. It was in the code before. I guess it was meant to be reducing the memory footprint by not storing the domain name string. Technically the name is not needed for a response, only if you use the dns cache. The domain name length should be checked yes, but i doubt that there are reasons for a value smaller then 254. Since parse set and freertos_addrinfo should be stack varibles. I'am fine with enabling them unconditionally or add an other config option to disable the name storing.

Regarding the prvReadNameField:
It was not able to parse a domain like "." or some other domains with pointed to the root. The root is a special label in a dns query. I also did rewrite it to be able to remove the code handling the char insertion into the destionation buffer.

I'll try your code and update the pr.

AniruddhaKanhere:
Hello @go2sh,
Do you have any update on this PR?

go2sh:
Hi,
I fixed the naming and the uxCount issue.
After thinking about it. The USE_DNS_CACHE option is there to completely disable storing any name in any structure right now. I don't know if this is necessary or not. Parsing the name is always needed for the offset, but the name itself is only used insed the DNS cache and the addrinfo. So for me there are two questions:
 * Should it be possible to run the code without any storing of the dns name?
 * Should storing the name in the addrinfo be decouple from the DNS cache?

BR

htibosch:
Hi @go2sh, I am sorry that your PR is taking so much time.

Yesterday I finally tested your proposed changes more fully, i.e. the following:

1) Lookup several hosts on the Internet
2) Lookup a host on my LAN using LLMNR
3) Lookup a host on my LAN using mDNS
4) Lookup my DUT from a laptop using LLMNR
5) Lookup my DUT from a laptop using mDNS
6) Test the asynchronous lookups ( using a call-back function )

Only test 1) succeeded, the others failed.

I did all tests for both IPv4 and IPv6 DNS packets, and asking for either IPv4 or IPv6 replies ( "A" and "AAA" responses ). So in fact there are 4 x 4 tests for LLMNR/mDNS.

When I enable DSN caching, the program doesn't compile

The DNS module from [the ZIP file](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/8009468/FreeRTOS_DNS_without_cache.zip) that I attached here above passed all tests, except one. I will look into that.

> It was not able to parse a domain like "." or some other domains with pointed to the root.

Can you give an example of a DNS lookup that has this "."? Or how can I replicate this problem?

Thanks

AniruddhaKanhere:
Hello @go2sh,

We have added the task to remove this dependency in our backlog. We have already done so in the main (IPv4) branch. We will do it in IPv6 branch soon.

Thanks for bringing this to our attention. We appreciate your efforts!

But, as Hein mentioned, your changes need more work before they can be accepted. For now, due to lack of response, I shall be closing this PR. Feel free to open this if you figure that the bugs have been resolved or if you'd like to contribute only a part of it.

Thanks again

PR #406, opened 2022-01-14
Fix:  compiler warnings and LLMNR fine tuning

alfred2g:
Fix: compiler warnings and LLMNR fine tuning

Description
-----------
Fix some compilation warnings and fine tune LLMNR behavior

### DIFF
--- a/DNS/DNS_Cache.c
+++ b/DNS/DNS_Cache.c
@@ -116,7 +116,7 @@
  * @brief perform a dns update in the local cache
  * @param pcName the lookup name
  * @param pulIP the ip value to insert/replace
- * @param ulTTL ignored
+ * @param ulTTL Time To live in seconds insidet eh dns cache
  * @return this is a dummy return, we are actually ignoring the return value
  *         from this function
  * @post the global structure \a xDNSCache might be modified
@@ -227,8 +227,8 @@
  */
     static BaseType_t prvFindEntryIndex( const char * pcName )
     {
-        int index = -1;
-        int x;
+        BaseType_t index = -1;
+        BaseType_t x;
 
         /* For each entry in the DNS cache table. */
         for( x = 0; x < ( int ) ipconfigDNS_CACHE_ENTRIES; x++ )
@@ -368,34 +368,5 @@
             }
         }
     }
-
-    #if 0
-        void dns_show_cache()
-        {
-            BaseType_t x;
-
-            for( x = 0; x < ( BaseType_t ) ipconfigDNS_CACHE_ENTRIES; x++ )
-            {
-                if( xDNSCache[ x ].pcName[ 0 ] == ( char ) 0 )
-                {
-                    continue;
-                }
-
-                TickType_t xCurrentTickCount = xTaskGetTickCount();
-                uint32_t ulCurrentTimeSeconds;
-
-                ulCurrentTimeSeconds = ( xCurrentTickCount / portTICK_PERIOD_MS ) / 1000UL;
-                uint32_t ulAge = ulCurrentTimeSeconds - xDNSCache[ x ].ulTimeWhenAddedInSeconds;
-                FreeRTOS_debug_printf( ( "Entry[%u] %xip Count %u/%u Age %u/%u (%s)\n",
-                                         ( int ) x,
-                                         ( unsigned ) FreeRTOS_ntohl( xDNSCache[ x ].ulIPAddresses[ 0 ] ),
-                                         xDNSCache[ x ].ucCurrentIPAddress,
-                                         xDNSCache[ x ].ucNumIPAddresses,
-                                         ( unsigned ) ulAge,
-                                         ( unsigned ) FreeRTOS_ntohl( xDNSCache[ x ].ulTTL ),
-                                         xDNSCache[ x ].pcName ) );
-            }
-        }
-    #endif /* if 0 */
 
 #endif /* if ( ipconfigUSE_DNS != 0 ) */
--- a/DNS/DNS_Cache.c
+++ b/DNS/DNS_Cache.c
@@ -116,7 +116,7 @@
  * @brief perform a dns update in the local cache
  * @param pcName the lookup name
  * @param pulIP the ip value to insert/replace
- * @param ulTTL Time To live in seconds insidet eh dns cache
+ * @param ulTTL Time To live (in seconds)
  * @return this is a dummy return, we are actually ignoring the return value
  *         from this function
  * @post the global structure \a xDNSCache might be modified
@@ -148,7 +148,7 @@
  * @param[in] pcName: the name of the host
  * @param[in,out] pulIP: when doing a lookup, will be set, when doing an update,
  *                       will be read.
- * @param[in] ulTTL: Time To Live
+ * @param[in] ulTTL: Time To Live (in seconds)
  * @param[in] xLookUp: pdTRUE if a look-up is expected, pdFALSE, when the DNS cache must
  *                     be updated.
  * @return whether the operation was successful
@@ -301,7 +301,7 @@
 /**
  * @brief update entry at \p index in the cache
  * @param[in] uxIndex : index in the cache
- * @param[in] ulTTL time to live
+ * @param[in] ulTTL time to live (in seconds)
  * @param[in] pulIP ip to update the cache with
  * @param[in] ulCurrentTimeSeconds current time
  * @post the global structure \a xDNSCache is modified
@@ -332,7 +332,7 @@
 /**
  * @brief insert entry in the cache
  * @param[in] pcName cache entry key
- * @param[in] ulTTL time to live
+ * @param[in] ulTTL time to live (in seconds)
  * @param[in] pulIP ip address
  * @param[in] ulCurrentTimeSeconds current time
  * @post the global structure \a xDNSCache is modified
--- a/DNS/DNS_Cache.c
+++ b/DNS/DNS_Cache.c
@@ -81,14 +81,14 @@
                                           uint32_t * pulIP,
                                           uint32_t ulCurrentTimeSeconds );
 
-    static void prvUpdateCacheEntry( int index,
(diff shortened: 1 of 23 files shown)


DNS/DNS_Cache.c
```diff
@@ -123,10 +123,12 @@
                                     uint32_t * pulIP,
```
AniruddhaKanhere, line 123: In the @brief, we need to add the return value description. Also, why didn't doxygen flag this? This file might not be getting included in the doxygen config.

DNS/DNS_Cache.c
```diff
@@ -301,9 +301,9 @@
  * @post the global structure \a xDNSCache is modified
  *
  */
-    static void prvUpdateCacheEntry( int index,
-                                     int ulTTL,
-                                     int32_t * pulIP,
+    static void prvUpdateCacheEntry( BaseType_t index,
```
AniruddhaKanhere, line 304: Same comment as above - why not a `UBaseType_t`?

FreeRTOS_DNS.c
```diff
@@ -468,38 +452,36 @@
...
         #if ( ipconfigUSE_LLMNR == 1 )
-            BaseType_t bHasDot = llmnr_has_dot( pcHostName );
-        #endif /* ipconfigUSE_LLMNR == 1 */
+            bHasDot = llmnr_has_dot( pcHostName );
 
-        #if ( ipconfigUSE_LLMNR == 1 )
             if( bHasDot == pdFALSE )
             {
                 /* Use LLMNR addressing. */
-                ( ipCAST_PTR_TO_TYPE_PTR( DNSMessage_t, pucUDPPayloadBuffer ) )->usFlags = 0;
```
AniruddhaKanhere, line 491: Why was this statement removed? Do we not need to reset flags?

FreeRTOS_DNS.c
```diff
@@ -538,47 +520,99 @@
...
+
+        return xReturn;
+    }
+
 /*!
  * @brief main dns operation description function
  *
  * @param [in] pcHostName hostname to get its ip address
- * @param [in] uxIdentifier
+ * @param [in] uxIdentifier identifier
```
AniruddhaKanhere, line 584: I am not sure that I understand what the description identifier means? What ID is this?

FreeRTOS_IP_Timers.c
```diff
@@ -484,3 +484,17 @@ void vIPSetARPResolutionTimerEnableState( BaseType_t xEnableState )
...
+/**
+ * @brief Reload the DNS timer.
+ *
+ * @param[in] ulCheckTime: The reload value.
+ */
+    void vIPReloadDNSTimer( uint32_t ulCheckTime )
+    {
+        prvIPTimerReload( &xDNSTimer, ulCheckTime );
+    }
+#endif /* ipconfigDNS_USE_CALLBACKS != 0 */
```
AniruddhaKanhere, line 499: There is already a definition at line 338 (albeit with a different name). I renamed the function to make all reloading functions have similar names.

FreeRTOS_DNS.c
alfred2g: I put it somewhere else
as with the changes at this point, we did not create this buffer yet/

htibosch:
Replying to LLMNR requests is still broken. Please have a look at this version of [DNS_Parser.zip](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/7882801/DNS_Parser.zip)

Aniruddha explained to me that the const variable `xBufferAllocFixedSize` is a bit difficult for unit testing. Coverity also explains about such code. Yet it is the only way to determine if BufferAllocation_1.c or BufferAllocation_2.c has been linked to the project.

And yes, it is stupid to write:
~~~c
    if( 1 )
    {
        /* Do something */
    }
    else
    {
        /* Do something else */
    }
~~~

We didn't want to force developers to `#define` something like "`USING_BUFFERALLOCATION_1`".

htibosch:
The problem with the network buffer seems to be solved. However, the branch [fix_dns_comments](https://github.com/alfred2g/FreeRTOS-Plus-TCP/tree/fix_dns_comments) is still not functional. Looking up an address always "fails" and it can not reply to an LLMNR request when BufferAllocation_1.c is being used, the TTL value in the cache were always zero, and a few more issues.

Here is a version that should correct the problems found: [DNS_version_hein_18_jan_2022.zip](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/7886322/DNS_version_hein_18_jan_2022.zip)

Here below a summary of the changes:

DNS/DNS_Cache.c
---

● Changed parameter name from "index" to "uxIndex"
● FreeRTOS_dns_update() must pass 'ulTTL' to FreeRTOS_ProcessDNSCache() so it gets stored in the cache
● Corrected a format when calling FreeRTOS_debug_printf()
● Corrected the age calculation to avoid problem when the timer goes through zero:
~~~c
- if( ulCurrentTimeSeconds < ( xDNSCache[ index ].ulTimeWhenAddedInSeconds +
-                            FreeRTOS_ntohl( xDNSCache[ index ].ulTTL ) ) )
+ uint32_t ulAge = ulCurrentTimeSeconds - xDNSCache[ uxIndex ].ulTimeWhenAddedInSeconds;
+ if( ulAge < FreeRTOS_ntohl( xDNSCache[ uxIndex ].ulTTL ) )
~~~

DNS/DNS_Callback.c
---

● Corrected a format when calling FreeRTOS_debug_printf()


DNS/DNS_GLobals.h
---

● Changed the value of `dnsLLMNR_TTL_VALUE` from 300,000 to 300, because the unit is seconds and not msec.


DNS/DNS_Parser.c
---

● `parseDNSAnswer()` should return the IP-address found, not just a boolean that indicates success or failure
● Please change the handling of `xBufferAllocFixedSize` as in my example. LLMNR is still broken.


DNS/DNS_Parser.h
---

● Changed the return type of `parseDNSAnswer()` to return an IPv4 address.
● The following would give a compiler warning:
~~~c
    BaseType_t bHasDot = pdFALSE;
    bHasDot = llmnr_has_dot( pcHostName );
~~~
● 'ipLLMNR_PORT' is defined as a host-endian value. Please note the following correction:
~~~c
- if( pxAddress->sin_port == ipLLMNR_PORT )
- {
-     ( ipCAST_PTR_TO_TYPE_PTR( DNSMessage_t, xDNSBuf.pucPayloadBuffer ) )->usFlags = 0;
- }
+ #if ( ipconfigUSE_LLMNR == 1 )
+ {
+     if( FreeRTOS_ntohs( pxAddress->sin_port ) == ipLLMNR_PORT )
+     {
+         ( ipCAST_PTR_TO_TYPE_PTR( DNSMessage_t, xDNSBuf.pucPayloadBuffer ) )->usFlags = 0;
+     }
+ }
+ #endif
~~~

include/FreeRTOS_IP.h
---

● It wuld be nice if it \#includes "FreeRTOS_IP_Utils.h"

And finally, I would like to get rid of `DNS_SkipNameField()` and only use `DNS_ReadNameField()`. The IPv6 branch already has [this change](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/74930044aba109d4886d61e4107da97fc93e517f/FreeRTOS_DNS.c#L2574).

htibosch:
Thank you Alfred so far! I hope that my changes don't disrupt the unit-testing too much.
I will give it another try and report back here.
Hein

alfred2g:
Thanks @htibosch for your review, and changes.
some little changes were made to the UT,  after the bug fixes, which is a good thing  :)

htibosch (dismissed):
I retested the PR in real hardware and it looks all good now, thank you.
As said, I only looked at the changes to the library, not to the unit-testing.
I tested DNS lookups on the net, also a site that has multiple IP-addresses, LLMNR passive and active lookup, followed by pings. I also tested the asynchronous lookups.

DNS/DNS_Cache.c
```diff
@@ -117,6 +117,8 @@
  * @param pcName the lookup name
  * @param pulIP the ip value to insert/replace
  * @param ulTTL ignored
```
AniruddhaKanhere, line 119: Is this parameter still ignored?

DNS/DNS_Cache.c
```diff
@@ -226,7 +231,7 @@
         int x;
 
         /* For each entry in the DNS cache table. */
-        for( x = 0; x < ipconfigDNS_CACHE_ENTRIES; x++ )
+        for( x = 0; x < ( int ) ipconfigDNS_CACHE_ENTRIES; x++ )
```
AniruddhaKanhere, line 234: Apologies for being a stickler for convention - but why not make the variable `x` unsigned rather than casting the value in test expression?

AniruddhaKanhere: I see the rationale: `index = x;`.
We'll need to cast at one place or another.

DNS/DNS_Cache.c
```diff
@@ -226,7 +231,7 @@
         int x;
```
AniruddhaKanhere, line 231: Also, since the return type is BaseType, can we make the int to a BaseType? (on this line and the line above).
Thanks

DNS/DNS_Cache.c
```diff
@@ -364,4 +368,34 @@
             }
         }
     }
+
+    #if 0
```
AniruddhaKanhere, line 372: Is this piece of code not needed?

DNS/DNS_Parser.c
```diff
@@ -585,16 +584,16 @@
  * @param xDoStore whether to update the cache
  * @return  \c pdTRUE if the frame is correct or \c pdFALSE otherwise
```
AniruddhaKanhere, line 585: The comment might need updating

AniruddhaKanhere (changes requested):
Some minor comments

AniruddhaKanhere (approved):

paulbartell (approved):

PR #407, opened 2022-01-18
Unit-tests for FreeRTOS_Sockets.c file

AniruddhaKanhere:
Description
-----------
This PR adds unit-test for the file FreeRTOS_Sockets.c with some minor changes to the source code itself.

Test Steps
-----------

Related Issue
-----------


alfred2g (approved):

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
```diff
@@ -58,6 +59,72 @@
 
 #include "FreeRTOSIPConfig.h"
 
+extern List_t xBoundUDPSocketsList;
```
paulbartell: This file is huge. It would probably be a good idea to split it up a bit.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
```diff
@@ -151,3 +218,7366 @@ void test_FreeRTOS_inet_pton4( void )
     TEST_ASSERT_EQUAL( pdFAIL, xResult );
     TEST_ASSERT_EQUAL_MESSAGE( ulInValidResponse, ulIPAddress, "Incorrectly converted string 8." );
 }
+
+void test_prvValidSocket_InvalidOrNULLSocket( void )
+{
+    BaseType_t xReturn;
+    FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) FREERTOS_INVALID_SOCKET;
+    BaseType_t xProtocol, xIsBound;
```
paulbartell: In this test case and other following test cases, xProtocol is not defined to a particular value.

paulbartell: Should be either FREERTOS_IPPROTO_UDP or FREERTOS_IPPROTO_TCP.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
```diff
@@ -151,3 +218,7366 @@ void test_FreeRTOS_inet_pton4( void )
...
+    size_t xSocketSize;
+
+    xIPIsNetworkTaskReady_ExpectAndReturn( pdFALSE );
+
+    xReturn = prvDetermineSocketSize( xDomain, xType, xProtocol, &xSocketSize );
+
+    TEST_ASSERT_EQUAL( pdFAIL, xReturn );
+}
+
+void test_prvDetermineSocketSize_CatchAssert( void )
```
paulbartell: Which assertion are you catching here?

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: True, however in this case, the protocol doesn't matter since the list check will return a NULL and the function will exit.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: That the domain should always be `FREERTOS_AF_INET`.

FreeRTOS_Sockets.c
```diff
@@ -1297,6 +1297,9 @@ BaseType_t vSocketBind( FreeRTOS_Socket_t * pxSocket,
         struct freertos_sockaddr xAddress;
     #endif /* ipconfigALLOW_SOCKET_SEND_WITHOUT_BIND */
 
+    configASSERT( pxSocket != NULL );
+    configASSERT( pxSocket != FREERTOS_INVALID_SOCKET );
```
htibosch, line 1301: Good catch to move these asserts to a location where `pxSocket` has not been used yet.

FreeRTOS_Sockets.c
```diff
@@ -1965,21 +1965,17 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                            break; /* will return -pdFREERTOS_ERRNO_EINVAL */
                        }
 
-                       if( ( pxTCP->txStream != NULL ) || ( pxTCP->rxStream != NULL ) )
-                       {
-                           FreeRTOS_debug_printf( ( "Set SO_WIN_PROP: buffer already created\n" ) );
-                           break; /* will return -pdFREERTOS_ERRNO_EINVAL */
-                       }
```
htibosch, line 1972: Also good: `prvSockopt_so_buffer()` will check if the stream buffers have already been created.

FreeRTOS_Sockets.c
```diff
@@ -2301,16 +2296,6 @@ const char * FreeRTOS_inet_ntoa( uint32_t ulIPAddress,
...
 
-        /* Write e.g. "192.", which is 3 digits and a dot. */
-        uxNeeded = ( ( socklen_t ) sockDIGIT_COUNT - uxSource ) + 1U;
-
-        if( ( uxIndex + uxNeeded ) > uxSize )
-        {
-            /* The result won't fit. */
-            pcResult = NULL;
-            break;
-        }
```
htibosch, line 2312: We already talked about this change and agreed on it.

FreeRTOS_Sockets.c
```diff
@@ -3067,13 +3052,6 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                 /* Did it get connected while sleeping ? */
                 xResult = FreeRTOS_issocketconnected( pxSocket );
 
-                /* Returns positive when connected, negative means an error */
-                if( xResult < 0 )
-                {
-                    /* Return the error */
-                    break;
-                }
```
htibosch, line 3075: The socket has already been checked, so `FreeRTOS_issocketconnected()` won't return an error.

htibosch (dismissed):
Thanks @AniruddhaKanhere for this PR that adds unit testing to FreeRTOS_Sockets.c
I only checked the source file FreeRTOS_Sockets.c and I approve the changes.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: How do you propose we do that?

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: From some later test cases, you'll see that the protocol value is set since that is what is being tested at that point

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
paulbartell: Perhaps a file per function or set of functions?

Regardless, there are some config definitions that are mutually exclusive, so should probably be tested separately.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
paulbartell: In this case, xProtocol is whatever is currently on the stack in that position, which could lead to test runs that are not reproducible.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
paulbartell: Could you adjust the same of this test case to make that clear or add a comment?

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
```diff
@@ -151,3 +218,7366 @@ void test_FreeRTOS_inet_pton4( void )
...
+
+void test_prvValidSocket_InvalidProtocol( void )
+{
+    BaseType_t xReturn;
+    FreeRTOS_Socket_t xSocket;
+    BaseType_t xProtocol, xIsBound = pdTRUE;
+
+    memset( &xSocket, 0, sizeof( FreeRTOS_Socket_t ) );
+
+    listLIST_ITEM_CONTAINER_ExpectAndReturn( &( xSocket.xBoundSocketListItem ), ( struct xLIST * ) ( uintptr_t ) 0x11223344 );
```
paulbartell: Using ExpectAndReturn for all of these list operations seems very cumbersome. A test double for the list api would be a less brittle option.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: done

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: Sometimes we want lists to return values without initializing all the values in the list. The expectAndReturn serves that purpose well. I can Pass a list which is just a pointer without any memory but still get a value out of it.
Additionally, we have this across all the files in unit-tests. We'll need a separate PR to address this issue.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: Done - I split the tests into multiple files.
For the configuration, I am planning to add more tests later.

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_utest.c
AniruddhaKanhere: Got it. Fixed.

alfred2g (approved):

n9wxu (approved):

PR #408, opened 2022-01-20
Make Ultrascale port more generic and add Jumbo Frame support

StefanBalt:
Make the Ultrascale port more versatile.

Description
-----------
The first commit replaces calls of the Xilinx function xil_printf() with the generic FreeRTOS_printf(). This is exactly how prints are done in the Zynq7000 port. This allows the developer to implement an arbitrary debug output independent of Xilinx drivers.

The second commit detects if the requested MTU requires Jumbo Frames and enables Jumbo Frame support if necessary. The use of large MTU's (and thus Jumbo Frames) can significantly reduce the CPU load as well as reduce the number of packets lost.

Test Steps
-----------
The Jumbo Frames worked fine using UDP up to a packet size of 7KB. However, for bigger packets it did not work.
I still think that it is worth merging because using Jumbo frames allowed me to double my data-rate from 300 to 600 Mbit/s (same CPU load) while drastically reducing packet losses (probably on the server side).

### DIFF
--- a/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
+++ b/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
@@ -70,12 +70,21 @@
     #define PHY_LS_LOW_CHECK_TIME_MS    1000
 #endif
 
+#if ( ipconfigNETWORK_MTU > 1526 )
+    #warning the use of Jumbo Frames has not been tested sufficiently yet.
+    #define USE_JUMBO_FRAMES    1
+#endif
+
 /* The size of each buffer when BufferAllocation_1 is used:
  * http://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/Embedded_Ethernet_Buffer_Management.html */
-#define niBUFFER_1_PACKET_SIZE    1536
+#if ( USE_JUMBO_FRAMES == 1 )
+    #define niBUFFER_1_PACKET_SIZE    10240
+#else
+    #define niBUFFER_1_PACKET_SIZE    1536
+#endif
 
 /* Naming and numbering of PHY registers. */
-#define PHY_REG_01_BMSR           0x01  /* Basic mode status register */
+#define PHY_REG_01_BMSR    0x01         /* Basic mode status register */
 
 #ifndef iptraceEMAC_TASK_STARTING
     #define iptraceEMAC_TASK_STARTING()    do {} while( 0 )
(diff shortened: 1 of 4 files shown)


htibosch:
@AniruddhaKanhere :

> 1 workflow awaiting approval
> First-time contributors need a maintainer to approve running workflows.

Aniruddha can you activate the automatic testing for @StefanBalt ?

~~~c
-    xil_printf( "Error during reset \n\r" );
+    FreeRTOS_printf( ( "Error during reset \n\r" ) );
~~~
That is indeed a welcome change.  Also because `xil_printf()` is not compatible with multitasking.

Jumbo frames:
---
It's funny, I have always stayed away from using jumbo frames because you can never be sure if the peer applications, peer hosts and the routers in between also support them.

TCP exchanges the MSS ( Maximum Segment Size ) at start-up, but for UDP there is not a standard way of knowing the MTU of the peer ( I think ).

You are right of course about the advantages, so let's go for it!
My 64-bit Ultrascale was already superfast when using an MTU of 1500, but other platforms ( my MicroZed ) will certainly profit from the changes.

I will also test the changes in real hardware. I see that my laptop can be configured to support jumbo frames with an MTU between 2 and 9KB. Not sure if the maximum is 9,000 or 9,216 bytes.

> The Jumbo Frames worked fine using UDP up to a packet size of 7KB. However, for bigger packets it did not work.

Is that due to hardware in-between? Does the target host allow jumbo frames larger than 7KB?

I will test the changes and come back to it. Thanks so far.

StefanBalt:
>  It's funny, I have always stayed away from using jumbo frames because you can never be sure if the peer applications, peer hosts and the routers in between also support them.

For point to point high bandwidth streaming applications I really benefited on both the sender and receiver side (Winsock host) from using Jumbo Frames. It's now the memory bandwidth of the CR5 and not the network interface that's holding me back.

> Is that due to hardware in-between? Does the target host allow jumbo frames larger than 7KB?

There was no HW in between. The host was a network card running on Windows which allows packets up to 9014 bytes according to the driver. I do not know what the exact maximum is where it stops working. 7KB (7168 bytes) worked and 8KByte did not work anymore.

portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -70,12 +70,21 @@
     #define PHY_LS_LOW_CHECK_TIME_MS    1000
 #endif
 
+#if ( ipconfigNETWORK_MTU > 1526 )
+    #warning the use of Jumbo Frames has not been tested sufficiently yet.
```
htibosch, line 74: I have tested your PR with an Ultrascale, connected through a switch to a Win 10 laptop.
I have created a command:
~~~
    jumbo send <length>
~~~
to send UDP messages of any length. My MTU is defined as 9000.

And as expected, the maximum payload size was 8972. Which leads to an Ethernet packet of 9014 bytes:
~~~c
Ethernet   14
IP         20
UDP         8
Payload  8972
-------------
         9014
~~~
It all looks good to me, and it works.

So if we leave in the \#warning about the use of jumbo frames, I approve this PR, and thank you for it.

htibosch (approved):
Thank you for this PR, I approve it.

AniruddhaKanhere (dismissed):

portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
StefanBalt: Thank you for testing. Interesting that it did not work for me for with payload size 8KB.
May also be a problem on my host side. I will test this a lot more in a few months.

paulbartell:
Squashed the formatting changes into the original commits for clarity.

paulbartell (approved):

PR #409, opened 2022-01-24
Fix FreeRTOS_closesocket when callbacks are enabled and TCP is disabled

holden-zenith:
Fix FreeRTOS_closesocket when callbacks are enabled and TCP is disabled

Changed pxSocket definition boolean logic from && to || otherwise pxSocket is undefined when clearing the UDP handlers.

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -1439,9 +1439,9 @@ BaseType_t FreeRTOS_closesocket( Socket_t xSocket )
 {
     BaseType_t xResult;
 
-    #if ( ipconfigUSE_TCP == 1 ) && ( ipconfigUSE_CALLBACKS == 1 )
+    #if ( ipconfigUSE_CALLBACKS == 1 )
         FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
-    #endif
+    #endif /* ipconfigUSE_CALLBACKS == 1 */
     IPStackEvent_t xCloseEvent;
     xCloseEvent.eEventType = eSocketCloseEvent;
     xCloseEvent.pvData = xSocket;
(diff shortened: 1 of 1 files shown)


FreeRTOS_Sockets.c
```diff
@@ -1439,7 +1439,7 @@ BaseType_t FreeRTOS_closesocket( Socket_t xSocket )
 {
     BaseType_t xResult;
 
-    #if ( ipconfigUSE_TCP == 1 ) && ( ipconfigUSE_CALLBACKS == 1 )
+    #if ( ipconfigUSE_TCP == 1 ) || ( ipconfigUSE_CALLBACKS == 1 )
         FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
```
htibosch, line 1443: Thank you for this change. But looking a second time, I think that testing `ipconfigUSE_TCP` is not needed:
~~~c
    #if ( ipconfigUSE_CALLBACKS == 1 )
        FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
    #endif
~~~
because when `ipconfigUSE_CALLBACKS` are disabled, `pxSocket ` will not be referred too?

htibosch (commented):
Thank you @holden-zenith for this PR. I only have one suggestion for change.
Hein

FreeRTOS_Sockets.c
holden-zenith: You are correct that makes more sense, fixed. Also, I believe the UDP receive callback is never called?

htibosch (approved):
Thank you very much, @holden-zenith , that will safe another compiler warning :-)
I approve it.

AniruddhaKanhere (approved):
Thank you @holden-zenith for this fix and taking the time to contribute to FreeRTOS-Plus-TCP.
We shall merge this change into mainline soon.

cobusve (approved):

Issue #414, opened 2022-01-31
FreeRTOS TCP V2.4.0: DHCP Fails


RuStrannik:
So, before, while I've been using FreeRTOS TCP V2.3.3 my device used to cold boot (fist time power on) with following code:

```
int main(void) {
	bsd_init();
	const uint8_t addr_dummy[4] = { 0, 0, 0, 0 };
	FreeRTOS_IPInit(addr_dummy, addr_dummy, addr_dummy, addr_dummy, bsd_get_MAC_Addr());
	// {"create tasks"}
	vTaskStartScheduler();
	bsd_critical_fail_trap(__func__);
};//main()
```
And everything worked flawlessly, until I tried FreeRTOS TCP V2.4.0, which caused the following problem: the system completely fails to digest received DHCP responce. Inifinitely.

Debugging revealed the following problem:
Before, in `FreeRTOSv202107.00/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_UDP_IP.c`, Line 318
`vARPRefreshCacheEntry()` was called unconditionally. And everything worked.

Now, in file `FreeRTOSv202112.00/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_UDP_IP.c`, Line 322
`xCheckRequiresARPResolution()` is called first instead. Inside, it checks if' 'SourceIPAddress' belongs to the same subnet. Obviously, when network is not initialized yet, this check fails, so network never gets initialized.

**The solution is to call `FreeRTOS_IPInit()` with NetMask of `255.255.255.255` on cold (first) boot.**

I would not call it a bug, but rather a significant change in behavior, which was not reflected in release notes and caused quite a bit of time loss.

So, please either mention this change in release notes, or allow skipping `xCheckRequiresARPResolution()` before network is initialized (e.g. `if (*ipLOCAL_IP_ADDRESS_POINTER == 0) { /* skip ARP processing */ }`).

htibosch:
Hello @RuStrannik, I am sorry to hear that you had these problems with +TCP V2.4.0.
And thanks a lot for reporting this. We were not yet aware of this change in behaviour.
I will sort it out and come back to it.
Hein

htibosch:
As you see here above, I already created a PR to repair this.
Thanks for sorting out the problem and reporting about it.
Hein

RuStrannik:
[...]

PR #415, opened 2022-01-31
No ARP verification during DHCP phase

htibosch:
Description
-----------
This PR addresses [issue 414](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/414) raised by @RuStrannik.

In his projects he calls `FreeRTOS_IPInit()` with zero's for all of the IP-addresses and netmask. That worked fine, until PR #338 was submitted: "Send ARP packets for unknown IP addresses". That PR made sure that the IP-addresses of incoming packets are verified by looking it up using ARP.

With the current PR, the ARP verification will be only be done after DHCP has set a proper IP-address and netmask.

Test Steps
-----------
Set `ipconfigUSE_DHCP` to 1 in your FreeRTOSIPConfig.h
Initialise the IP-stack as show in the issue:
~~~c
    const uint8_t addr[4] = { 0, 0, 0, 0 };
    FreeRTOS_IPInit(addr, addr, addr, addr, ucMacAddress);
~~~
Now boot the device with or without the change in FreeRTOS_UDP_IP.c

---
Most developers **never noticed** the above problem because `FreeRTOS_IPInit()` is normally not called with a zero netmask.

Thank you @RuStrannik for taking the effort to create the issue.

### DIFF
--- a/FreeRTOS_UDP_IP.c
+++ b/FreeRTOS_UDP_IP.c
@@ -319,19 +319,27 @@ BaseType_t xProcessReceivedUDPPacket( NetworkBufferDescriptor_t * pxNetworkBuffe
     {
         if( pxSocket != NULL )
         {
-            if( xCheckRequiresARPResolution( pxNetworkBuffer ) == pdTRUE )
+            if( *ipLOCAL_IP_ADDRESS_POINTER != 0U )
             {
-                /* Mark this packet as waiting for ARP resolution. */
-                *pxIsWaitingForARPResolution = pdTRUE;
+                if( xCheckRequiresARPResolution( pxNetworkBuffer ) == pdTRUE )
+                {
+                    /* Mark this packet as waiting for ARP resolution. */
+                    *pxIsWaitingForARPResolution = pdTRUE;
 
-                /* Return a fail to show that the frame will not be processed right now. */
-                xReturn = pdFAIL;
-                break;
+                    /* Return a fail to show that the frame will not be processed right now. */
+                    xReturn = pdFAIL;
+                    break;
+                }
+                else
+                {
+                    /* Update the age of this cache entry since a packet was received. */
+                    vARPRefreshCacheEntryAge( &( pxUDPPacket->xEthernetHeader.xSourceAddress ), pxUDPPacket->xIPHeader.ulSourceIPAddress );
+                }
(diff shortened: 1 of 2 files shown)


FreeRTOS_UDP_IP.c
```diff
@@ -319,19 +319,27 @@ BaseType_t xProcessReceivedUDPPacket( NetworkBufferDescriptor_t * pxNetworkBuffe
     {
         if( pxSocket != NULL )
         {
-            if( xCheckRequiresARPResolution( pxNetworkBuffer ) == pdTRUE )
+            if( *ipLOCAL_IP_ADDRESS_POINTER != 0U )
```
AniruddhaKanhere, line 322: I would like it more if we can move this check to `xCheckRequiresARPResolution` [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/ab0d4f84beee3acce2097020ef631169ac55f618/FreeRTOS_ARP.c#L404).

Because, the function `xCheckRequiresARPResolution` is supposed to return `pdFALSE`  if ARP resolution should not be done. And when it returns a `pdFALSE`, we'll just update the ARP cache entry's age and move on to proceed with processing of the packet.

What do you think @htibosch?

AniruddhaKanhere (commented):
Thank you for this PR @htibosch.
I have a comment. Would you mind taking a look?

Thank you.

FreeRTOS_UDP_IP.c
htibosch: During a DHCP negotiation, ARP does not play a role, and therefore I would not even call `xCheckRequiresARPResolution()`.
It doesn't really matter much, because `vARPRefreshCacheEntryAge()` would probably not find the IP-address to be refreshed.

AniruddhaKanhere: Very well. I am satisfied with that argument.

Thank you @htibosch. Would you mind making the change to unit-tests or shall I do that on your behalf?

AniruddhaKanhere (approved):

htibosch:
> @htibosch, I took the liberty of fixing the unit-tests.

I am very grateful for that.

paulbartell (approved):

PR #416, opened 2022-02-01
Make WAKEUP_CALLBACK and SET_SEMAPHORE accessible when TCP is disabled

holden-zenith:
Make WAKEUP_CALLBACK and SET_SEMAPHORE accessible when TCP is disabled

These options were not available to UDP in setsockopt when TCP is disabled. UDP should call wakeup callback after a bind event.

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -1897,31 +1897,29 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                     break;
             #endif /* ipconfigUSE_CALLBACKS */
 
-            #if ( ipconfigUSE_TCP != 0 )
-                #if ( ipconfigSOCKET_HAS_USER_SEMAPHORE != 0 )
+            #if ( ipconfigSOCKET_HAS_USER_WAKE_CALLBACK != 0 )
+                case FREERTOS_SO_WAKEUP_CALLBACK:
+
+                    /* Each socket can have a callback function that is executed
+                     * when there is an event the socket's owner might want to
+                     * process. */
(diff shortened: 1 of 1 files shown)


FreeRTOS_Sockets.c
```diff
@@ -1897,31 +1897,29 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
...
                      * sleeps. */
-                    case FREERTOS_SO_SET_SEMAPHORE:
-                       {
-                           pxSocket->pxUserSemaphore = *( ipPOINTER_CAST( SemaphoreHandle_t *, pvOptionValue ) );
-                       }
-                        xReturn = 0;
-                        break;
-                #endif /* ipconfigSOCKET_HAS_USER_SEMAPHORE */
-
-                #if ( ipconfigSOCKET_HAS_USER_WAKE_CALLBACK != 0 )
```
htibosch, line 1913: You took the 2 cases 'FREERTOS_SO_WAKEUP_CALLBACK' and 'FREERTOS_SO_SET_SEMAPHORE' out of an \#if block 'ipconfigUSE_TCP'. That is good.

But I don't know why you swapped the order of the two cases? Would you mind to swap them back as shown here?
~~~c
    case FREERTOS_SO_SET_SEMAPHORE:
    case FREERTOS_SO_WAKEUP_CALLBACK:
~~~

Thanks

htibosch (commented):
If you are so kind to swap the two switch cases, I will approve this PR.
I test-compiled it with a UDP-only bootloader.
Thanks

AniruddhaKanhere (approved):

paulbartell (approved):

PR #424, opened 2022-02-14
Handle zero protocol argument in FreeRTOS_socket()

Zalasus:
Description
-----------

In the Berkeley Socket API, a protocol argument of 0 indicates to the socket layer that it should pick a sensible default protocol. 

I simply added a check to `FreeRTOS_socket()` that catches an `xProtocol` of zero and sets it to an appropriate value based off the `xType` argument. Nothing fancy.

Test Steps
-----------

Call `FreeRTOS_socket(FREERTOS_AF_INET, FREERTOS_SOCK_DGRAM, 0)`. It will return `FREERTOS_INVALID_SOCKET`, whereas the equivalent Berkeley socket call would succeed.

### DIFF
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -387,7 +387,7 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
     EventGroupHandle_t xEventGroup;
     Socket_t xReturn;
 
-    /* A protocol of 0 indicates to the socket layer that it should pick a 
+    /* A protocol of 0 indicates to the socket layer that it should pick a
      * sensible default protocol based off the given socket type. If we can't,
      * prvDetermineSocketSize will catch it as an invalid type/protocol combo.
      */
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -388,11 +388,13 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
     Socket_t xReturn;
     BaseType_t xProtocolCpy = xProtocol;
 
-    /* A protocol of 0 indicates to the socket layer that it should pick a
-     * sensible default protocol based off the given socket type. If we can't,
-     * prvDetermineSocketSize will catch it as an invalid type/protocol combo.
+    /* The special protocol FREERTOS_SOCK_DEPENDENT_PROTO, which is equivalent
+     * to passing 0 as defined by POSIX, indicates to the socket layer that it
+     * should pick a sensible default protocol based off the given socket type.
+     * If we can't, prvDetermineSocketSize will catch it as an invalid
+     * type/protocol combo.
      */
-    if( xProtocol == 0 )
+    if( xProtocol == FREERTOS_SOCK_DEPENDENT_PROTO )
     {
         switch( xType )
         {
--- a/FreeRTOS_Sockets.c
+++ b/FreeRTOS_Sockets.c
@@ -386,6 +386,7 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
     size_t uxSocketSize = 1;
     EventGroupHandle_t xEventGroup;
     Socket_t xReturn;
+    BaseType_t xProtocolOrDefault;
 
     /* A protocol of 0 indicates to the socket layer that it should pick a
      * sensible default protocol based off the given socket type. If we can't,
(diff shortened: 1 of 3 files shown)


FreeRTOS_Sockets.c
```diff
@@ -387,6 +387,24 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
     EventGroupHandle_t xEventGroup;
     Socket_t xReturn;
 
+    /* A protocol of 0 indicates to the socket layer that it should pick a 
+     * sensible default protocol based off the given socket type. If we can't,
+     * prvDetermineSocketSize will catch it as an invalid type/protocol combo.
+     */
+    if( xProtocol == 0 )
```
AniruddhaKanhere, line 394: As an 'aside'. Maybe we should convert 0 to a Macro?

AniruddhaKanhere (commented):
Hello @Zalasus,

Thank you for this PR. I have a couple of comments. Can you please take a look at them?

Thanks,
Aniruddha

FreeRTOS_Sockets.c
Zalasus: While I'm also not a huge fan of magic numbers, the POSIX description of socket() explicitly mentions the numeric value 0. Linux defines IPPROTO_IP as 0 for this purpose, but I don't find that name particularly obvious. Still, I'll gladly patch it in if you want.

FreeRTOS_Sockets.c
```diff
@@ -387,6 +387,24 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
...
+    {
+        switch( xType )
+        {
+            case FREERTOS_SOCK_DGRAM:
+                xProtocol = FREERTOS_IPPROTO_UDP;
```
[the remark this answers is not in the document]
Zalasus: Sorry, I was not aware there was a MISRA rule for that. I added a new commit addressing the issue.

FreeRTOS_Sockets.c
AniruddhaKanhere: Yes, I understand. Users will get to use 0 directly. But while someone is reading the code, one might wonder what `0` is about if they are not familiar with 0 being used in POSIX sockets (as I wasn't either). 
But generally macros look better. Maybe we should come up with a good descriptive name for the macro. Something like `FREERTOS_SOCK_DEPENDENT_PROTO`.

AniruddhaKanhere (commented):
@Zalasus, thanks for the quick fix.

I have one response to your question. I am fine either way - we can go with `0` or a macro.

FreeRTOS_Sockets.c
```diff
@@ -386,8 +386,31 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
...
+        switch( xType )
+        {
+            case FREERTOS_SOCK_DGRAM:
+                xProtocolOrDefault = FREERTOS_IPPROTO_UDP;
+                break;
+
+            case FREERTOS_SOCK_STREAM:
+                xProtocolOrDefault = FREERTOS_IPPROTO_TCP;
+                break;
+        }
```
htibosch, line 406: Oops, if `xType` is undefined, `xProtocolOrDefault` will also become undefined.
Maybe just initialise it at the declaration?
~~~c
BaseType_t xProtocolOrDefault = xProtocol;
~~~
and forget about the `else`?

And a minor remark about the name `xProtocolOrDefault`: actually the parameter can be either 0 or denote a protocol. The variable `xProtocolOrDefault` should never be zero.

In other functions we solved the MISRA rule by introducing a new name follow by Copy, such as:
~~~c
    BaseType_t xProtocolCpy = xProtocol;
~~~

htibosch (approved):
Except for some remarks here below, I approve this PR.
Thanks a lot!

FreeRTOS_Sockets.c
Zalasus: I see your point. I addressed the issue.

FreeRTOS_Sockets.c
Zalasus: I went ahead and used your suggestion for a macro name :)

FreeRTOS_Sockets.c
```diff
@@ -386,8 +386,29 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
...
+    {
+        switch( xType )
+        {
+            case FREERTOS_SOCK_DGRAM:
+                xProtocolCpy = FREERTOS_IPPROTO_UDP;
+                break;
+
+            case FREERTOS_SOCK_STREAM:
+                xProtocolCpy = FREERTOS_IPPROTO_TCP;
+                break;
```
AniruddhaKanhere, line 407: One more comment here. Apologies that I did not catch this earlier and it is taking a bit long to get this PR merged.
MISRA rule 16.4 says that all switch statements must have a `default` label.
In this case, I think we can get away with just adding a comment because an incorrect `xType` will be caught by `prvDetermineSocketSize`.

FreeRTOS_Sockets.c
AniruddhaKanhere: Thanks @htibosch for catching this and thank you @Zalasus for fixing the issue.

AniruddhaKanhere (commented):
One last comment (hopefully). Apologies that I didn't catch this earlier.
You might also want to add `xprotocolcpy` to `.github/lexicon.txt`.

Thanks,
Aniruddha

FreeRTOS_Sockets.c
Zalasus: Yeah, I should have thought as much. Sorry, I should have read through some of these MISRA rules it seems. Didn't really think that much could be wrong with 10 lines of code :D

Anyways, I fixed the issue. Hopefully everything is in order now.

htibosch (dismissed):
> Didn't really think that much could be wrong with 10 lines of code :D

Haha. Things weren't really wrong, the comments were also addressing "style of coding".

I tested your PR in a real-live application : it compiled and ran without problems.

Thanks

PR #425, opened 2022-02-15
Support the KSZ8795 phy in the phyHandling driver.

ancantus:
Description
-----------
Support the [KSZ8795](https://www.microchip.com/en-us/product/KSZ8795) switch in the phyHandling driver.
Driver's MIIM interface phyAddress for the 4 ports starts at 0x1 instead of 0x0. 

Test Steps
-----------
Connected to the MIIM interface of a KSZ8795 switch.

Related Issue
-----------
N/A

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #426, opened 2022-02-21
Added Network Interface driver for TI TM4C1294NCPDT and Compiler supp…

jscott-hotstart:
…ort for TI Code Composer Studio.

Network Interface driver for TI TM4C1294NCPDT

Description
-----------
Added NIC driver for TM4C1294NCPDT to FreeRTOS+TCP stack.

Test Steps
-----------
Not an issue, but a contribution to the project. To test the implementation, acquire the TI EK-TM4C1294XL evaluation kit, compile FreeRTOS and FreeRTOS+TCP and include the TCP Echo Server module.

Related Issue
-----------
N/A

### DIFF
--- a/portable/NetworkInterface/TM4C/NetworkInterface.c
+++ b/portable/NetworkInterface/TM4C/NetworkInterface.c
@@ -1,0 +1,599 @@
+/**
+ *       @file: NetworkInterface.c
+ *	   @author: jscott <jscott@hotstart.com>
+ *       @date: Feb 1, 2022
+ *  @copyright: Hotstart 2022 Hotstart Thermal Management. All Rights Reserved.
+ *
+ *      Permission is hereby granted, free of charge, to any person obtaining a copy of
+ *      this software and associated documentation files (the "Software"), to deal in
+ *      the Software without restriction, including without limitation the rights to
+ *      use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ *      the Software, and to permit persons to whom the Software is furnished to do so,
+ *      subject to the following conditions:
+ *
+ *      The above copyright notice and this permission notice shall be included in all
+ *      copies or substantial portions of the Software.
+ *
+ *      THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *      IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ *      FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ *      COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ *      IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ *      CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ *      @brief:Network Interface driver for the Texas Instruments TM4C line of microcontrollers.
+ *
+ *      This driver was written and tested with the TM4C1294NCPDT, which includes a built-in MAC and
+ *      PHY. The expectation is that this driver should function correctly across all the MAC/PHY
+ *      integrated parts of the TM4C129X parts.
+ */
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+#include "inc/hw_ints.h"
+#include "inc/hw_emac.h"
+#include "inc/hw_memmap.h"
+#include "inc/hw_nvic.h"
+
+#include "driverlib/flash.h"
+#include "driverlib/interrupt.h"
+#include "driverlib/gpio.h"
+#include "driverlib/rom_map.h"
+#include "driverlib/sysctl.h"
+#include "driverlib/systick.h"
+#include "driverlib/emac.h"
+
+#include "FreeRTOS.h"
+#include "task.h"
+#include "queue.h"
+
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_Sockets.h"
+#include "FreeRTOS_IP_Private.h"
+#include "NetworkBufferManagement.h"
+#include "NetworkInterface.h"
+
+#define BUFFER_SIZE                             (ipTOTAL_ETHERNET_FRAME_SIZE + ipBUFFER_PADDING)
+#define BUFFER_SIZE_ROUNDED_UP                  ((BUFFER_SIZE + 7) & ~0x7UL)
+#define PHY_PHYS_ADDR                           0
+#define ETH_DEFERRED_INT_TASK_STACK_SIZE        512
+
+#define SYSCONFIG_SYSFREQ_HZ                            120000000
+#define SYSCONFIG_FREERTOS_TCP_TX_DESC_COUNT            8
+#define SYSCONFIG_FREERTOS_TCP_RX_DESC_COUNT            8
+#define SYSCONFIG_FREERTOS_TCP_MAX_PENDING_RX_PACKETS   SYSCONFIG_FREERTOS_TCP_RX_DESC_COUNT
+
+typedef struct {
+    uint32_t number_descriptors;
+    uint32_t write;
+    uint32_t read;
+} tDescriptorList;
+
+static tEMACDMADescriptor _tx_descriptors[SYSCONFIG_FREERTOS_TCP_TX_DESC_COUNT];
+static tEMACDMADescriptor _rx_descriptors[SYSCONFIG_FREERTOS_TCP_RX_DESC_COUNT];
+
+static tDescriptorList _tx_descriptor_list = { .number_descriptors = SYSCONFIG_FREERTOS_TCP_TX_DESC_COUNT, 0 };
+static tDescriptorList _rx_descriptor_list = { .number_descriptors = SYSCONFIG_FREERTOS_TCP_RX_DESC_COUNT, 0 };
+
+static uint8_t _network_buffers[ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS][BUFFER_SIZE_ROUNDED_UP];
+#pragma DATA_ALIGN(_network_buffers, 4)
+
+static QueueHandle_t _received_packets_queue = NULL;
+
+/**
+ * Reads the Ethernet MAC from user Flash.
+ * @param mac_address_bytes[out] The byte array which will hold the MAC address
+ * @return pdPASS on success, pdFAIL if the MAC is invalid from user Flash
+ */
+static BaseType_t _ethernet_mac_get(uint8_t *mac_address_bytes);
+
+/**
+ * Initialize DMA descriptors
+ */
+static void _dma_descriptors_init(void);
+
+/**
+ * Frees previously sent network buffers
+ */
+static void _process_transmit_complete(void);
+
+/**
+ * Processes received packets and forwards those acceptable to the network stack
+ */
+static void _process_received_packet(void);
+
+/**
+ * Processes PHY interrupts.
+ */
+static void _process_phy_interrupt(void);
+
+/**
+ * Thread to forward received packets from the ISR to the network stack
+ * @param parameters Not used
+ */
+static void _packet_received_thread(void *parameters);
+
+BaseType_t xNetworkInterfaceInitialise(void)
+{
+    uint8_t mac_address_bytes[6];
+    BaseType_t success;
+    uint16_t ui16Val;
+
+    _received_packets_queue = xQueueCreate(SYSCONFIG_FREERTOS_TCP_MAX_PENDING_RX_PACKETS, sizeof(NetworkBufferDescriptor_t*));
+
+    if (NULL == _received_packets_queue)
+    {
+        success = pdFAIL;
+    }
+    else
+    {
+        success = xTaskCreate(_packet_received_thread, "tcprx", 512, NULL, tskIDLE_PRIORITY + 5, NULL);
+    }
+
+    if (pdPASS == success)
+    {
+        success = _ethernet_mac_get(&mac_address_bytes[0]);
+    }
+
+    if (pdPASS == success)
+    {
+        MAP_SysCtlPeripheralReset(SYSCTL_PERIPH_EMAC0);
+        MAP_SysCtlPeripheralReset(SYSCTL_PERIPH_EPHY0);
+
+        while (!MAP_SysCtlPeripheralReady(SYSCTL_PERIPH_EMAC0))
+        {
+        }
+
+        MAP_EMACPHYConfigSet(
+                EMAC0_BASE,
+                EMAC_PHY_TYPE_INTERNAL | EMAC_PHY_INT_MDIX_EN
+                        | EMAC_PHY_AN_100B_T_FULL_DUPLEX);
+
+        MAP_EMACInit(EMAC0_BASE, SYSCONFIG_SYSFREQ_HZ,
+                     EMAC_BCONFIG_MIXED_BURST | EMAC_BCONFIG_PRIORITY_FIXED, 4,
+                     4, 0);
+
+        MAP_EMACConfigSet(
+                EMAC0_BASE,
+                (
+                    EMAC_CONFIG_100MBPS |
+                    EMAC_CONFIG_FULL_DUPLEX |
+                    EMAC_CONFIG_CHECKSUM_OFFLOAD |
+                    EMAC_CONFIG_7BYTE_PREAMBLE |
+                    EMAC_CONFIG_IF_GAP_96BITS |
+                    EMAC_CONFIG_USE_MACADDR0 |
+                    EMAC_CONFIG_SA_FROM_DESCRIPTOR |
+                    EMAC_CONFIG_BO_LIMIT_1024
+                ),
+                (
+                    EMAC_MODE_RX_STORE_FORWARD |
+                    EMAC_MODE_TX_STORE_FORWARD |
+                    EMAC_MODE_RX_THRESHOLD_64_BYTES |
+                    EMAC_MODE_TX_THRESHOLD_64_BYTES),
+                0);
+
+
+        // Clear any stray PHY interrupts that may be set.
+        ui16Val = MAP_EMACPHYRead(EMAC0_BASE, PHY_PHYS_ADDR, EPHY_MISR1);
+        ui16Val = MAP_EMACPHYRead(EMAC0_BASE, PHY_PHYS_ADDR, EPHY_MISR2);
+
+        // Configure and enable the link status change interrupt in the PHY.
+        ui16Val = MAP_EMACPHYRead(EMAC0_BASE, PHY_PHYS_ADDR, EPHY_SCR);
+        ui16Val |= (EPHY_SCR_INTEN_EXT | EPHY_SCR_INTOE_EXT);
+        MAP_EMACPHYWrite(EMAC0_BASE, PHY_PHYS_ADDR, EPHY_SCR, ui16Val);
+        MAP_EMACPHYWrite(EMAC0_BASE, PHY_PHYS_ADDR, EPHY_MISR1, (EPHY_MISR1_LINKSTATEN |
+                     EPHY_MISR1_SPEEDEN | EPHY_MISR1_DUPLEXMEN | EPHY_MISR1_ANCEN));
+
+        // Read the PHY interrupt status to clear any stray events.
+        ui16Val = MAP_EMACPHYRead(EMAC0_BASE, PHY_PHYS_ADDR, EPHY_MISR1);
+
+        // Set MAC filtering options.  We receive all broadcast and mui32ticast
+        // packets along with those addressed specifically for us.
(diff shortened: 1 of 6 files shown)


htibosch:
Thanks a lot for sharing this new driver! I will have a closer look later today.

In the past, when creating demo software for evaluation boards, every chip vendor would provide yet another PHY driver. That driver would work for one particular PHY and not for others.

A couple of years ago, I downloaded all PDF's about all major 100 Mbps PHY's, and I looked for what they have in common, and what is different. As you know some registers are obligatory, and other registers are part-specific.

The result of this can be found in [phyHandling.c](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/portable/NetworkInterface/Common/phyHandling.c). It will poll all 32 ports, form 0 up to 31. I tries to read a PHY ID on every port. Some PHY's are actually network switches showing their ID on 4 or 5 PHY ports. These will all be reset and initialised.

Here is a list of PHY's that work well with this driver:
~~~c
/* ID's of supported PHY's : */
    #define PHY_ID_LAN8742A        0x0007c130
    #define PHY_ID_LAN8720         0x0007c0f0

    #define PHY_ID_KSZ8041         0x000010A1
    #define PHY_ID_KSZ8051         0x000010A1
    #define PHY_ID_KSZ8081         0x000010A1

    #define PHY_ID_KSZ8863         0x00221430
    #define PHY_ID_KSZ8081MNXIA    0x00221560

    #define PHY_ID_DP83848I        0x20005C90
    #define PHY_ID_DP83TC811S      0x2000A250

    #define PHY_ID_MV88E6071       0xFF000710
~~~

and the latest will be [KSZ8795](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/425).

Many network interfaces make use of phyHandling.c, for instance the [STM32Fxx driver](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/portable/NetworkInterface/STM32Fxx/NetworkInterface.c).

Initialisation is this simple:
~~~c
static EthernetPhy_t xPhyObject;
void vMACBProbePhy( void )
{
    vPhyInitialise( &xPhyObject, xSTM32_PhyRead, xSTM32_PhyWrite );
    xPhyDiscover( &xPhyObject );
    xPhyConfigure( &xPhyObject, &xPHYProperties );
}
~~~

As you see, it needs two access functions: `xSTM32_PhyRead()`, and `xSTM32_PhyWrite()`.

After this, the object will tell you the actual properties of the connection, and on which ports there is a Link Status.

Would it be possible to integrate this PHY-driver into you Network Interface? I will help where necessary.

Thanks,

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,599 @@
...
+
+        // Enable all processor interrupts.
+        MAP_IntMasterEnable();
+
+        // Tell the PHY to start an auto-negotiation cycle.
+        MAP_EMACPHYWrite(EMAC0_BASE, PHY_PHYS_ADDR, EPHY_BMCR, (EPHY_BMCR_ANEN |
+                EPHY_BMCR_RESTARTAN));
+    }
+
+    return success;
```
htibosch, line 230: So here, the function will return pdPASS when the initialisation went well, **and** when a Link Status is high:
~~~
success = ( xPhyObject.ulLinkStatusMask != 0U ) ? pdPASS : pdFAIL.
~~~
When the initialisation has failed, the driver will never return pdPASS.

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,599 @@
...
+}
+
+BaseType_t xNetworkInterfaceOutput(NetworkBufferDescriptor_t * const pxNetworkBuffer, BaseType_t xReleaseAfterSend)
+{
```
htibosch, line 234: The parameter `xReleaseAfterSend` will be non-zero in case the ownership of the network packet is passed to the driver.
This is a zero-copy driver, and so `xReleaseAfterSend` sill always be true. So it is OK not to check it. OR at most you could:
~~~c
    /* Check if `ipconfigZERO_COPY_TX_DRIVER` was set in FreeRTOSIPConfig.h. */
    configASSERT( xReleaseAfterSend != 0 );
~~~

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,599 @@
...
+        _tx_descriptor_list.write++;
+
+        // Wrap around if required
+        if (_tx_descriptor_list.write == SYSCONFIG_FREERTOS_TCP_TX_DESC_COUNT)
+        {
+            _tx_descriptor_list.write = 0;
+        }
+
+        // Give the DMA descriptor to the DMA controller
+        dma_descriptor->ui32CtrlStatus |= DES0_TX_CTRL_OWN;
```
htibosch, line 264: Did this work properly in your hardware? Often it is necessary to create a memory barrier, e.g. by reading back the descriptor.
Is D-caching disabled for the DMA-descriptor memory?

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,599 @@
...
+        dma_descriptor = &_tx_descriptors[_tx_descriptor_list.read];
+
+        // If the descriptor is still owned by the DMA controller, exit
+        if (dma_descriptor->ui32CtrlStatus & DES0_TX_CTRL_OWN)
+        {
+            break;
+        }
+
+        // Get the 'hidden' reference to the stack descriptor from the buffer
+        stack_descriptor = *((NetworkBufferDescriptor_t **) (((uint8_t *) dma_descriptor->pvBuffer1) - ipBUFFER_PADDING));
```
htibosch, line 437: A beautiful line of code, but we have a standard function that does the same, along with some security checks.
It is this function:
~~~c
    NetworkBufferDescriptor_t * pxPacketBuffer_to_NetworkBuffer( const void * pvBuffer );

    /* Find the Network packet that contains the buffer: */
    stack_descriptor = pxPacketBuffer_to_NetworkBuffer( dma_descriptor->pvBuffer1 );
~~~
You could even add a `configASSERT(stack_descriptor!=NULL)`, because there is something really wrong if this fails.

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,599 @@
...
+static void _process_received_packet(void)
+{
+    NetworkBufferDescriptor_t *new_stack_descriptor;
+    NetworkBufferDescriptor_t *cur_stack_descriptor;
+    tEMACDMADescriptor *dma_descriptor;
+    BaseType_t higher_priority_task_woken = pdFALSE;
+    uint32_t i;
+
+    // Go through the list of RX DMA descriptors
+    for (i = 0; i < SYSCONFIG_FREERTOS_TCP_RX_DESC_COUNT; i++)
```
htibosch, line 460: This ISR does a lot of processing, the code will call several `FromISR` functions. It looks correct to me, but personally I would find it more elegant to just wakeup the MAC-task `_packet_received_thread()`, and let that function work on the DMA queues.
[Here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/55b3b94eb8b4cfae320d20386e8c3f5b1dea13be/portable/NetworkInterface/STM32Fxx/NetworkInterface.c#L316) is an example of how a network ISR will delegate the actual processing to the EMAC thread. This is done for both reception, and for checking the transmission buffers.

It should not be complicated to move the code from the ISR to the deferred interrupt handler.
You will not need the queue `_received_packets_queue` anymore.
The function can block by calling :
~~~
ulTaskNotifyTake( pdFALSE, ulMaxBlockTime );
~~~
Remember that it should also [check the Link Status](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/55b3b94eb8b4cfae320d20386e8c3f5b1dea13be/portable/NetworkInterface/STM32Fxx/NetworkInterface.c#L1301) from time to time.
As long as the Link Status is low, the driver should not attempt to send packets, because that may block and fill all network buffers.

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,599 @@
...
+        dma_descriptor->ui32CtrlStatus |= DES0_TX_CTRL_OWN;
+
+        // Inform the DMA it has a new descriptor
+        MAP_EMACTxDMAPollDemand(EMAC0_BASE);
+
+        iptraceNETWORK_INTERFACE_TRANSMIT();
+    }
+    else
+    {
+        success = pdFALSE;
```
htibosch, line 273: At this point, sending of a packet has failed. Remember the release the network buffer:
~~~diff
     else
     {
         success = pdFALSE;
     }
     return success;
~~~
otherwise you'd get a leak.

htibosch (commented):
Hello @jscott-hotstart, I reviewed the network interface. I have a few remarks, some are more essential than others.
I am describing a certain model of a driver that employs a "deferred interrupt handler". This model has been used since the beginning of FreeRTOS+TCP and it resulted in safe high-performance network drivers that are easy to debug (at task level you can add printf's if you want).

A general remark about task priorities and FreeRTOS+TCP:
- Highest: the deferred interrupt handler of the EMAC (`niEMAC_HANDLER_TASK_PRIORITY`)
- Medium: the IP-task (`ipconfigIP_TASK_PRIORITY`)
- Lowest: tasks that make use of the TCP/IP stack

Other tasks are free to chose their task priority.

So you don't have to worry about performance when moving code from an ISR to an ordinary task.
If you have any questions or need help with the changes, please tell.

Thanks

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,599 @@
...
+ */
+static void _packet_received_thread(void *parameters);
+
+BaseType_t xNetworkInterfaceInitialise(void)
+{
+    uint8_t mac_address_bytes[6];
+    BaseType_t success;
+    uint16_t ui16Val;
+
+    _received_packets_queue = xQueueCreate(SYSCONFIG_FREERTOS_TCP_MAX_PENDING_RX_PACKETS, sizeof(NetworkBufferDescriptor_t*));
```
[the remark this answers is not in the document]
jscott-hotstart: Ok, I think I understand the intention here and will make the appropriate changes. Thanks!

portable/NetworkInterface/TM4C/NetworkInterface.c
jscott-hotstart: Got it. I'll add the assert.

portable/NetworkInterface/TM4C/NetworkInterface.c
jscott-hotstart: That would make things much easier! The porting guide for FreeRTOS+TCP does not mention this, and I'll be honest in that I hadn't looked at all the other implementations. I will make this change as well.

portable/NetworkInterface/TM4C/NetworkInterface.c
jscott-hotstart: This does work in hardware and is how TI implemented their reference driver for lwIP. I'll see if I can find references in the datasheet to confirm.

portable/NetworkInterface/TM4C/NetworkInterface.c
jscott-hotstart: I see what you're saying here. From the porting guide, it seemed to be implied that it was more advantageous to handle RX packets in the ISR. I'll take a look.

portable/NetworkInterface/TM4C/NetworkInterface.c
jscott-hotstart: Whoops! Thanks!

portable/NetworkInterface/TM4C/NetworkInterface.c
jscott-hotstart: I can find no information regarding a memory barrier being needed here, and the TI supplied lwIP driver does not use volatile keywords, etc. to enforce a barrier. The DMA apparently fetches the TX descriptors when a transmission is requested:

TXDescriptorTDES0

The RX descriptor documentation only mentions that this bit is cleared.

With that in mind, and test code working as expected, I feel comfortable using this bit as is.

portable/NetworkInterface/TM4C/NetworkInterface.c
jscott-hotstart: Ok, I think I have this handled.

jscott-hotstart:
@htibosch I've addressed your comments as of the latest commit (5fa21f4ed35ad457b8037059136dc5cced7a7c56). I also enabled and tested checksum offloading to the MAC. Thank you for your comments. I'll keep an eye out for more suggestions.

jscott-hotstart:
> 

My mistake; the correct commit should be 8a8717b9e09bdb485a1de5daef0628dcc5d3c583. I just removed a stale debug statement. Apologies.

portable/NetworkInterface/TM4C/NetworkInterface.c
htibosch: I see lots of changes, thanks for that!

Also you mentioned that : 

> From the porting guide, it seemed to be implied that it was more advantageous to handle RX packets in the ISR

Oops, I was not aware of that. Can you give me the URL of the page that says so?

And you are right about the conversion function `pxPacketBuffer_to_NetworkBuffer()`, it should be mentioned when talking about zero-copy method.

htibosch: I see, you have read that well. At the same time, the comment says that "The deferred interrupt handler is a standard RTOS task".

I will propose to change that text.

My feeling is that the effect of dropping packets during the ISR will won't have any noticeable effect on the performance. The CPU load will be the same, it's just that the packet will be dropped a fraction of a uS later.
Note that the unwanted packet kept the LAN busy for about 120 µS before it reached our device ( full-size packet on a 100 Mbps LAN ).

I will download your adapted sources and have a new look again.
Thanks a lot so far!

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,780 @@
...
+
+    // As this driver is strictly zero-copy, assert that the stack does not call this function with
+    // xReleaseAfterSend as false
+    configASSERT(0 != xReleaseAfterSend);
+
+    dma_descriptor = &_tx_descriptors[_tx_descriptor_list.write];
+
+    // If the DMA controller still owns the descriptor, all DMA descriptors are in use, bail out
+    if (0 == (dma_descriptor->ui32CtrlStatus & DES0_RX_CTRL_OWN))
+    {
```
htibosch, line 356: As `ui32CtrlStatus` looks like an unsigned value, the MISRA checker will like to see a comparison with `0U` in stead of just `0`.

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,780 @@
...
+        configASSERT(NULL != dma_descriptor->pvBuffer1);
+
+        // If the descriptor is still in use by DMA, stop processing here
+        if (DES0_RX_CTRL_OWN == (dma_descriptor->ui32CtrlStatus & DES0_RX_CTRL_OWN))
+        {
+            break;
+        }
+
+        // If there is NOT an error in the frame
+        if (0 == (dma_descriptor->ui32CtrlStatus & DES0_RX_STAT_ERR))
```
htibosch, line 604: Change `0` to `0U`?

htibosch (dismissed):
That looks very good, thank you.

One remark is serious: I recommend to add a new state `eMACInitError` which is reached when a fatal error has occurred.

The other remarks are minor and have to do with style and MISRA rules ( change `0` to `0U` when compared to an unsigned expression ). Although I do not know what to imagine about "unsigned zero" :-)

It is a pity that I don't have a TM4C board on my desk, otherwise I would like to test it.

Have you already run iperf3 or something like a flood-ping?

I approve this PR and want to thank you for your efforts and your patience.

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,780 @@
...
+    uint32_t write;
+    uint32_t read;
+} tDescriptorList;
+
+typedef enum
+{
+    eMACInitTask,      /* Task must be initialized. */
+    eMACInitHw,        /* Must initialize MAC. */
+    eMACInitComplete,  /* Initialization was successful. */
+} eMAC_INIT_STATUS_TYPE;
```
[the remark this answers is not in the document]
jscott-hotstart: > It makes no sense to call xTaskCreate() again and again.

I see what you mean here. I misunderstood your note re: multiple calls to `xNetworkInterfaceInitialise`, and maybe misunderstood the failure mode of FreeRTOS when it runs out of dynamic memory. I had assumed that at some point the dynamic memory pool _could_ have some memory freed to it and allow for the task to be created, but that's probably naive. I haven't had many opportunities to use an RTOS in my career.

I'll clean this up to always fail if any single initialization operation fails.

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,780 @@
...
+            break;
+        }
+
+        xMacInitStatus = eMACInitHw;
+
+    case eMACInitHw:
+        // Read the MAC from user Flash
+        if (pdPASS != _ethernet_mac_get(&mac_address_bytes[0]))
+        {
+            break;
```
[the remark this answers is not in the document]
jscott-hotstart: > I guess that a MAC address is also essential to operate?

That is correct given how the driver configures the MAC. The hardware can be configured to allow all Ethernet frames (i.e. promiscuous mode) but I haven't tested. The MAC is configured to filter non-multicast frames that are not addressed to the configured MAC address.

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,780 @@
...
+        }
+
+        MAP_SysCtlPeripheralReset(SYSCTL_PERIPH_EMAC0);
+
+        while (!MAP_SysCtlPeripheralReady(SYSCTL_PERIPH_EMAC0))
```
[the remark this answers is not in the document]
jscott-hotstart: > [...] I assume that it always return non-zero after a while?

It should unless something is damaged in the hardware or the user of the driver doesn't enable the clock to the MAC/PHY. Source code from vendor driver library is below (via macro redirect):

```c
bool
SysCtlPeripheralReady(uint32_t ui32Peripheral)
{
    //
    // Check the arguments.
    //
    ASSERT(_SysCtlPeripheralValid(ui32Peripheral));

    //
    // See if this peripheral is ready.
    //
    return(HWREGBITW(SYSCTL_PRBASE + ((ui32Peripheral & 0xff00) >> 8),
                     ui32Peripheral & 0xff));
}
```

Would you prefer I add a timeout?

portable/NetworkInterface/TM4C/NetworkInterface.c
```diff
@@ -0,0 +1,780 @@
...
+    case eMACInitTask:
+        // Create the RX packet forwarding task
+        if (pdFAIL == xTaskCreate(_deferred_task, "EMAC", configEMAC_TASK_STACK_SIZE, NULL, niEMAC_HANDLER_TASK_PRIORITY, &_deferred_task_handle))
+        {
+            break;
+        }
+
+        xMacInitStatus = eMACInitHw;
+
+    case eMACInitHw:
```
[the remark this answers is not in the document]
jscott-hotstart: I removed the case statement.

jscott-hotstart:
> I approve this PR and want to thank you for your efforts and your patience.

@htibosch  Thanks! And no problem at all. I appreciate the feedback. My apologies if I made this PR more difficult than it needed to be. I don't often contribute to GitHub projects, so this was a learning experience. I appreciate your help and patience as well!

portable/NetworkInterface/TM4C/NetworkInterface.c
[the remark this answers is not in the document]
htibosch: > The MAC is configured to filter non-multicast frames that are not addressed to the configured MAC address.

That sounds good.

In the library, multicast is currently only used by LLMNR and by [mDNS in the [IPv6/multi branch](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi).

> and maybe misunderstood the failure mode of FreeRTOS when it runs out of dynamic memory

To be more precise, once a system is up and running, a heap shortage can be possibly coped with. But when a basic task can not be created, a solution is needed.
It is also be possible to use [`xTaskCreateStatic()`](https://www.freertos.org/xTaskCreateStatic.html), which avoids the use of the heap.
When `configSUPPORT_STATIC_ALLOCATION` is defined, the IP-task will be [declared static](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/55b3b94eb8b4cfae320d20386e8c3f5b1dea13be/FreeRTOS_IP.c#L1333).

jscott-hotstart:
@htibosch Is there something else I need to do to get this finished? It looks like we're pending the pr-bar-raiser reviews.

htibosch (approved):
@jscott-hotstart wrote:

> @htibosch Is there something else I need to do to get this finished? It looks like we're pending the pr-bar-raiser reviews

I think so to, I had approved your PR already and have thanked you for it.

@AniruddhaKanhere, can you collect more approvals for this PR?

AniruddhaKanhere (approved):

paulbartell (approved):

PR #428, opened 2022-03-04
Return false in ZynqUS port when auto-negotiation fails

StefanBalt:
Phy_Setup_US() will return XST_FAILURE if the auto-negotiation fails
(e.g. no cable connected). XEmacPs_SetOperatingSpeed() must be called
with a valid ulLinkSpeed or it will run into an assertion.

Description
-----------

Test Steps
-----------

Related Issue
-----------


htibosch:
Thank you @StefanBalt for this PR.
After applying this patch, have you tried to use the auto-speed option?
Does it work properly when later on you connect the device to a switch?

I will also try it.

htibosch:
@StefanBalt : in stead of responding to this PR, I just created a new PR to solve the same problem. I mistakenly thought that you had submitted an "issue" in stead of a "pull request". Sorry about that.

It is PR #429 "Check the result of Phy_Setup_US before continuing". I don't mind to withdraw that PR, as long as you take notice of the proposed changes :-)

Also I would be grateful if you can test the proposed code. For me, the testing went as expected.

Many network interfaces use a `enum` variable that describes the steps to be taken.

Thanks

StefanBalt:
My PR is just a quick and dirty fix. I prefer having a improved structure as proposed in your PR.

PR #429, opened 2022-03-06
Check the result of Phy_Setup_US before continuing

htibosch:
Description
-----------
This PR changes the NetworkInterface.c for Xilinx UltraScale.

The initialisation is now divided in several stages:
~~~c
typedef enum xEMAC_STATE
{
    xEMAC_Init,      /* Initialise the EMAC. */
    xEMAC_SetupPHY,  /* Detect and initialise the PHY. May be called repeatedly. */
    xEMAC_WaitPHY,   /* Wait for the PHY to report a Link Status and a speed. */
    xEMAC_Ready,     /* The EMAC is initialised and the Link Status is high. */
    xEMAC_Fatal      /* A fatal error has occurred. */
} EMACState_t;
~~~

When the LAN is connected, it will only be called once and pass all states up to `xEMAC_Ready`.
When the LAN is disconnected, it will stay in `xEMAC_SetupPHY` until the speed is known and the Link Status goes high.

Test Steps
-----------
Start the DUT with a LAN, and then start it again without a LAN connection.
Also, you can temporarily disconnect the LAN and reconnect it.

Related Issue
-----------
Without this patch, the function `XEmacPs_SetOperatingSpeed()` would assert because the speed ( of 1 ) is invalid.

### DIFF
--- a/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
+++ b/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
@@ -184,7 +184,7 @@ BaseType_t xNetworkInterfaceInitialise( void )
 {
     uint32_t ulLinkSpeed, ulDMAReg;
     BaseType_t xStatus, xReturn = pdFAIL;
-    XEmacPs * pxEMAC_PS;
+    XEmacPs * pxEMAC_PS = &( xEMACpsif.emacps );
     const TickType_t xWaitLinkDelay = pdMS_TO_TICKS( 1000U );
 
     switch( eEMACState )
@@ -192,7 +192,6 @@ BaseType_t xNetworkInterfaceInitialise( void )
         case xEMAC_Init:
 
             ulPHYLinkStatus = 0U;
-            pxEMAC_PS = &( xEMACpsif.emacps );
             memset( &xEMACpsif, '\0', sizeof( xEMACpsif ) );
 
             xStatus = XEmacPs_CfgInitialize( pxEMAC_PS, &mac_config, mac_config.BaseAddress );
--- a/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
+++ b/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
@@ -274,26 +274,30 @@ BaseType_t xNetworkInterfaceInitialise( void )
         case xEMAC_WaitPHY:
             prvGMACWaitLS( xWaitLinkDelay );
 
-            if( xGetPhyLinkStatus() != pdFALSE )
+            if( xGetPhyLinkStatus() == pdFALSE )
             {
+                /* The Link Status is not yet high, Stay in 'xEMAC_WaitPHY'. */
+                break;
+            }
+
+            if( xEMACTaskHandle == NULL )
+            {
+                /* The deferred interrupt handler task is created at the highest
+                 * possible priority to ensure the interrupt handler can return directly
+                 * to it.  The task's handle is stored in xEMACTaskHandle so interrupts can
+                 * notify the task when there is something to process. */
+                xTaskCreate( prvEMACHandlerTask, "EMAC", configEMAC_TASK_STACK_SIZE, NULL, niEMAC_HANDLER_TASK_PRIORITY, &xEMACTaskHandle );
+
                 if( xEMACTaskHandle == NULL )
                 {
-                    /* The deferred interrupt handler task is created at the highest
-                     * possible priority to ensure the interrupt handler can return directly
-                     * to it.  The task's handle is stored in xEMACTaskHandle so interrupts can
-                     * notify the task when there is something to process. */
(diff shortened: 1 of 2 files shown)


portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -165,87 +165,150 @@ static u32 ulPHYIndex;
...
+            eEMACState = xEMAC_SetupPHY;
 
-        /* Setting the operating speed of the MAC needs a delay. */
-        vTaskDelay( pdMS_TO_TICKS( 25UL ) );
+        /* Fall through. */
 
-        ulDMAReg = XEmacPs_ReadReg( pxEMAC_PS->Config.BaseAddress, XEMACPS_DMACR_OFFSET );
-        /* Enable 16-bytes AHB bursts */
-        ulDMAReg = ulDMAReg | XEMACPS_DMACR_INCR16_AHB_BURST;
+        case xEMAC_SetupPHY:
```
AniruddhaKanhere, line 241: If we are always going to fall through, what is the point of having this case instead of a comment?
Because all the breaks end with fatal error.

portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -165,87 +165,150 @@ static u32 ulPHYIndex;
...
-         * packets from the receiver packet buffer memory when no AHB resource is available. */
-        XEmacPs_WriteReg( pxEMAC_PS->Config.BaseAddress, XEMACPS_DMACR_OFFSET,
-                          ulDMAReg /*| XEMACPS_DMACR_DISC_WHEN_NO_AHB_MASK*/ );
+            if( ulLinkSpeed == XST_FAILURE )
+            {
+                /* The speed could not be determined yet.
+                 * This is not a fatal error.
+                 * xNetworkInterfaceInitialise() will be called again
+                 * by the IP-task.
+                 */
```
AniruddhaKanhere, line 250: Ah I see, this is the reason why :)

portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -165,87 +165,150 @@ static u32 ulPHYIndex;
...
+
+                /* Only return pdTRUE when the Link Status of the PHY is high, otherwise the
+                 * DHCP process and all other communication will fail. */
+
+                xReturn = pdPASS;
+            }
+
+            eEMACState = xEMAC_Ready;
+
+            break;
```
AniruddhaKanhere, line 297: I did not understand the purpose of this break. Why not fall through?

AniruddhaKanhere (commented):
Some questions on this PR.
Also, updating the kernel submodule pointer seems like an unwanted change.

Thanks,
Aniruddha

portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -165,87 +165,150 @@ static u32 ulPHYIndex;
...
-    if( xEMACTaskHandle == NULL )
+    switch( eEMACState )
     {
-        pxEMAC_PS = &( xEMACpsif.emacps );
-        memset( &xEMACpsif, '\0', sizeof( xEMACpsif ) );
+        case xEMAC_Init:
 
-        xStatus = XEmacPs_CfgInitialize( pxEMAC_PS, &mac_config, mac_config.BaseAddress );
+            ulPHYLinkStatus = 0U;
+            pxEMAC_PS = &( xEMACpsif.emacps );
```
StefanBalt, line 195: This pointer is only set in the case xEMAC_Init.
In the second xNetworkInterfaceInitialise() call it is used uninitialized in the xEMAC_SetupPHY case.

StefanBalt (changes requested):
Due to the uninitialized `pxEMAC_PS ` pointer I run into an data abort exception. When fixing it as suggested it works fine.

portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
htibosch: Yes sure, good catch!
I just submitted this change.
Thanks

StefanBalt: lookin good now.

portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
htibosch: Hello @AniruddhaKanhere , thank you for these comments. I made a few changes, and also I attach a file called [`xilinx_us_driver.zip`](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/8338995/xilinx_us_driver.zip)

It shows symbolically how initialisation takes place, leaving out the details.

Also I ran my Xilinx UltraScale testing project:
- A startup while LAN connected: goes to `xEMAC_Ready`
- A startup where LAN is connected after 30 seconds: stays in `xEMAC_SetupPHY`
- Disconnect/connect the LAN while running

All tests went as expected.

AniruddhaKanhere (approved):

n9wxu (approved):

PR #431, opened 2022-03-09
Flaten DNS files

alfred2g:
Flatten DNS files

Description
-----------
Flatten DNS files by removing the DNS/ directory and merging all files in the parent directory.
change the buildsystem to point to the new location of files

Test Steps
-----------
Test cases build and pass


FreeRTOS_DNS.c
```diff
@@ -47,11 +47,11 @@
...
-#include "DNS/DNS_Globals.h"
-#include "DNS/DNS_Cache.h"
-#include "DNS/DNS_Parser.h"
-#include "DNS/DNS_Networking.h"
-#include "DNS/DNS_Callback.h"
+#include "DNS_Globals.h"
+#include "DNS_Cache.h"
+#include "DNS_Parser.h"
+#include "DNS_Networking.h"
+#include "DNS_Callback.h"
```
htibosch, line 54: @alfred2g, good job, thank you.  It compiled perfectly and all DNS-related tests went well.

htibosch (dismissed):
I also tried the responsiveness to LLMNR and NBNS. So I looked up the DUT from a host with either LLMNR or NBNS.  And also I looked up my host by calling `FreeRTOS_gethostbyname()`.

One thing, I wondered about the difference between the new filenames for DNS and e.g. TCP:

    DNS_Cache.c
    DNS_Callback.c
    DNS_Networking.c
    DNS_Parser.c

    FreeRTOS_TCP_Reception.c
    FreeRTOS_TCP_State_Handling.c
    FreeRTOS_TCP_Transmission.c
    FreeRTOS_TCP_Utils.c

Shall we leave that like this?

Otherwise I approve this PR.

AniruddhaKanhere (approved):

paulbartell (approved):
I am not a fan of this change in directory structure because it makes it harder to separate out what code is for what module. If DNS was fully implemented in a single file, I would probably feel differently on this.

I am also not a fan of adding "FreeRTOS_" to the beginning of every file name because in an editor with many tabs open, the portion of the name which is most important is also most likely to be truncated.
It's no fun figuring out which "FreeRTOS_..." tab is the one you're actually looking for.

Given that this conforms to how the project is currently organized, I'll approve this PR.

htibosch:
@paulbartell wrote:

> I am also not a fan of adding "FreeRTOS_" to the beginning of every file name because in an editor with many tabs open, the portion of the name which is most important is also most likely to be truncated.

That's a very good reason, I hadn't thought of that yet.

PR #441, opened 2022-03-29
IPv6/multi: Updating driver STM32Fxx

htibosch:
Description
-----------

This PR contains an update for the STM32Fxx driver.

- Introduction of EMAC_DMA_BUFFER_SIZE: the maximum packet size for DMA
- The inclusion of "stm32fxx_hal_eth.h" comes earlier
- Changes for MDNS
- `ipFRAGMENT_OFFSET_BIT_MASK` does not need `FreeRTOS_ntohs()`
- Now also compatible with stm32F1xx

The introduction of `EMAC_DMA_BUFFER_SIZE` is the most important change: without that change it may happen that an incoming packet overwrites the buffer of the next packet.

### DIFF
--- a/portable/NetworkInterface/STM32Fxx/NetworkInterface.c
+++ b/portable/NetworkInterface/STM32Fxx/NetworkInterface.c
@@ -4,8 +4,8 @@
  */
 
 /*
- * FreeRTOS+TCP V2.3.1
- * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ * FreeRTOS+TCP V2.4.0
+ * Copyright (C) 2021 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
  *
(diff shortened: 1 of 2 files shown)


htibosch:
This PR is related to PR #440 which corrects MSS in case IPv6 is enabled. An IPv6 header is 20 bytes larger than an IPv4 header, and so MSS should be 20 bytes less.

AniruddhaKanhere (dismissed):

karkhaz:
Hi! The change to run-proofs.py is making the CI fail. We're working on it, ETA end of this week. @markrtuttle

karkhaz:
@htibosch can I suggest that you discard the change to `run-cbmc-proofs.py` if it's not needed? If you really need this change we can get it fixed up this week, but discarding your change to that file should allow the CI to pass. @AniruddhaKanhere can also help with CI changes like this. Cheers.

aggarg (approved):

PR #444, opened 2022-04-03
Suppress spurious warnings from the CCS compiler

jonathangjertsen:
Supress spurious warnings from the CCS compiler

Description
-----------
Fixes warnings when building with CCS:

`error #1916-D: definition at end of file not followed by a semicolon or a declarator`

Test Steps
-----------
Compile a project with this fix, and the warning should go away

Related Issue
-----------
N/A

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #190, opened 2021-02-07
Remove function defs from header files

AniruddhaKanhere:
Description
-----------
This PR removes function definitions in header files and puts them in proper source (.c) files.

Doing so required fixing CBMC proofs, unit tests, and build checks. Moreover, adding the function definitions to c files required the doxygen documentation to be updated.
As a result of the above changes, the total changed files have increased from 5->24.

Test Steps
-----------

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #445, opened 2022-04-05
Make conversions from uint8 to eIPTCPState_t explicit

jonathangjertsen:
Description
-----------
Prevents this warning from the TI CCS compiler:

    warning #190-D: enumerated type mixed with another type

I haven't seen this warning from other compilers, but it seems like a good idea regardless to signal that this conversion is OK and expected. Explicit casts are already being used for the opposite conversion.

This and #444 should be sufficient to compile the library without warnings with TI's compiler, with one exception which I will create an issue for because I don't know the library well enough to resolve it

Test Steps
-----------
None

Related Issue
-----------
N/A


AniruddhaKanhere (approved):

AniruddhaKanhere:
Thank you @jonathangjertsen for this PR, I have approved it and working on merging it after another review.

paulbartell (approved):

Issue #446, opened 2022-04-05
pxProtPack is set but unused in xCheckSizeFields


jonathangjertsen:
In FreeRTOS_IP.c, in `xCheckSizeFields`, the variable `pxProtPack` is set but not used. This generates a warning when compiling with the TI CCS compiler:

     line 2566: error #552-D: variable "pxProtPack" was set but never used

There is a `sizeof( pxProtPack->xUDPPacket.xUDPHeader )`, but `sizeof` just looks at the type of the expression. Would it be OK to replace it with the equivalent `sizeof( UDPHeader_t )`?

The code looks like it has been copy-pasted from `usGenerateProtocolChecksum`, so I don't know if the variable is intended to be used for something or not.

archigup:
Hi, thanks for the information; I have made PR #448 to fix this.

PR #448, opened 2022-04-07
Remove unusued variable

archigup:
Fixes https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/446

### DIFF
--- a/FreeRTOS_IP.c
+++ b/FreeRTOS_IP.c
@@ -2563,7 +2563,6 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
         size_t uxLength;
         const IPPacket_t * pxIPPacket;
         UBaseType_t uxIPHeaderLength;
-        const ProtocolPacket_t * pxProtPack;
         uint8_t ucProtocol;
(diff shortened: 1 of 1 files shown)


aggarg (approved):

FreeRTOS_IP.c
```diff
@@ -2625,10 +2624,6 @@ static eFrameProcessingResult_t prvProcessIPPacket( IPPacket_t * pxIPPacket,
              * protocol (Layer 3 or 4) header will be aligned, which is the convenience
```
jonathangjertsen, line 2625: This comment should be removed too, I think

htibosch: Yes sure, the entire comment can be removed now. The same comment can be found earlier in this source file.

htibosch (dismissed):
Thank you for this small but useful change, @archigup

FreeRTOS_IP.c
archigup: Removed the comment

aggarg (approved):

kstribrnAmzn (approved):

Issue #456, opened 2022-04-22
FreeRTOSIPConfigDefaults define error


holden-zenith:
This file is written as if ipconfigEVENT_QUEUE_LENGTH and ipconfigNETWORK_MTU aren't user defined then they will be set to a default. However, the checks 
```
#if ( ipconfigEVENT_QUEUE_LENGTH < ( ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS + 5 ) )
    #error The ipconfigEVENT_QUEUE_LENGTH parameter must be at least ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS + 5
#endif
```

```
#if ( ipconfigNETWORK_MTU < 46 )
    #error ipconfigNETWORK_MTU must be at least 46.
#endif
```
prevent the defaults from ever being used because user definitions are required to pass these checks before the defaults are set.

AniruddhaKanhere:
Hello @holden-zenith, thank you for bringing this to our attention. I have created PR https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/460 to resolve this issue.

AniruddhaKanhere:
I also created a similar PR in IPv6 branch https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/461.

PR #460, opened 2022-04-25
IPv4/single: Update FreeRTOSIPConfigDefaults.h

AniruddhaKanhere:
Description
-----------
As pointed out by @holden-zenith in https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/456, there is some reordering required in FreeRTOSIPConfigDefaults.h to allow the code to compile even in absence of user provided definitions.

This PR modifies the file to that extent.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/include/FreeRTOSIPConfigDefaults.h
+++ b/include/FreeRTOSIPConfigDefaults.h
@@ -1,6 +1,6 @@
 /*
- * FreeRTOS+TCP V2.3.3
- * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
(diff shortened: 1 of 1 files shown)


htibosch (approved):
It looks all good to me.
I never saw compilation errors because I had both macros defined in my `FreeRTOSIPConfig.h`.
Maybe we should also test-compile some project with an empty `FreeRTOSIPConfig.h`?
Thank you [Aniruddha](https://github.com/AniruddhaKanhere) for this PR, I also approve it.

aggarg (approved):

PR #462, opened 2022-04-26
TM4C: Enable PHY interrupts for link changes

martinaao:
Enable PHY interrupts for link changesEnable PHY interrupts for link changes

Description
-----------
The EMAC must be reconfigured after PHY link changes. In the current NetworkInterface.c code this is done in the PHY interrupt handler, however the interrupts for the internal PHY are not enabled, so the MAC is never reconfigured after initial link-up.

This results in unstable startup of the EMAC. 

See the documentation in Tivaware for EMACPHYConfigSet.

@jscott-hotstart Thanks a lot for adding support for this board

Test Steps
-----------
This issue is hard to reproduce since it triggers undefined behaviour in the EMAC. It seems to be very timing sensitive. When failing the symptomes are missing interrupts from the EMAC.

Related Issue
-----------
N/A

### DIFF
--- a/portable/NetworkInterface/TM4C/NetworkInterface.c
+++ b/portable/NetworkInterface/TM4C/NetworkInterface.c
@@ -278,8 +278,18 @@ BaseType_t xNetworkInterfaceInitialise( void )
                     0 );
 
 
-                /* Clear any stray PHY interrupts that may be set. */
+                /* Clear any stray MISR1 PHY interrupts that may be set. */
                 ui16Val = MAP_EMACPHYRead( EMAC0_BASE, PHY_PHYS_ADDR, EPHY_MISR1 );
+                /* Enable link status change interrupts */
+                ui16Val |= 
+                    (   EPHY_MISR1_LINKSTATEN |
+                        EPHY_MISR1_SPEEDEN |
+                        EPHY_MISR1_DUPLEXMEN |
+                        EPHY_MISR1_ANCEN 
+                    );
+                MAP_EMACPHYWrite( EMAC0_BASE, PHY_PHYS_ADDR, EPHY_MISR1, ui16Val );
+
+                /* Clear any stray MISR2 PHY interrupts that may be set. */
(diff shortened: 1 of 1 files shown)


htibosch:
Thank you @martinaao for this PR.

Unfortunately, I still don't have a TM4C evaluation board to test the changes.

Were you able to test the change in a real application?

I understand that your allows interrupts triggered by a Link Status change. What happens if the LS goes down? Is there a speed and duplex when a PHY is not connected?

martinaao:
> Were you able to test the change in a real application?

Yes, i have tested on a real world application but only on the Eval Board EK-TM4C1294XL for now. In our application every 4-5 builds (with code changes) would fail until i made this change.

> 
> I understand that your allows interrupts triggered by a Link Status change. What happens if the LS goes down? Is there a speed and duplex when a PHY is not connected?
I have not seen problems after taking link down and up again, when the LS goes down the MAC is not re-configured. 

According to the documentation from TI the PHY must be configured first, then the EMAC since the MAC parameters are latched by the HW after the peripheral is reset. Since the configuration of the PHY can be reset in phyHandling.c we must wait for the speed change interrupt (that always comes after reset) before re-configuring the EMAC.

htibosch (approved):
Unfortunately I do not have the hardware, but I trust on your reports about testing in real hardware.
I approve this PR.
Thanks a lot, Hein

paulbartell (approved):

cobusve (approved):

AniruddhaKanhere:
@martinaao apologies but can you please rebase your branch?
The `portable/NetworkInterface/TM4C/NetworkInterface.c` file is now moved one level deep in source folder like so 'source/portable/NetworkInterface/TM4C/NetworkInterface.c`

AniruddhaKanhere:
Another option is that I can create a PR on your behalf. Would you like me to do that?

AniruddhaKanhere:
I have rebased and recreated this PR here: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/478.

Closing this PR now. Thank you @martinaao.

PR #465, opened 2022-05-12
IPv4/single: STM32H7xx driver check result of HAL_ETH_Init

htibosch:
Description
-----------
This PR is in response to [an issue](https://forums.freertos.org/t/freertos-tcp-with-no-phy/14533/6) reported on the FreeRTOS forum.

The function `HAL_ETH_Init()` checks if the PHY hardware is OK. The result of the function was not checked, which may result in a crash.

[Dweb_2](https://forums.freertos.org/u/Dweb_2), thank you for reporting this.

Test Steps
-----------
It is difficult to reproduce without changing the hardware.


AniruddhaKanhere:
Opened a rebased version of this PR: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/479
Closing this now.

PR #467, opened 2022-05-14
Fix ACK handling at ESTABLISHED state

xuelix:
Description
This is to fix an issue discovered in TCP protocol testing.
When in ESTABLISHED state, if ACK bit is not set in the incoming segment packet, the packet should be dropped.  (RFC792 page69-page76)

Test Steps
-----------

Related Issue
-----------


htibosch (approved):
At this moment I am traveling, and I can not do a real-life test.
But the change is clear and is conformant with mentioned NFC.
I approve it. Thank you @xuelix

AniruddhaKanhere (dismissed):

PR #471, opened 2022-05-19
Merge changes to mainline

AniruddhaKanhere:
Description
-----------
This PR is to compare any and all changes different between `IntegrationTesting1` and `main` branch.

Test Steps
-----------

Related Issue
-----------


AniruddhaKanhere:
Thank you @htibosch for that information.
I am not sure why would that happen. Because most of the functions are intact and have just been moved around.
The only module which was almost completely re-written from scratch is DNS module.

Let me know what comes up?

htibosch:
I found the problem and solved it in PR #481. It had to do with the TCP timer `xTCPTimer`.

PR #476, opened 2022-05-30
Misra check fix and update

xuelix:
Description
-----------
Fixed MISRA violations.  Updated the MISRA.md

Test Steps
-----------

Related Issue
-----------


source/FreeRTOS_Sockets.c
```diff
@@ -4098,7 +4101,7 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
             size_t uxLittlePerc = sock20_PERCENT;
             size_t uxEnoughPerc = sock80_PERCENT;
             size_t uxSegmentCount = pxSocket->u.xTCP.uxRxStreamSize / pxSocket->u.xTCP.usMSS;
-            static const struct xPercTable
+            static const struct Percent
```
AniruddhaKanhere, line 4104: Why was this structure name changed?

We are not changing that below in line 4108?

source/FreeRTOS_TCP_Reception.c
```diff
@@ -194,10 +169,10 @@
  *         negative value wherein the calling function should not process this packet any
  *         further and drop it.
  */
-    static int32_t prvSingleStepTCPHeaderOptions( const uint8_t * const pucPtr,
-                                                  size_t uxTotalLength,
-                                                  FreeRTOS_Socket_t * const pxSocket,
-                                                  BaseType_t xHasSYNFlag )
+    int32_t prvSingleStepTCPHeaderOptions( const uint8_t * const pucPtr,
```
AniruddhaKanhere, line 172: Why did we remove the static qualifier? Are these not used in this file only?

I think we should keep the static qualifier and remove the declaration from the header.

source/FreeRTOS_TCP_Reception.c
```diff
@@ -369,9 +344,9 @@
  * @param[in] uxIndex: Index of options in the TCP packet options.
  * @param[in] pxSocket: Socket handling the TCP connection.
  */
-        _static void prvReadSackOption( const uint8_t * const pucPtr,
-                                        size_t uxIndex,
-                                        FreeRTOS_Socket_t * const pxSocket )
+        void prvReadSackOption( const uint8_t * const pucPtr,
```
AniruddhaKanhere, line 347: Why did we remove the static qualifier? I think this function is local to this file.

I think we should keep the static qualifier and remove the declaration from the header.

source/include/FreeRTOS_DNS_Callback.h
```diff
@@ -27,37 +27,46 @@
...
+    #ifdef __cplusplus
+        extern "C" {
+    #endif
 
 /* FreeRTOS includes. */
     #include "FreeRTOS.h"
 
 /* FreeRTOS+TCP includes. */
     #include "FreeRTOS_IP.h"
+    #include "FreeRTOSIPConfigDefaults.h"
```
AniruddhaKanhere, line 41: `FreeRTOS_IP.h` already includes the `FreeRTOSIPConfigDefaults.h` file, why do we need to include it separately?

source/include/FreeRTOS_DNS_Callback.h
```diff
@@ -27,37 +27,46 @@
...
 
     #include "FreeRTOS_DNS_Globals.h"
 
 /* Standard includes. */
     #include <stdint.h>
```
AniruddhaKanhere, line 46: I think we should move this to the top. Before FreeRTOS.h is included.

AniruddhaKanhere (commented):
Thank you for this PR @xuelix.

I have some comments/questions. Can you please take a look?

source/FreeRTOS_Sockets.c
xuelix: This is the datatype name, below is the variable.

source/FreeRTOS_TCP_Reception.c
xuelix: These two were added to the header file for running unittest.

source/FreeRTOS_TCP_Reception.c
xuelix: These two were added to the header file for running unittest.

source/include/FreeRTOS_DNS_Callback.h
xuelix: The file includes some default value, only needed if they are not defined in FreeRTOS.h

AniruddhaKanhere (dismissed):

AniruddhaKanhere (dismissed):

source/FreeRTOS_DNS_Parser.c
```diff
@@ -549,6 +549,8 @@
         uint32_t ulIPAddress = 0U;
         uint32_t ulReturnIPAddress = 0U;
         uint16_t usDataLength;
+        uint8_t * pucBuffer = &( pucByte[ 0 ] );
```
paulbartell, line 552:
```suggestion
        uint8_t * pucBuffer = pucByte);
```

source/FreeRTOS_IP.c
```diff
@@ -1435,8 +1435,8 @@ static eFrameProcessingResult_t prvAllowIPPacket( const IPPacket_t * const pxIPP
                  * broadcast address. */
                 eReturn = eReleaseBuffer;
             }
-            else if( memcmp( ( void * ) &xBroadcastMACAddress,
-                             ( void * ) &( pxIPPacket->xEthernetHeader.xSourceAddress ),
+            else if( memcmp( ( const void * ) ( xBroadcastMACAddress.ucBytes ),
```
paulbartell, line 1438: Are the "( const void * )" casts still necessary?

source/FreeRTOS_IP_Utils.c
```diff
@@ -291,8 +289,12 @@ NetworkBufferDescriptor_t * pxUDPPayloadBuffer_to_NetworkBuffer( const void * pv
 BaseType_t xIsCallingFromIPTask( void )
 {
     BaseType_t xReturn;
+    TaskHandle_t xCurrentHandle, xCurrentIPTaskHandle;
+
+    xCurrentHandle = xTaskGetCurrentTaskHandle();
+    xCurrentIPTaskHandle = FreeRTOS_GetIPTaskHandle();
 
-    if( xTaskGetCurrentTaskHandle() == FreeRTOS_GetIPTaskHandle() )
+    if( xCurrentHandle == xCurrentIPTaskHandle )
```
paulbartell, line 297: Is this to fix a particular MISRA violation?

paulbartell (dismissed):

paulbartell:
Added a few minor suggestions.

source/FreeRTOS_IP_Utils.c
xuelix: Yes, MISRA check does not like the expression inside evaluation.  It might have unexpected result if the execution order might change the result.

source/FreeRTOS_IP.c
xuelix: This is to fix one MISRA complain.  memcmp has the first parameter defined as  const void*

AniruddhaKanhere (approved):

source/FreeRTOS_IP_Utils.c
paulbartell: Can you clarify the Rule #?

paulbartell: I'm guessing that xTaskGetCurrentTaskHandle performs multiple volatile access operations which MISRA considers a potentially persistent side-effect. Seems very unlikely to be a problem though.

paulbartell (approved):

htibosch:
Thank you @xuelix for this PR. I inspected all changes and I wouldn't have done it differently, all good.
Like @AniruddhaKanhere, this change confused me:
~~~diff
static const struct xPercTable
static const struct Percent
~~~
until you reminded us that `xPercTable` is only a type name, which conflicted with `xPercTable` as an object,

PR #481, opened 2022-06-10
Let the TCP timer becomes expired in stead of active

htibosch:
Description
-----------
The timer `xTCPTimer` is a timer that tells when the TCP engine must become active by calling `xTCPTimerCheck()`. In some occasions, the timer is forced to expire, for instance when `recv()` was called and created enough space in the reception stream: a WIN update shall be sent immediately.
However, after a certain commit to the branch `IntegrationTesting1`, the TCP timer was set as **enabled**, instead of **expired**.

This PR makes the following change:
~~~diff
-void vIPSetTCPTimerEnableState( BaseType_t xEnableState );
+void vIPSetTCPTimerExpiredState( BaseType_t xEnableState );
~~~
and
~~~diff
+        xTCPTimer.bActive = pdTRUE_UNSIGNED;
         if( xEnableState != pdFALSE )
         {
-            xTCPTimer.bActive = pdTRUE_UNSIGNED;
+            xTCPTimer.bExpired = pdTRUE_UNSIGNED;
         }
         else
         {
-            xTCPTimer.bActive = pdFALSE_UNSIGNED;
+            xTCPTimer.bExpired = pdFALSE_UNSIGNED;
         }
~~~

Test Steps
-----------
Send a constant stream of TCP data and look at the performance. There will be periods of inactivity.

### DIFF
--- a/source/FreeRTOS_IP_Timers.c
+++ b/source/FreeRTOS_IP_Timers.c
@@ -382,18 +382,19 @@ static BaseType_t prvIPTimerCheck( IPTimer_t * pxTimer )
 /**
  * @brief Enable/disable the TCP timer.
  *
- * @param[in] xEnableState: pdTRUE - enable timer; pdFALSE - disable timer.
+ * @param[in] xExpiredState: pdTRUE - set as expired; pdFALSE - set as non-expired.
  */
-    void vIPSetTCPTimerExpiredState( BaseType_t xEnableState )
+    void vIPSetTCPTimerExpiredState( BaseType_t xExpiredState )
     {
         xTCPTimer.bActive = pdTRUE_UNSIGNED;
-        if( xEnableState != pdFALSE )
+
+        if( xExpiredState != pdFALSE )
         {
-			xTCPTimer.bExpired = pdTRUE_UNSIGNED;
+            xTCPTimer.bExpired = pdTRUE_UNSIGNED;
         }
         else
         {
-			xTCPTimer.bExpired = pdFALSE_UNSIGNED;
+            xTCPTimer.bExpired = pdFALSE_UNSIGNED;
         }
     }
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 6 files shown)


angelonakos:
CBMC_RETRY

source/FreeRTOS_IP_Timers.c
```diff
@@ -382,17 +382,19 @@ static BaseType_t prvIPTimerCheck( IPTimer_t * pxTimer )
...
  * @brief Enable/disable the TCP timer.
  *
- * @param[in] xEnableState: pdTRUE - enable timer; pdFALSE - disable timer.
+ * @param[in] xExpiredState: pdTRUE - set as expired; pdFALSE - set as non-expired.
  */
-    void vIPSetTCPTimerEnableState( BaseType_t xEnableState )
+    void vIPSetTCPTimerExpiredState( BaseType_t xExpiredState )
     {
-        if( xEnableState != pdFALSE )
+        xTCPTimer.bActive = pdTRUE_UNSIGNED;
```
AniruddhaKanhere, line 389: If the timer is just being marked as expired, why do we need to mark it as active?
Is there any place where we will mark it as inactive thus potentially requiring the timer to be set as active again?

AniruddhaKanhere (commented):
@htibosch thank you for finding the bug and the raising the PR to fix it.

I just have one question. Grateful if you can answer that.

Thanks

source/FreeRTOS_IP_Timers.c
htibosch: @AniruddhaKanhere wrote:
> If the timer is just being marked as expired, why do we need to mark it as active?

Because when a timer is inactive, it will be ignored by the function `prvIPTimerCheck()`:
~~~c
static BaseType_t prvIPTimerCheck( IPTimer_t * pxTimer )
{
    BaseType_t xReturn;

    if( pxTimer->bActive == pdFALSE_UNSIGNED )
    {
        /* The timer is not enabled. */
        xReturn = pdFALSE;
    }
    else
~~~

and so `prvIPTimerCheck()` will never see the expired flag :-(

> Is there any place where we will mark it as inactive thus potentially requiring the timer to be set as active again?

The flag `bActive` is set during every call to `vIPSetTCPTimerExpiredState()`, while nobody will ever clear it. It would be enough to set it once at start-up.

At the other hand, that would require a new function `vIPSetTCPTimerEnableState()` and an extra function-call at start-up, and one more unit test... without a measurable performance gain.

AniruddhaKanhere (approved):

paulbartell (approved):

PR #463, opened 2022-05-01
IPv4/Single: Update ARP table from gratuitous ARP packets

htibosch:
Description
-----------
When receiving a gratuitous ARP **request**, and if there is a cache entry for that protocol address, the cache entry should be updated with the MAC-address provided,

The "issue" was reported on the FreeRTOS forum as [FreeRTOS+TCP not updating ARP table with gratuitous ARP request with a changed MAC address
Libraries](https://forums.freertos.org/t/freertos-tcp-not-updating-arp-table-with-gratuitous-arp-request-with-a-changed-mac-address/14670).
Thank you [@nlangenb](https://forums.freertos.org/u/nlangenb) for this.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #482, opened 2022-06-13
IPv4/single: Update ARP cache from gratuitous ARP packets v2

htibosch:
Description
-----------
The changes proposed in this PR were presented earlier in PR #463. Now I have adapted it to the new source structure of the library that was introduced about 3 weeks ago.

When receiving a gratuitous ARP **request**, and if there is a cache entry for that protocol address, the cache entry should be updated with the MAC-address provided,

The "issue" was reported on the FreeRTOS forum as [FreeRTOS+TCP not updating ARP table with gratuitous ARP request with a changed MAC address
Libraries](https://forums.freertos.org/t/freertos-tcp-not-updating-arp-table-with-gratuitous-arp-request-with-a-changed-mac-address/14670).
Thank you [@nlangenb](https://forums.freertos.org/u/nlangenb) for this.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -2,4 +2,4 @@ blank_issues_enabled: false
 contact_links:
   - name: FreeRTOS Community Support Forum
     url: https://forums.freertos.org/
-    about: Please ask and answer questions about FreeRTOS here.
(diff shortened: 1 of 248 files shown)


shubnil:
Hi Hein, Do we still this need this change. If yes, it will be good if we can create a PR for dev/ipv6_integration branch.

tony-josi-aws:
Closing this PR as #974 takes care of the changes.

PR #483, opened 2022-06-13
IPv4/single: resubmit ARP cache changes found in PR #366

htibosch:
Description
-----------
In December 2021, after protocol testing, @AniruddhaKanhere submitted some changes regarding the usage of the ARP cache, which he documented as follows:

"This PR removes the call to refresh ARP cache entry when a UDP/TCP packet is received. It replaces that call with a function to only update the ARP entry age.
Additionally, the translation of IP->MAC must be done through ARP cache. Thus this PR translates an IP to MAC using ARP cache instead of switching the source and destination MAC address before sending any packet"

In today's main branch, some of the changes have disappeared. I assume that this happened accidentally.


AniruddhaKanhere:
Thank you @htibosch for this PR. I am also not sure why the changes disappear.
Regardless, grateful if you could fix the unit-tests.

shubnil:
Hi Hein, Do we still this need this change. If yes, please help in fixing the unit-test and it will be good if we can create a PR for dev/ipv6_integration branch.

htibosch:
@shubnil wrote:
> Hi Hein, Do we still this need this change. If we do, please help in fixing the unit tests and it will be good if we can create a PR for dev/ipv6_integration branch.

Yes we still need it, both in IPv4 and dev/IPv6_integration.

sirnish (approved):

shubnil (approved):

PR #488, opened 2022-06-20
Fix misra Rule 8.4

alfred2g:
Fix misra Rule 8.4

Description
-----------
Change the build system to add static a satisfy mira rule 8.4

Test Steps
-----------
Rule violation disappeared from the code checker

### DIFF
--- a/test/Coverity/ConfigFiles/FreeRTOSIPConfig.h
+++ b/test/Coverity/ConfigFiles/FreeRTOSIPConfig.h
@@ -30,7 +30,7 @@
 #ifndef FREERTOS_IP_CONFIG_H
 #define FREERTOS_IP_CONFIG_H
 
-#define _static
+#define _static static
 
 #define ipconfigUSE_ARP_REMOVE_ENTRY       1
 #define ipconfigUSE_ARP_REVERSED_LOOKUP    1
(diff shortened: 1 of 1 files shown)


test/Coverity/ConfigFiles/FreeRTOSIPConfig.h
```diff
@@ -30,7 +30,7 @@
 #ifndef FREERTOS_IP_CONFIG_H
 #define FREERTOS_IP_CONFIG_H
 
-#define _static
+#define _static static
```
AniruddhaKanhere, line 33: I think the correct way is to remove `_static` from files - and replace it with `static`.

We had added that to make CBMC work - but it is not needed anymore.

Can we please do that?

AniruddhaKanhere (commented):
Thanks @alfred2g for the PR, I have a suggestion which should fix this problem once and for all. Can you please take a look?

test/Coverity/ConfigFiles/FreeRTOSIPConfig.h
alfred2g: we use it for unit tests as well in some parts no ?

htibosch (approved):
Looks all good, thank you.

aggarg (approved):

PR #490, opened 2022-06-20
Misra: suppress rule 5.4

alfred2g:
Misra: suppress rule 5.4

Description
-----------
Misra: suppress rule 5.4: identifiers should be unique up to the 30th character
but changing that might cause some backward incompatibility

### DIFF
--- a/source/include/FreeRTOSIPConfigDefaults.h
+++ b/source/include/FreeRTOSIPConfigDefaults.h
@@ -803,7 +803,9 @@
  * as early as possible.
  */
 
-/* suppressing misra rule 5.4 as changing the value
+/* in C90 First 31 characters should be different; in C99 first 63 characters
+ * should be significant
+ * suppressing misra rule 5.4 as changing the value
  * could lead to backward incompatibility */
 /* coverity[misra_c_2012_rule_5_4_violation] */
 #ifndef ipconfigETHERNET_DRIVER_FILTERS_FRAME_TYPES
@@ -817,6 +819,8 @@
  * That function checks if there is a UDP socket listening to a
  * given port number.
  */
+/* in C90 First 31 characters should be different; in C99 first 63 characters
+ * should be significant */
 /* coverity[misra_c_2012_rule_5_4_violation] */
 #ifndef ipconfigETHERNET_DRIVER_FILTERS_PACKETS
     #define ipconfigETHERNET_DRIVER_FILTERS_PACKETS    ( 0 )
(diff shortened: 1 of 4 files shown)


htibosch (approved):
Thank you, approved.

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -800,6 +800,7 @@
  * It should be noted that it is most efficient to drop unwanted packets
  * as early as possible.
  */
+/* coverity[misra_c_2012_rule_5_4_violation] */
```
AniruddhaKanhere, line 803: Can we add some comment about what this rule is and why are suppressing it?
(Maybe we can copy the PR description).

Doing it, in my opinion, should make the comment more understandable to the reader - otherwise, one unaware about MISRA/coverity might get confused by this.

AniruddhaKanhere (commented):
I think we should add comments saying that why was this rule suppressed and not fixed. To a new reader, it would seem like a fixable violation - but it is not as it will break backward compatibility.

source/include/FreeRTOSIPConfigDefaults.h
alfred2g: i don't think it is legal to state what the rule is

alfred2g: will add a little description on top

paulbartell (changes requested):
Rule 5.4 suppression needs to be more clearly explained. Something like: 

ISO C90 5.2.4.1 specifies the minimum number of character in an identifier which _must_ be considered significant by a compiler. Care must be taken the ensure that the C preprocessor used to compile FreeRTOS+TCP recognizes at least ?64? characters of an identifier as significant. 

It's probably best to handle this deviation at the project level.

source/include/FreeRTOSIPConfigDefaults.h
paulbartell: The problem is that the project docs are not up front about this non-compliance with the C90 / C89 standard. I have yet to run into a compiler that actually only checks the first 31 characters though.

In this case, all macros starting with ipconfigETHERNET_DRIVER_FILTER could be considered the same by a compiler.

Or we could just claim C99 compliance instead of C89/C90.

shubnil (approved):

shubnil (approved):

paulbartell (approved):

PR #492, opened 2022-06-22
Mirsa Rule 10.3 violations

alfred2g:
Mirsa Rule 10.3 violations

Description
-----------
fix Mirsa Rule 10.3 violations

Test Steps
-----------
Violations disappeared from the report

### DIFF
--- a/source/FreeRTOS_DNS.c
+++ b/source/FreeRTOS_DNS.c
@@ -426,12 +426,15 @@
               pxReceiveBuffer->pucPayloadBuffer );
 
         /* See if the identifiers match. */
+        printf("ux identifier is %d\n", uxIdentifier );
         if( uxIdentifier == ( TickType_t ) pxDNSMessageHeader->usIdentifier )
         {
+            printf("case 1\n");
             xExpected = pdTRUE;
         }
         else
         {
+            printf("case 2\n");
             xExpected = pdFALSE;
         }
 
--- a/source/FreeRTOS_DNS.c
+++ b/source/FreeRTOS_DNS.c
@@ -453,9 +453,9 @@
  * @returns whether sending the data was successful
  */
     static uint32_t prvSendBuffer( const char * pcHostName,
-                                     TickType_t uxIdentifier,
-                                     Socket_t xDNSSocket,
-                                     struct freertos_sockaddr * pxAddress )
+                                   TickType_t uxIdentifier,
+                                   Socket_t xDNSSocket,
+                                   struct freertos_sockaddr * pxAddress )
     {
         uint32_t xReturn = pdFAIL;
         struct xDNSBuffer xDNSBuf = { 0 };
--- a/source/FreeRTOS_DNS.c
+++ b/source/FreeRTOS_DNS.c
@@ -273,7 +273,6 @@
                 xHasRandom = xApplicationGetRandomNumber( &( ulNumber ) );
                 /* DNS identifiers are 16-bit. */
                 uxIdentifier = ( TickType_t ) ( ulNumber & 0xffffU );
-                printf("setting identifier\n", uxIdentifier );
             }
 
             #if ( ipconfigDNS_USE_CALLBACKS == 1 )
@@ -304,7 +303,6 @@
 
             if( ( ulIPAddress == 0U ) && ( xHasRandom != pdFALSE ) )
             {
-                printf("identifier vgethostbyname %d\n",  uxIdentifier );
                 ulIPAddress = prvGetHostByName( pcHostName,
                                                 uxIdentifier,
                                                 uxReadTimeOut_ticks );
@@ -428,17 +426,13 @@
               pxReceiveBuffer->pucPayloadBuffer );
 
         /* See if the identifiers match. */
-        printf("identifier %d identifier 2 %d\n",
-                uxIdentifier, pxDNSMessageHeader->usIdentifier );
         if( uxIdentifier == ( TickType_t ) pxDNSMessageHeader->usIdentifier )
         {
             xExpected = pdTRUE;
-            printf("first case\n");
         }
         else
         {
             xExpected = pdFALSE;
-            printf("second case\n");
         }
 
         /* The reply was received.  Process it. */
--- a/source/FreeRTOS_DNS.c
+++ b/source/FreeRTOS_DNS.c
@@ -273,6 +273,7 @@
                 xHasRandom = xApplicationGetRandomNumber( &( ulNumber ) );
                 /* DNS identifiers are 16-bit. */
                 uxIdentifier = ( TickType_t ) ( ulNumber & 0xffffU );
+                printf("setting identifier\n", uxIdentifier );
             }
 
             #if ( ipconfigDNS_USE_CALLBACKS == 1 )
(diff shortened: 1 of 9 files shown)


htibosch (commented):
Hello @alfred2g, you propose the following type of change:
~~~diff
-uint32_t ulIPAddress = 0UL;
+uint32_t ulIPAddress = ( uint32_t ) 0UL;
~~~
If possible, do not to use the L suffix and make the following:
~~~diff
-uint32_t ulIPAddress = 0UL;
+uint32_t ulIPAddress = 0U;
~~~
The problem with UL is that "L" depends on the platform, it may mean 16 bits, 32 bits, 64 bits or more, thus making it useless for a multi-platform library.

All compilers that I use are happy with this assignment:
~~~c
uint32_t ulIPAddress = 0U;
~~~

Hein

source/FreeRTOS_DNS.c
```diff
@@ -207,7 +207,7 @@
         static uint32_t prvPrepareLookup( const char * pcHostName )
     #endif
     {
-        uint32_t ulIPAddress = 0UL;
+        uint32_t ulIPAddress = ( uint32_t ) 0UL;
```
htibosch, line 210: @alfred2g, as I commented in another PR about MISRA: please do not use the "L" suffix, just write:
~~~c
 uint32_t ulIPAddress = 0U;
~~~

source/FreeRTOS_DNS.c
```diff
@@ -368,7 +368,7 @@
          * still needs to be tested. */
         *ppxNetworkBuffer = pxGetNetworkBufferWithDescriptor( uxExpectedPayloadLength +
                                                               uxHeaderBytes,
-                                                              0UL );
+                                                              ( TickType_t ) 0UL );
```
htibosch, line 372: Can you please search the whole source code for "[0-9]UL" and remove the "L"?
I think that in this case the explicit cast to `TickType_t` will not be necessary.

source/FreeRTOS_DNS.c
```diff
@@ -419,14 +419,14 @@
...
-        uint32_t ulIPAddress = 0UL;
+        uint32_t ulIPAddress = ( uint32_t ) 0UL;
         BaseType_t xExpected;
         const DNSMessage_t * pxDNSMessageHeader =
             ( ( const DNSMessage_t * )
               pxReceiveBuffer->pucPayloadBuffer );
 
         /* See if the identifiers match. */
-        xExpected = ( uxIdentifier == ( TickType_t ) pxDNSMessageHeader->usIdentifier );
+        xExpected = ( BaseType_t ) ( uxIdentifier == ( TickType_t ) pxDNSMessageHeader->usIdentifier );
```
htibosch, line 429: My coverity doesn't like an expression of the type:
~~~c
    int same = ( a == b );
~~~
If can either be solved in two ways:
~~~c
    int same;
    if( a == b )
    {
        same = pdTRUE;
    }
    else
    {
        same = pdFALSE;
    }
~~~
And this one-liner also works :
~~~
    int same = ( a == b ) ? pdTRUE : pdFALSE;
~~~

source/FreeRTOS_DNS.c
```diff
@@ -452,12 +452,12 @@
...
-                                     TickType_t uxIdentifier,
-                                     Socket_t xDNSSocket,
-                                     struct freertos_sockaddr * pxAddress )
+    static uint32_t prvSendBuffer( const char * pcHostName,
+                                   TickType_t uxIdentifier,
+                                   Socket_t xDNSSocket,
+                                   struct freertos_sockaddr * pxAddress )
     {
-        BaseType_t xReturn = pdFAIL;
+        uint32_t xReturn = pdFAIL;
```
htibosch, line 460: When `xReturn` becomes unsigned, it normally changes to `uxReturn`.
Please also check other appearances of `uint32_t xReturn`.

source/FreeRTOS_DNS_Cache.c
```diff
@@ -158,16 +159,17 @@
                                          uint32_t ulTTL,
                                          BaseType_t xLookUp )
     {
-        BaseType_t x;
+        UBaseType_t x;
```
htibosch, line 162: And the name `x` shall become `ux` :-)

source/FreeRTOS_DNS_Cache.c
```diff
@@ -158,16 +159,17 @@
...
+        BaseType_t xResult;
         TickType_t xCurrentTickCount = xTaskGetTickCount();
         uint32_t ulCurrentTimeSeconds;
 
         configASSERT( ( pcName != NULL ) );
 
-        ulCurrentTimeSeconds = ( xCurrentTickCount / portTICK_PERIOD_MS ) / 1000UL;
-        x = prvFindEntryIndex( pcName );
+        ulCurrentTimeSeconds = ( xCurrentTickCount / portTICK_PERIOD_MS ) / ( uint32_t ) 1000UL;
+        xResult = prvFindEntryIndex( pcName, &x );
```
htibosch, line 170:
~~~diff
-( uint32_t ) 1000UL
+1000U
~~~

htibosch (commented):
@alfred2g, thank you again for this MISRA PR.
Please consider getting rid of all "UL" suffixes of constant values.
The "L" stands for LONG, but as you know, the size of long is not always 32 bits.

source/FreeRTOS_DNS_Cache.c
```diff
@@ -207,44 +209,47 @@
...
-        return x;
+        return xResult;
     }
 
 /**
  * @brief returns the index of the hostname entry in the dns cache.
  * @param[in] pcName find it in the cache
- * @returns index number or -1 if not found
+ * @param [out] xResult index number
+ * @returns res 0 if index in found else -1
```
AniruddhaKanhere, line 228: Why not make it consistent with all the other functions and return a pdTRUE/pdFALSE value instead of 0 and -1?

source/FreeRTOS_DNS_Networking.c
```diff
@@ -96,7 +96,7 @@
                               struct freertos_sockaddr * xAddress,
                               struct xDNSBuffer * pxDNSBuf )
     {
-        BaseType_t xReturn = pdFALSE;
+        uint32_t xReturn = pdFALSE;
```
AniruddhaKanhere, line 99: I think we should do it the other way around. Make the return type of the function as `BaseType_t`. This will make it more consistent with the library and a tiny bit optimized on smaller (16-bit) platforms where `BaseType_t` is 16-bits.

AniruddhaKanhere (commented):
Thank you for this PR @alfred2g, I have some comments. Can you please take a look?

source/FreeRTOS_DNS.c
```diff
@@ -512,18 +519,18 @@
 
         struct freertos_sockaddr xAddress;
         struct xDNSBuffer xReceiveBuffer = { 0 };
-        BaseType_t xReturn = pdFAIL;
+        BaseType_t uxReturn = pdFAIL;
```
paulbartell, line 522: BaseType_t is signed. UBaseType_t is unsigned.

source/FreeRTOS_DNS.c
```diff
@@ -512,18 +519,18 @@
...
 
         prvFillSockAddress( &xAddress, pcHostName );
 
         do
         {
-            xReturn = prvSendBuffer( pcHostName,
-                                     uxIdentifier,
-                                     xDNSSocket,
-                                     &xAddress );
+            uxReturn = prvSendBuffer( pcHostName,
```
paulbartell, line 528:
```suggestion
            xReturn = prvSendBuffer( pcHostName,
```

source/FreeRTOS_DNS.c
paulbartell, line 522:
```suggestion
        BaseType_t xReturn = pdFAIL;
```

paulbartell (approved):

PR #493, opened 2022-06-23
MISRA rule 11.3 suppression change.

xuelix:
Description
-----------
Added comments to suppress MISRA rule 11.3 violation.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_ARP.c
+++ b/source/FreeRTOS_ARP.c
@@ -329,7 +329,7 @@ static void vProcessARPPacketReply( ARPPacket_t * pxARPFrame,
     if( pxARPWaitingNetworkBuffer != NULL )
     {
         /* MISRA C-2012 Rule 11.3 warns about casting pointer type to a different data type.
-         * The casting is used to map the databuffer to a predefined data structure
+         * The casting is used to map the data buffer to a predefined data structure
          * so that the data can be easily accessed. The buffer length has been validated to be
          * at least as big as the size of the data structure to be casted to. */
         /* coverity[misra_c_2012_rule_11_3_violation] */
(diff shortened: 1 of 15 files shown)


source/FreeRTOS_ARP.c
```diff
@@ -328,6 +328,9 @@ static void vProcessARPPacketReply( ARPPacket_t * pxARPFrame,
 
     if( pxARPWaitingNetworkBuffer != NULL )
     {
+        /* MISRA C-2012 Rule 11.3 warns about casting pointer type to a different data type.
+         * To be able to access various predefined fields from a data buffer, this mapping is intentional. */
+        /* coverity[misra_c_2012_rule_11_3_violation] */
```
AniruddhaKanhere, line 333: The mapping is intentional, but what a reviewer might want to know that whether this will cause an out of bound access. I think we should reword it similar to the following:

```suggestion
        /* MISRA C-2012 Rule 11.3 warns about casting pointer type to a different data
         * type. To be able to access various predefined fields from a data buffer,
         * this mapping is intentional and has been validated to not exceed the buffer. */
        /* coverity[misra_c_2012_rule_11_3_violation] */
```

And change it in all the places...

AniruddhaKanhere (changes requested):
Thanks @xuelix for this PR. Can you take a look at the suggestions I have for the wording of the comment?
Thanks again.

source/FreeRTOS_ARP.c
xuelix: looks good, will change the wording.

source/FreeRTOS_TCP_Reception.c
```diff
@@ -510,6 +528,12 @@
                                uint32_t ulReceiveLength )
     {
         /* Map the ethernet buffer onto the ProtocolHeader_t struct for easy access to the fields. */
+
+        /* MISRA C-2012 Rule 11.3 warns about casting pointer type to a different data type.
```
paulbartell, line 532: Rule 11.3 does not concern the size of pucEthernetBuffer at all. Rather it concerns the alignment requirements oof the ProtocolHeaders_t type. ProtocolHeaders_t is a packed struct and packed structs are by definition aligned to 1 byte.

AniruddhaKanhere (approved):

xuelix:
Change was merged to another PR

PR #494, opened 2022-06-23
Fix various compiler warnings

holden-zenith:
This fixes various compiler warnings found when testing with stm32f7

### DIFF
--- a/source/include/FreeRTOSIPConfigDefaults.h
+++ b/source/include/FreeRTOSIPConfigDefaults.h
@@ -252,8 +252,7 @@
 #ifdef ipconfigHAS_DEBUG_PRINTF
     #if ( ipconfigHAS_DEBUG_PRINTF == 0 )
         #ifdef FreeRTOS_debug_printf
-            #warning Do not define FreeRTOS_debug_print if ipconfigHAS_DEBUG_PRINTF is set to 0
-            #undef FreeRTOS_debug_printf
+            #error Do not define FreeRTOS_debug_print if ipconfigHAS_DEBUG_PRINTF is set to 0
         #endif /* ifdef FreeRTOS_debug_printf */
     #endif /* ( ipconfigHAS_DEBUG_PRINTF == 0 ) */
 #endif /* ifdef ipconfigHAS_DEBUG_PRINTF */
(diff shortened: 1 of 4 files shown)


source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -252,13 +252,15 @@
 #ifdef ipconfigHAS_DEBUG_PRINTF
     #if ( ipconfigHAS_DEBUG_PRINTF == 0 )
         #ifdef FreeRTOS_debug_printf
-            #error Do not define FreeRTOS_debug_print if ipconfigHAS_DEBUG_PRINTF is set to 0
+            #warning Do not define FreeRTOS_debug_print if ipconfigHAS_DEBUG_PRINTF is set to 0
```
AniruddhaKanhere, line 255: The `#warning` option is not supported on windows simulator.

Since this file is not specific to STM, I think we should not make this change...

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -274,13 +276,15 @@
 #ifdef ipconfigHAS_PRINTF
     #if ( ipconfigHAS_PRINTF == 0 )
         #ifdef FreeRTOS_printf
-            #error Do not define FreeRTOS_print if ipconfigHAS_PRINTF is set to 0
+            #warning Do not define FreeRTOS_print if ipconfigHAS_PRINTF is set to 0
```
AniruddhaKanhere, line 279: Same here.

AniruddhaKanhere (changes requested):
Thank you @holden-zenith for this PR.
I have one suggested change, can you please take a look?

htibosch (approved):
Turn the warning into an error, I'm OK with that change.
Thank you

**EDIT**
Sorry I pressed on a link from my email client and I only saw part of your PR. I will read (and review) more.

AniruddhaKanhere (approved):
Thanks for this PR @holden-zenith. I am approving this.

We'll need one more review - I am asking for it. Once we have that, we can merge the PR!

paulbartell (approved):

PR #497, opened 2022-06-30
MISRA: Supress  rule 21.6

alfred2g:
MISRA: Supress misra rule 21.6

Description
-----------
Suppress MISRA rule 21.6 as the use or snprintf is for logging purposes only

### DIFF
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -1109,6 +1109,8 @@ const char * FreeRTOS_strerror_r( BaseType_t xErrnum,
 
         default:
             /* Using function "snprintf". */
+            /* exception used for logging purposes only */
+            /* coverity[misra_c_2012_rule_21_6_violation] */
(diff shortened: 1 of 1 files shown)


shubnil (approved):

source/FreeRTOS_IP_Utils.c
```diff
@@ -1113,6 +1113,8 @@ const char * FreeRTOS_strerror_r( BaseType_t xErrnum,
 
         default:
             /* Using function "snprintf". */
+            /* exception used for logging purposes only */
+            /* coverity[misra_c_2012_rule_21_6_violation] */
```
paulbartell, line 1117: Rule 21.6 is ridiculous. We should just use a global deviation for it and ensure that Rules 22.X are enabled. 

I think a good justification regarding printf usage is that "-Wformat -Werror=format-security" is used when compiling, which should prevent format string errors.

paulbartell (approved):

PR #506, opened 2022-07-01
MISRA: suppress rule 10.5

alfred2g:
MISRA: suppress rule 10.5

Description
-----------
MISRA: suppress rule 10.5, no valid path to convert integer into an enum

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -392,6 +392,7 @@ static void prvProcessIPEventsAndTimers( void )
                      * pointer size of the platform */
                     /* coverity[misra_c_2012_rule_11_6_violation] */
                     uxState = ( uintptr_t ) xReceivedEvent.pvData;
+                    /* coverity[misra_c_2012_rule_10_5_violation] */
                     eState = ( eDHCPState_t ) uxState;
 
                     /* Process DHCP messages for a given end-point. */
--- a/source/include/FreeRTOS_IP_Private.h
+++ b/source/include/FreeRTOS_IP_Private.h
@@ -610,7 +610,7 @@ BaseType_t xIPIsNetworkTaskReady( void );
         uint8_t ucRepCount;            /**< Send repeat count, for retransmissions
                                         * This counter is separate from the xmitCount in the
                                         * TCP win segments */
-        uint8_t ucTCPState;            /**< TCP state: see eTCP_STATE */
+        eIPTCPState_t ucTCPState;      /**< TCP state: see eTCP_STATE */
         struct xSOCKET * pxPeerSocket; /**< for server socket: child, for child socket: parent */
         #if ( ipconfigTCP_KEEP_ALIVE == 1 )
             uint8_t ucKeepRepCount;
--- a/source/include/FreeRTOS_IP_Private.h
+++ b/source/include/FreeRTOS_IP_Private.h
@@ -610,7 +610,7 @@ BaseType_t xIPIsNetworkTaskReady( void );
         uint8_t ucRepCount;            /**< Send repeat count, for retransmissions
                                         * This counter is separate from the xmitCount in the
                                         * TCP win segments */
-        eIPTCPState_t ucTCPState;      /**< TCP state: see eTCP_STATE */
+        eIPTCPState_t eTCPState;       /**< TCP state: see eTCP_STATE */
         struct xSOCKET * pxPeerSocket; /**< for server socket: child, for child socket: parent */
         #if ( ipconfigTCP_KEEP_ALIVE == 1 )
             uint8_t ucKeepRepCount;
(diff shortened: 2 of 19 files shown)


source/FreeRTOS_IP.c
```diff
@@ -387,6 +387,7 @@ static void prvProcessIPEventsAndTimers( void )
 
                     /* Cast in two steps to please MISRA. */
                     uxState = ( uintptr_t ) xReceivedEvent.pvData;
+                    /* coverity[misra_c_2012_rule_10_5_violation] */
```
AniruddhaKanhere, line 390: If we are suppressing this rule anyways, then why not remove line 389? It seems that the only purpose of that line was to "please MISRA"

alfred2g: this violates 2 misra rules
the first line removes one, and the comment another one

source/include/FreeRTOS_IP_Private.h
```diff
@@ -606,7 +606,7 @@ BaseType_t xIPIsNetworkTaskReady( void );
         uint8_t ucRepCount;            /**< Send repeat count, for retransmissions
                                         * This counter is separate from the xmitCount in the
                                         * TCP win segments */
-        uint8_t ucTCPState;            /**< TCP state: see eTCP_STATE */
+        eIPTCPState_t ucTCPState;      /**< TCP state: see eTCP_STATE */
```
htibosch, line 609: Initially I never used enum's as a struct member because their size is unpredictable, it may vary from 8 to 32-bytes.
The advantage of storing it as a char is that the we can control the alignment of the struct.

I know that MISRA doesn't like the conversions from `uint8_t` to an `enum` and back because theoretically they might lead to a loss of bits. We know that the conversions are safe by design ( there are only 12 TCP states ).

Only one remark: variables and fields of type `enum` start with the prefix `e`, so it should be `eIPTCPState_t eTCPState`.

source/FreeRTOS_IP.c
htibosch: @alfred2g wrote:

> the first line removes one, and the comment another one

Exactly, it was indeed a way to please MISRA:
~~~c
    uintptr_t uxState = ( uintptr_t ) xReceivedEvent.pvData;
    eState = ( eDHCPState_t ) uxState;
~~~
The `pvData` field already exists for many years, it was there in the +UDP library version 1.0 ten years ago. It is a void pointer used for anything.

htibosch (approved):
Thank you @alfred2g for this PR, I approve it so far.
Please check my 2 remarks.

source/include/FreeRTOS_IP_Private.h
alfred2g: I didn't change it initially for this reason, but I got a comment to change it.
What do you suggest in that case? revert back to a fixed size or keep it as it is? 
we could also put the states as defines.
alternatively we could use 
```
enum __attribute__ ((__packed__)) my_enum {...};
```
this will make the enum 1 byte
we already have an include "packed.h"  ready

htibosch: > I didn't change it initially for this reason, but I got a comment to change it. What do you suggest in that case? revert back to a fixed size or keep it as it is? we could also put the states as defines. alternatively we could use
> 
> ```
> enum __attribute__ ((__packed__)) my_enum {...};
> ```
> 
> this will make the enum 1 byte we already have an include "packed.h" ready, with a MISRA exception for it

Haha, I haven't seen tha packed attribute on an enum declaration. That sounds like an ideal solution.

AniruddhaKanhere: @htibosch wrote:
> Initially I never used enum's as a struct member because their size is unpredictable

Yes, I thought of that but could not come up with a good reason of not leaving it up to the compiler. And then you proposed:

> The advantage of storing it as a char is that the we can control the alignment of the struct

I did think of this as well. But, is there a very strong reason that we want these `structs` to be aligned?

shubnil (approved):

paulbartell (commented):
The exception on source/FreeRTOS_IP.c line 390 needs to be handled correctly.

shubnil (approved):

n9wxu (changes requested):

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_privates_utest.c
```diff
@@ -2189,29 +2189,29 @@ void test_prvTCPSendCheck_InvalidValues( void )
 
     for( int i = 0; i < sizeof( array ); i++ )
     {
-        xSocket.u.xTCP.ucTCPState = array[ i ];
+        xSocket.u.xTCP.eTCPState = array[ i ];
```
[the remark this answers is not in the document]
alfred2g: it is a UT, but i will fix it

n9wxu (approved):

PR #511, opened 2022-07-01
MISRA: fix rule 8.9 violations

alfred2g:
MISRA: fix rule 8.9 violations

Description
-----------
variables should be defined inside functions if used only in those functions

### DIFF
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -184,11 +184,11 @@ const MACAddress_t xBroadcastMACAddress = { { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff
 /** @brief Structure that stores the netmask, gateway address and DNS server addresses. */
 NetworkAddressingParameters_t xNetworkAddressing = { 0, 0, 0, 0, 0 };
 
-/* used for unit testing */
-/* coverity[misra_c_2012_rule_8_9_violation] */
-
 /** @brief Default values for the above struct in case DHCP
  * does not lead to a confirmed request. */
+
+/* used for unit testing */
+/* coverity[misra_c_2012_rule_8_9_violation] */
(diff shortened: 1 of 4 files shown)


source/portable/BufferManagement/BufferAllocation_2.c
```diff
@@ -84,7 +84,6 @@ static size_t uxMinimumFreeNetworkBuffers;
  * in this array.  The array is not accessed directly except during initialisation,
  * when the xFreeBuffersList is filled (as all the buffers are free when the system
  * is booted). */
```
shubnil, line 86: Please move the comment as well to line 98.

source/FreeRTOS_IP.c
```diff
@@ -186,6 +186,9 @@ NetworkAddressingParameters_t xNetworkAddressing = { 0, 0, 0, 0, 0 };
 
 /** @brief Default values for the above struct in case DHCP
  * does not lead to a confirmed request. */
+
+/* used for unit testing */
```
AniruddhaKanhere, line 190: One might ask - "How is this used for unit-testing?"
Can we maybe make the comment a bit more descriptive? I think we should take a page from FreeRTOS-Kernel about the comments.

AniruddhaKanhere (commented):
Small NIT pick.

AniruddhaKanhere (approved):

source/FreeRTOS_TCP_IP.c
```diff
@@ -64,6 +64,10 @@
 /* Just make sure the contents doesn't get compiled if TCP is not enabled. */
 #if ipconfigUSE_TCP == 1
 
+
+/* declared global for unit testing purposes */
+/* coverity[misra_c_2012_rule_8_9_violation] */
+
```
paulbartell, line 70: This is static, but becomes a global during unit testing right?

paulbartell (approved):

PR #512, opened 2022-07-02
Misra rule 11.3, 11.4 suppression and 4.6 fix

xuelix:
Description
-----------

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_DHCP.c
+++ b/source/FreeRTOS_DHCP.c
@@ -739,10 +739,7 @@
             /* Map a DHCP structure onto the received data. */
 
             /* MISRA C-2012 Rule 11.3 warns about casting pointer type to a different data type.
-             * The casting is used to map the data buffer to a predefined data structure
-             * so that the data can be easily accessed. The buffer length has been validated to be
-             * at least as big as the size of the data structure to be casted to. */
-            /* coverity[misra_c_2012_rule_11_3_violation] */
+             * The struct to be casted to is defined as a packed struct.  The cast won't cause misalignment. */   /* coverity[misra_c_2012_rule_11_3_violation] */
             pxDHCPMessage = ( ( DHCPMessage_IPv4_t * ) pucUDPPayload );
 
             /* Sanity check. */
@@ -1024,10 +1021,7 @@
             pucUDPPayloadBuffer = &( pxNetworkBuffer->pucEthernetBuffer[ ipUDP_PAYLOAD_OFFSET_IPv4 ] );
 
             /* MISRA C-2012 Rule 11.3 warns about casting pointer type to a different data type.
-             * The casting is used to map the data buffer to a predefined data structure
-             * so that the data can be easily accessed. The buffer length has been validated to be
-             * at least as big as the size of the data structure to be casted to. */
-            /* coverity[misra_c_2012_rule_11_3_violation] */
+             * The struct to be casted to is defined as a packed struct.  The cast won't cause misalignment. */   /* coverity[misra_c_2012_rule_11_3_violation] */
             pxDHCPMessage = ( ( DHCPMessage_IPv4_t * ) pucUDPPayloadBuffer );
 
             /* Most fields need to be zero. */
--- a/source/FreeRTOS_DHCP.c
+++ b/source/FreeRTOS_DHCP.c
@@ -616,8 +616,8 @@
         {
             xDHCPSocket = FreeRTOS_socket( FREERTOS_AF_INET, FREERTOS_SOCK_DGRAM, FREERTOS_IPPROTO_UDP );
 
-            /* MISRA Rule 11.4 warns about casting pointer to a different type.
-             * The casting here is to use pointer to pass error code.
+            /* MISRA Rule 11.4 warns about conversion between a pointer and an integer.
+             * The conversion here is to use pointer to pass error code.
              * The pointer will be checked against the error code value
              * before any further pointer action. */
             /* coverity[misra_c_2012_rule_11_4_violation] */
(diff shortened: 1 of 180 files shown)


shubnil (dismissed):

n9wxu (dismissed):

source/FreeRTOS_DHCP.c
```diff
@@ -616,6 +616,11 @@
         {
             xDHCPSocket = FreeRTOS_socket( FREERTOS_AF_INET, FREERTOS_SOCK_DGRAM, FREERTOS_IPPROTO_UDP );
 
+            /* MISRA Rule 11.4 warns about casting pointer to a different type.
```
paulbartell, line 619: Which cast is being referenced here?

xuelix: I'll change the wording here.  It's a conversion.

AniruddhaKanhere (approved):

htibosch (approved):
> `/* MISRA C-2012 Rule 11.3 warns about casting pointer type`
> ` * to a different data type. The struct to be casted to is`
> ` * defined as a packed struct.  The cast won't cause misalignment. */`
> `/* coverity[misra_c_2012_rule_11_3_violation] */`

Thank you for these additions, I approve it.

A small comment for those who wonder why we ignore MISRA C-2012 Rule 11.3:

The TCP/IP library receives and sends all sorts of packets: ARP, TCP/IP, UDP/IP, DHCP, DNS etc. These packets are stored in a buffer `pucEthernetBuffer` which is always aligned in the same manner: a four-byte alignment **plus two** bytes. The reason for this is that the Ethernet header is 14 (12+2) bytes long. So after the Ethernet header, the words and shorts are properly aligned.

It is convenient (and also efficient) to access a word as a word, as in `pxIPHeader->ulDestinationIPAddress`.

There is one alignment exception, which is in the ARP header: the sender protocol address, or `ucSenderProtocolAddress[4]`, is declared as an array of 4 bytes. It falls on a 2-byte alignment.

In some situations, in stead of copying shorts and words, byte access will be done like here in FreeRTOS_DNS_Parser.c :
~~~c
#define vSetField16( pxBase, xType, xField, usValue )                                                        \
    {                                                                                                        \
        ( ( uint8_t * ) ( pxBase ) )[ offsetof( xType, xField ) + 0 ] = ( uint8_t ) ( ( usValue ) >> 8 );    \
        ( ( uint8_t * ) ( pxBase ) )[ offsetof( xType, xField ) + 1 ] = ( uint8_t ) ( ( usValue ) & 0xffU ); \
    }
~~~
which works independent from the endianness.

In other situations, where the alignment is unknown, `memcpy()` is used to access data.

When structs are copied, `memcpy()` should be used, and not a struct assignment as in `xCopy = xOrg`. The latter may lead to an alignment exception. The compiler can be mistaken about the alignment of the structs.

I recommend to use the compiler flags `-fno-builtin-memcpy` and `-fno-builtin-memset`. In some situations, compilers make the wrong assumptions about alignment, when replacing a 4-byte `memcpy()` with a 32-byte assignment.

The use of **packed** structs is necessary, the alignment in the declarations must be guaranteed the same as in the packets on the wire.

The FreeRTOS+TCP library has been tested and used on a wide variety of platforms: little and big endian, and on both 32- and 64-bit platforms.

Hein

paulbartell:
> 



> > `/* MISRA C-2012 Rule 11.3 warns about casting pointer type`
> > ` * to a different data type. The struct to be casted to is`
> > ` * defined as a packed struct.  The cast won't cause misalignment. */`
> > `/* coverity[misra_c_2012_rule_11_3_violation] */`
> 
> Thank you for these additions, I approve it.
> 
> A small comment for those who wonder why we ignore MISRA C-2012 Rule 11.3:
> 
> The TCP/IP library receives and sends all sorts of packets: ARP, TCP/IP, UDP/IP, DHCP, DNS etc. These packets are stored in a buffer `pucEthernetBuffer` which is always aligned in the same manner: a four-byte alignment **plus two** bytes. The reason for this is that the Ethernet header is 14 (12+2) bytes long. So after the Ethernet header, the words and shorts are properly aligned.
> 
> It is convenient (and also efficient) to access a word as a word, as in `pxIPHeader->ulDestinationIPAddress`.
> 
> There is one alignment exception, which is in the ARP header: the sender protocol address, or `ucSenderProtocolAddress[4]`, is declared as an array of 4 bytes. It falls on a 2-byte alignment.
> 
> In some situations, in stead of copying shorts and words, byte access will be done like here in FreeRTOS_DNS_Parser.c :
> 
> ```c
> #define vSetField16( pxBase, xType, xField, usValue )                                                        \
>     {                                                                                                        \
>         ( ( uint8_t * ) ( pxBase ) )[ offsetof( xType, xField ) + 0 ] = ( uint8_t ) ( ( usValue ) >> 8 );    \
>         ( ( uint8_t * ) ( pxBase ) )[ offsetof( xType, xField ) + 1 ] = ( uint8_t ) ( ( usValue ) & 0xffU ); \
>     }
> ```
> 
> which works independent from the endianness.
> 
> In other situations, where the alignment is unknown, `memcpy()` is used to access data.
> 
> When structs are copied, `memcpy()` should be used, and not a struct assignment as in `xCopy = xOrg`. The latter may lead to an alignment exception. The compiler can be mistaken about the alignment of the structs.
> 
> I recommend to use the compiler flags `-fno-builtin-memcpy` and `-fno-builtin-memset`. In some situations, compilers make the wrong assumptions about alignment, when replacing a 4-byte `memcpy()` with a 32-byte assignment.
> 
> The use of **packed** structs is necessary, the alignment in the declarations must be guaranteed the same as in the packets on the wire.
> 
> The FreeRTOS+TCP library has been tested and used on a wide variety of platforms: little and big endian, and on both 32- and 64-bit platforms.
> 
> Hein

When a struct is declared as packed, that struct and it's members are treated as byte-aligned. This means that a compiler should not make any alignment assumptions when copying or assigning members of the struct.

paulbartell (approved):

shubnil (approved):

shubnil (approved):

PR #515, opened 2022-07-05
Misra: suppress misra 2.3 and 2.4

alfred2g:
Misra: suppress misra 2.3 and 2.4

Description
-----------
Misra: suppress misra 2.3 and 2.4
we define structures to be conforming to FreeRTOS such as
```
typedef struct struct_name
{
     int a;
} struct_name_2;
```
which allows the use to use either one, and leaves the other one as not defined.

### DIFF
--- a/tools/coverity_misra.config
+++ b/tools/coverity_misra.config
@@ -36,7 +36,7 @@
         },
         {
             deviation: "Rule 2.3",
-            reason: "The way we declare structures are conformant with the FreeRTOS kernel, which leaves somes types unsued"
+            reason: "The way we declare structures are conformant with the FreeRTOS kernel, which leaves somes types unused."
         },
         {
             deviation: "Rule 2.4",
--- a/tools/coverity_misra.config
+++ b/tools/coverity_misra.config
@@ -40,7 +40,7 @@
         },
         {
             deviation: "Rule 2.4",
-            reason: "The way we declare structures are conformant with the FreeRTOS kernel, which leaves some tags unused"
+            reason: "Similar to the FreeRTOS Kernel, structures are always declared with both a struct tag and typedef alias. Some of these structs are always referred to by their typedef alias and thus the corresponding tags are unused."
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


htibosch:
image

Hi @alfred2g, when I check out your branch [`alfred2g:misra_r_2_3`](https://github.com/alfred2g/FreeRTOS-Plus-TCP/tree/misra_r_2_3), I find hundreds of differences with [`FreeRTOS:main`](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP). Is one of the branches running behind the other?
To mention one example, in FreeRTOS_ARP.c, line 405:
~~~diff
-    IPHeader_t * pxIPHeader = &( pxIPPacket->xIPHeader );
+    const IPHeader_t * pxIPHeader = &( pxIPPacket->xIPHeader );
~~~
I am not sure if it will become `const`, or if it has been `const`.

shubnil (approved):

tools/coverity_misra.config
```diff
@@ -33,6 +33,14 @@
         {
             deviation: "Rule 5.4",
             reason: "In C90 First 31 characters should be different; in C99 first 63 characters should be significant"
+        },
+        {
+            deviation: "Rule 2.3",
+            reason: "The way we declare structures are conformant with the FreeRTOS kernel, which leaves somes types unsued"
```
paulbartell, line 39:
```suggestion
            reason: "The way we declare structures are conformant with the FreeRTOS kernel, which leaves somes types unused."
```

tools/coverity_misra.config
```diff
@@ -33,6 +33,14 @@
...
             deviation: "Rule 5.4",
             reason: "In C90 First 31 characters should be different; in C99 first 63 characters should be significant"
+        },
+        {
+            deviation: "Rule 2.3",
+            reason: "The way we declare structures are conformant with the FreeRTOS kernel, which leaves somes types unused."
+        },
+        {
+            deviation: "Rule 2.4",
+            reason: "The way we declare structures are conformant with the FreeRTOS kernel, which leaves some tags unused"
```
paulbartell, line 43:
```suggestion
            reason: "Similar to the FreeRTOS Kernel, structures are always declared with both a struct tag and typedef alias. Some of these structs are always referred to by their typedef alias and thus the corresponding tags are unused."
```

paulbartell (approved):

AniruddhaKanhere (approved):

PR #517, opened 2022-07-06
Fix race condition in stream buffers when priority(IP task)==priority(task using TCP/IP)

AniruddhaKanhere:
Description
-----------
When the priority of the TCP/IP stack and the priority of the task using TCP/IP stack are equal AND when the task in question is using both `send` and `recv` functions, then there is a possible race condition that exists in the stream buffers which lead to advertisement of 0 window.
This PR fixes that by adding a critical section around the stream buffer member update section.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,11 @@ jobs:
            cmake -S test/build-combination -B test/build-combination/build/ \
             -DTEST_CONFIGURATION=DISABLE_ALL
            make -C test/build-combination/build/
+      - name: Build checks (Default configuration)
+        run: |
+           cmake -S test/build-combination -B test/build-combination/build/ \
+            -DTEST_CONFIGURATION=DEFAULT_CONF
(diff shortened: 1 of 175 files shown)


aggarg (approved):

htibosch (approved):
@AniruddhaKanhere wrote:

> ... when the task in question is using both send and recv functions

A small correction: there was a race condition between the IP-task and the task that calls `recv()`.
The transmission stream works independently.

And maybe it is also worth telling that the problem was reported by [Shayan Mukhtar](https://forums.freertos.org/u/shayan_mukhtar) in [this post](https://forums.freertos.org/t/possible-race-condition-in-tcp-stack/15361) on the FreeRTOS forum.

To summarise the problem once more: the members `uxHead` and `uxFront` must be updated together, in a single atomic operation. Suspend/resume will be called from the IP-task only.

The solution was tested and I approve it.

Thank you

PR #527, opened 2022-07-27
Fixing MISRA violation 4.6, 10.4

xuelix:
Description
-----------

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -1151,7 +1151,7 @@ const char * FreeRTOS_strerror_r( BaseType_t xErrnum,
             /* MISRA Ref 21.6.1 [snprintf and logging] */
 /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-216 */
             /* coverity[misra_c_2012_rule_21_6_violation] */
-            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int ) xErrnum );
+            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int32_t ) xErrnum );
             pcName = NULL;
             break;
     }
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -1151,7 +1151,7 @@ const char * FreeRTOS_strerror_r( BaseType_t xErrnum,
             /* MISRA Ref 21.6.1 [snprintf and logging] */
 /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-216 */
             /* coverity[misra_c_2012_rule_21_6_violation] */
-            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int32_t ) xErrnum );
+            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", xErrnum );
             pcName = NULL;
             break;
     }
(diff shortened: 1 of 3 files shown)


test/Coverity/ConfigFiles/FreeRTOSConfig.h
```diff
@@ -46,7 +46,7 @@
 #define configUSE_TICK_HOOK                        1
 #define configUSE_DAEMON_TASK_STARTUP_HOOK         1
 #define configTICK_RATE_HZ                         ( 1000U )                        /* In this non-real time simulated environment the tick frequency has to be at least a multiple of the Win32 tick frequency, and therefore very slow. */
-#define configMINIMAL_STACK_SIZE                   ( 70 )                           /* In this simulated case, the stack only has to hold one small structure as the real stack is part of the win32 thread. */
+#define configMINIMAL_STACK_SIZE                   ( ( uint16_t ) 70 )              /* In this simulated case, the stack only has to hold one small structure as the real stack is part of the win32 thread. */
```
paulbartell, line 49:
```suggestion
#define configMINIMAL_STACK_SIZE                   ( ( configSTACK_DEPTH_TYPE ) 70U )              /* In this simulated case, the stack only has to hold one small structure as the real stack is part of the win32 thread. */
```

source/FreeRTOS_IP_Utils.c
```diff
@@ -1151,7 +1151,7 @@ const char * FreeRTOS_strerror_r( BaseType_t xErrnum,
             /* MISRA Ref 21.6.1 [snprintf and logging] */
 /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-216 */
             /* coverity[misra_c_2012_rule_21_6_violation] */
-            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int ) xErrnum );
+            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int32_t ) xErrnum );
```
paulbartell, line 1154:
```suggestion
            ( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int ) xErrnum );
```

paulbartell (changes requested):
Types need adjustment.

test/Coverity/ConfigFiles/FreeRTOSConfig.h
AniruddhaKanhere: +1

AniruddhaKanhere (dismissed):

source/FreeRTOS_IP_Utils.c
xuelix: "int" is the violation:  Using basic numerical type "int" rather than a typedef that includes size and signedness information.

paulbartell: "%d" printf strings match an int which is almost always the type of BaseType_t.

int32_t -> 32 bits
int on avr / BaseType_t -> 16 bits
int on arm / BaseType_t -> 32 bits

Thus no cast is necessary.

xuelix: That makes sense.  I will remove the cast

paulbartell (approved):

alfred2g (approved):

source/FreeRTOS_IP_Utils.c
htibosch: @paulbartell 

> Thus no cast is necessary.

I'm afraid that without the cast many compilers ( except 64-bit ) will complain. The `%d` can not be used in combination with `int32_t` or `BaseType_t`, even when they have the same size as a natural `int`.

I wrote "guidelines" about the use of formats and casts [as a comment on PR197](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/197#pullrequestreview-595091959).
~~~
    int int_value;
    int32_t int32_value;
    BaseType_t base_value;
    printf( "int_value = %d\n", int_value );
    printf( "int32_value = %d\n", int32_value );  // Compiler warning
    printf( "base_value = %d\n", base_value );    // Compiler warning
~~~
@xuelix's PR is already merged, When MISRA is ( almost ) ready, I will check it once more.

PR #529, opened 2022-07-29
Misra fix or suppress remaining violations

alfred2g:
Fix Remaining Misra violations

Description
-----------
Fix Remaining and new misra violations

Test Steps
-----------
Misra checker + unit tests

### DIFF
--- a/MISRA.md
+++ b/MISRA.md
@@ -31,6 +31,13 @@ _Ref 8.9.1_
        order of execution, some variables have file scope definitions rather
        than function scope.
 
+#### Rule 11.1
+_Ref 11.1.1_
+
+- MISRA C-2012 Rule 11.1 Converting from a void pointer to a function pointer,
+    it is part of the API and the user is responsible of providing that pointer
+    and their responsibility for that pointer to be valid.
+
 #### Rule 11.3
 _Ref 11.3.1_
 
--- a/MISRA.md
+++ b/MISRA.md
@@ -52,9 +52,13 @@ _Ref 10.5.1_
 #### Rule 11.1
 _Ref 11.1.1_
 
-- MISRA C-2012 Rule 11.1 Converting from a void pointer to a function pointer,
-    it is part of the API and the user is responsible of providing that pointer
-    and their responsibility for that pointer to be valid.
+- MISRA C-2012 Rule 11.1 Converting from a void pointer to a function pointer.
+   The `FreeRTOS_setsockopt` API allows users to configure sockets by setting
+   various options. In order to do so, the function must accept one parameter
+   which, based on the option value, can be casted to the corresponding socket
+   field. To that end, that parameter is of `void *` type to accommodate all values.
+   The caller of the API is responsible for providing correct function pointer to the
+   API. Thus, this violation can be safely suppressed.
 
 #### Rule 11.3
 _Ref 11.3.1_
--- a/MISRA.md
+++ b/MISRA.md
@@ -129,16 +129,6 @@ _Ref 14.3.1_
 - MISRA C-2012 Rule 14.3 False positive as the value might be changed
         depending on the conditionally compiled code
 
-#### Rule 21.6
-_Ref 21.6.1_
-
-- MISRA C-2012 Rule 21.6 warns about the use of standard library input/output
-        functions as they might have implementation defined or undefined
-        behaviour. The function `snprintf` is used to insert information in a
-        logging string. This is only used in a utility function which aids in
-        debugging and is not part of the 'core' code governing the
-        functionality of the TCP/IP stack.
-
 #### Rule 17.2
 _Ref 17.2.1_
 
@@ -153,7 +143,7 @@ _Ref 17.2.1_
 #### Rule 20.5
 _Ref 20.5.1_
 
-- MISRA C-2012 Rule 20.10 warns against the use of #undef
+- MISRA C-2012 Rule 20.5 warns against the use of #undef
         However, in this case, it must be used to make sure some macros are
         defined to a certain value
 
@@ -164,3 +154,14 @@ _Ref 20.10.1_
         However, in this case, it must be used to support compile time
         assertions in case the preprocessor does not suppport sizeof. This
         operation (assert) has no runtime execution.
+
+#### Rule 21.6
+_Ref 21.6.1_
+
+- MISRA C-2012 Rule 21.6 warns about the use of standard library input/output
+        functions as they might have implementation defined or undefined
+        behaviour. The function `snprintf` is used to insert information in a
+        logging string. This is only used in a utility function which aids in
+        debugging and is not part of the 'core' code governing the
+        functionality of the TCP/IP stack.
+
--- a/MISRA.md
+++ b/MISRA.md
@@ -31,6 +31,21 @@ _Ref 8.9.1_
        order of execution, some variables have file scope definitions rather
        than function scope.
 
+#### Rule 8.13
+_Ref 8.13.1_
+
+- MISRA C-2012 Rule 8.13 Parameter passed is never used, should be declared as
+    const. This is a preefined API by the FreeRTOS-kernel function xTaskCreate,
+    and it depends on the user if they want to use that parameter, making it
+    const would break the build.
+
+#### Rule 10.5
+_Ref 10.5.1_
+
+- MISRA C-2012 Rule 10.5 Converting from an unsigned to an enum type. The
+    operation is safe to perform in that case, as we are using a generic API
+    to send and receive data, in that case the exact data sent it is received
+
 #### Rule 11.1
 _Ref 11.1.1_
 
--- a/MISRA.md
+++ b/MISRA.md
@@ -150,9 +150,13 @@ _Ref 17.2.1_
 #### Rule 20.5
 _Ref 20.5.1_
 
-- MISRA C-2012 Rule 20.5 warns against the use of #undef
-        However, in this case, it must be used to make sure some macros are
-        defined to a certain value
+- MISRA C-2012 Rule 20.5 warns against the use of #undef.
+   FreeRTOS-Plus-TCP allows its users to set some configuration macros
+   to modify the behavior/performance of the library according to their
+   needs. However, the macros values must be within certain bounds.
+   To achieve that, if the macro values lie outside of the bounds, they
+   are undefined using `#undef` before being redefined to a proper
+   value.
 
 #### Rule 20.10
 _Ref 20.10.1_
--- a/MISRA.md
+++ b/MISRA.md
@@ -135,6 +135,13 @@ _Ref 17.2.1_
         have a secondary child socket thereby limiting the number of recursive
         calls to one.
 
+#### Rule 20.5
+_Ref 20.5.1_
+
+- MISRA C-2012 Rule 20.10 warns against the use of #undef
+        However, in this case, it must be used to make sure some macros are
+        defined to a certain value
+
 #### Rule 20.10
 _Ref 20.10.1_
 
--- a/MISRA.md
+++ b/MISRA.md
@@ -35,9 +35,12 @@ _Ref 8.9.1_
 _Ref 8.13.1_
 
 - MISRA C-2012 Rule 8.13 Parameter passed is never used, should be declared as
-    const. This is a preefined API by the FreeRTOS-kernel function xTaskCreate,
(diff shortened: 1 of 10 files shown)


source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -611,6 +617,9 @@
 #else
     /* A sanity check to avoid a possible overflow of size_t. */
     #if ipconfigNETWORK_MTU > ( SIZE_MAX >> 1 )
```
paulbartell, line 619: This should just trigger a #error.

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -282,6 +285,9 @@
 
 #ifndef FreeRTOS_printf
     #define FreeRTOS_printf( MSG )    do {} while( ipFALSE_BOOL )
+    /* MISRA Ref 20.5.1 [Misaligned access] */
```
paulbartell, line 288: Similarly, this should just trigger a compile time error.

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -259,6 +259,9 @@
 
 #ifndef FreeRTOS_debug_printf
     #define FreeRTOS_debug_printf( MSG )    do {} while( ipFALSE_BOOL )
+    /* MISRA Ref 20.5.1 [Misaligned access] */
```
paulbartell, line 262: "Misaligned access" doesn't make sense here.

MISRA.md
```diff
@@ -31,6 +31,28 @@ _Ref 8.9.1_
        order of execution, some variables have file scope definitions rather
        than function scope.
 
+#### Rule 8.13
+_Ref 8.13.1_
+
+- MISRA C-2012 Rule 8.13 Parameter passed is never used, should be declared as
+    const. This is a preefined API by the FreeRTOS-kernel function xTaskCreate,
+    and it depends on the user if they want to use that parameter, making it
+    const would break the build.
```
AniruddhaKanhere, line 40:
```suggestion
    const.  The argument passed to the `prvIPTask` function is left unused which is
    considered as the variable not being used and thus warranting the use of `const`.
    However, the FreeRTOS-kernel function `xTaskCreate` expects a function signature
    of type `void vSomeFunction( void * pvArgs )`. To satisfy that requirement, the
    function signature of `prvIPTask` does not have a `const` qualifier in the
    parameter signature.
```

MISRA.md
```diff
@@ -31,6 +31,31 @@ _Ref 8.9.1_
...
+- MISRA C-2012 Rule 10.5 Converting from an unsigned to an enum type. The
+    operation is safe to perform in that case, as we are using a generic API
+    to send and receive data, in that case the exact data sent it is received
+
+#### Rule 11.1
+_Ref 11.1.1_
+
+- MISRA C-2012 Rule 11.1 Converting from a void pointer to a function pointer,
+    it is part of the API and the user is responsible of providing that pointer
+    and their responsibility for that pointer to be valid.
```
AniruddhaKanhere, line 57:
```suggestion
- MISRA C-2012 Rule 11.1 Converting from a void pointer to a function pointer.
   The `FreeRTOS_setsockopt` API allows users to configure sockets by setting
   various options. In order to do so, the function must accept one parameter
   which, based on the option value, can be casted to the corresponding socket
   field. To that end, that parameter is of `void *` type to accommodate all values.
   The caller of the API is responsible for providing correct function pointer to the
   API. Thus, this violation can be safely suppressed.
```

MISRA.md
```diff
@@ -128,10 +143,28 @@ _Ref 17.2.1_
         have a secondary child socket thereby limiting the number of recursive
         calls to one.
 
+#### Rule 20.5
+_Ref 20.5.1_
+
+- MISRA C-2012 Rule 20.5 warns against the use of #undef
+        However, in this case, it must be used to make sure some macros are
+        defined to a certain value
```
AniruddhaKanhere, line 151:
```suggestion
- MISRA C-2012 Rule 20.5 warns against the use of #undef.
   FreeRTOS-Plus-TCP allows its users to set some configuration macros
   to modify the behavior/performance of the library according to their
   needs. However, the macros values must be within certain bounds.
   To achieve that, if the macro values lie outside of the bounds, they
   are undefined using `#undef` before being redefined to a proper
   value.
```

source/FreeRTOS_Sockets.c
```diff
@@ -1957,8 +1965,13 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                          * type "B" removes const qualifier from the pointed to type. */
 
                         /* MISRA Ref 11.8.1 [Function pointer and use of const pointer] */
-/* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-118 */
+                        /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-118 */
+
+                        /* MISRA Ref _Ref 11.1.1_ [ Conversion between pointer to
```
AniruddhaKanhere, line 1970:
```suggestion
                        /* MISRA Ref 11.1.1 [ Conversion between pointer to
```

source/portable/BufferManagement/BufferAllocation_2.c
```diff
@@ -393,30 +399,34 @@ NetworkBufferDescriptor_t * pxResizeNetworkBufferWithDescriptor( NetworkBufferDe
 {
     size_t xOriginalLength;
     uint8_t * pucBuffer;
+    size_t xSizeBytes = xNewSizeBytes;
```
AniruddhaKanhere, line 402:
```suggestion
    size_t uxSizeBytes = xNewSizeBytes;
```

Also the function parameter `xNewSizeBytes` should be `uxNewSizeBytes`.

MISRA.md
```diff
@@ -31,6 +31,31 @@ _Ref 8.9.1_
        order of execution, some variables have file scope definitions rather
        than function scope.
 
+#### Rule 8.13
+_Ref 8.13.1_
+
+- MISRA C-2012 Rule 8.13 Parameter passed is never used, should be declared as
```
paulbartell, line 37: This should be re-worded to be more generic. Something like:

MISRA C-2012 Rule 8.13 encourages the use of const-qualified types whenever possible. However, the FreeRTOS API requires that all task entry functions conform to the same function signature (TaskFunction_t) which includes a single void * pointer argument. Due to this restriction, it is not possible to use a const qualifier for the entry point function to any FreeRTOS task.

AniruddhaKanhere (changes requested):

MISRA.md
AniruddhaKanhere: I think it was fixed with this commit: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/529/commits/81e90d8ac5bf905159c46b0675b957876336c182

MISRA.md
```diff
@@ -128,10 +143,28 @@ _Ref 17.2.1_
...
 
 - MISRA C-2012 Rule 20.10 warns against the use of ## concatination operator.
         However, in this case, it must be used to support compile time
         assertions in case the preprocessor does not suppport sizeof. This
         operation (assert) has no runtime execution.
+
+#### Rule 21.6
+_Ref 21.6.1_
+
+- MISRA C-2012 Rule 21.6 warns about the use of standard library input/output
```
paulbartell, line 172: This justification should mention how we prevent misuse of printf-style functions. Mentioning "-Wformat-security" and/or specific Wformat options that are used when building unit tests would be helpful here.

MISRA.md
paulbartell: IMO: It's too specific to the particular instance. We're likely to run into this exact violation quite frequently.

MISRA.md
```diff
@@ -31,6 +31,35 @@ _Ref 8.9.1_
...
+- MISRA C-2012 Rule 11.1 Converting from a void pointer to a function pointer.
+   The `FreeRTOS_setsockopt` API allows users to configure sockets by setting
+   various options. In order to do so, the function must accept one parameter
+   which, based on the option value, can be casted to the corresponding socket
+   field. To that end, that parameter is of `void *` type to accommodate all values.
+   The caller of the API is responsible for providing correct function pointer to the
+   API. Thus, this violation can be safely suppressed.
+
 #### Rule 11.3
 _Ref 11.3.1_
```
paulbartell, line 65: What about for 64-bit targets that do not allow 32-bit aligned access?

source/FreeRTOS_IP.c
```diff
@@ -390,9 +394,11 @@ static void prvProcessIPEventsAndTimers( void )
                     eDHCPState_t eState;
 
                     /* MISRA Ref 11.6.1 [DHCP events and conversion to void] */
```
paulbartell, line 396:
```suggestion
                    /* MISRA Ref 11.6.1 [DHCP events and conversion from void *] */
```

MISRA.md
alfred2g: it is ok to be specific might be even desirable, we do support multiple References for specific cases

MISRA.md
alfred2g: We currently do not support any hardware with that feature/drawback the only 64 bit we currently support is RiscV and they do support unaligned access (if i am not mistaken)
when we support more 64bit hardware we will have to add code to cover that case.

source/portable/BufferManagement/BufferAllocation_2.c
```diff
@@ -293,12 +296,15 @@ NetworkBufferDescriptor_t * pxGetNetworkBufferWithDescriptor( size_t xRequestedS
                      * buffer storage area, then move the buffer pointer on past the
                      * stored pointer so the pointer value is not overwritten by the
                      * application when the buffer is used. */
+                    /* MISRA Ref 11.3.1 [Misaligned access] */
+                    /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-113 */
+                    /* coverity[misra_c_2012_rule_11_3_violation] */
                     *( ( NetworkBufferDescriptor_t ** ) ( pxReturn->pucEthernetBuffer ) ) = pxReturn;
```
htibosch, line 302:
~~~
*( ( NetworkBufferDescriptor_t ** ) ( pxReturn->pucEthernetBuffer ) ) = pxReturn;
~~~
In other places in the library, I have put asserts before assuming that a pointer can be written at a byte address. In this case, the character pointer was returned by `pvPortMalloc()`, and it is assumed that it is 4-byte aligned.

On my Xilinx UltraScale, the size of a pointer is 8 and `pvPortMalloc()` will make sure that only 64-bit aligned pointers are returned, see the `portBYTE_ALIGNMENT` macros for ARM_CA53_64_BIT:
~~~
#define portBYTE_ALIGNMENT         16
#define portBYTE_ALIGNMENT_MASK    0x000f 
~~~
In case of a 64-bit platform, the macro `ipconfigBUFFER_PADDING` will be checked with asserts: it must be at least 14, and it must be a multiple of 4 **plus** 2 bytes:
~~~
    configASSERT( ( sizeof( uintptr_t ) != 8U ) || ( ipconfigBUFFER_PADDING >= 14 ) );

    /* But it must have this strange alignment: */
    configASSERT( ( sizeof( uintptr_t ) != 8U ) || ( ( ( ( ipconfigBUFFER_PADDING ) + 2 ) % 4 ) == 0 ) );
~~~
When `ipconfigBUFFER_PADDING` is defined, it will be assigned to `ipBUFFER_PADDING`.

htibosch (approved):
@alfred2g  It sounds like the MISRA checks are almost all ready!
I only have two comments, mostly about 64-bit platforms.
I approve it, thanks

MISRA.md
paulbartell: The Cortex-A53 is only strict about alignment when configured such in the SCTLR. It's pretty common to configure it to allow unaligned access.

In contrast, a Cortex-M0 (ARMv6-M) will always fail unaligned accesses.. but is 4-byte aligned.

SPARC and MIPS are both example of a 64bit architectures that do not support unaligned access... So de-referencing a 32-bit aligned pointer as a 64bit type could cause a fault at runtime.

In the case of FR+TCP, we're unlikely to run into this problem in particular.

paulbartell (approved):

AniruddhaKanhere (approved):

PR #530, opened 2022-07-29
MISRA violation fix 17.7, 1.2, 8.9

xuelix:
Fixing MISRA violation 17.7, 1.2 and 8.9
Description
-----------

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_DNS_Cache.c
+++ b/source/FreeRTOS_DNS_Cache.c
@@ -173,9 +173,9 @@
         { /* Element found */
             if( xLookUp == pdTRUE )
             {
-                prvGetCacheIPEntry( uxIndex,
-                                    pulIP,
-                                    ulCurrentTimeSeconds );
+                ( void ) prvGetCacheIPEntry( uxIndex,
+                                             pulIP,
+                                             ulCurrentTimeSeconds );
             }
             else
             {
--- a/source/FreeRTOS_DNS_Cache.c
+++ b/source/FreeRTOS_DNS_Cache.c
@@ -173,6 +173,10 @@
         { /* Element found */
             if( xLookUp == pdTRUE )
             {
+                /* This statement can only be reached when xResult is true; which
+                 * implies that the entry is present and a 'get' operation will result
+                 * in success. Therefore, it is safe to ignore the return value of the
+                 * below function. */
                 ( void ) prvGetCacheIPEntry( uxIndex,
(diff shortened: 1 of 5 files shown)


source/FreeRTOS_DNS_Cache.c
```diff
@@ -173,9 +173,9 @@
         { /* Element found */
             if( xLookUp == pdTRUE )
             {
-                prvGetCacheIPEntry( uxIndex,
-                                    pulIP,
-                                    ulCurrentTimeSeconds );
+                ( void ) prvGetCacheIPEntry( uxIndex,
```
AniruddhaKanhere, line 176: Maybe we should add a comment about why the result is being discarded here and why is it safe to do so.
Something along the lines of:
```
/* This statement can only be reached when xResult is true; which
 * implies that the entry is present and a 'get' operation will result
 * in success. Therefore, it is safe to ignore the return value of the
 * below function. */
```

source/FreeRTOS_IP_Utils.c
```diff
@@ -321,6 +321,10 @@ BaseType_t xIsCallingFromIPTask( void )
 /**
  * @brief Process a 'Network down' event and complete required processing.
  */
+/* MISRA Ref 8.9.1 [File scoped variables] */
+/* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-89 */
+/* coverity[misra_c_2012_rule_8_9_violation] */
+/* coverity[single_use] */
```
AniruddhaKanhere, line 327: Does this suppress the violation in the report? If I remember correctly, we had found that doing this doesn't help.

AniruddhaKanhere (commented):
Thanks for this PR @xuelix.
I have a couple of comments. Can you please take a look?

source/FreeRTOS_IP_Utils.c
xuelix: It does suppress the violation in the report

source/FreeRTOS_DNS_Cache.c
xuelix: We have too many (void) in front of functions (e.g. thousands of them including memcpy) .  I don't think we need to add for one specific function calls.

n9wxu (approved):

PR #531, opened 2022-07-30
IPv4/single: some minor comments after the MISRA changes

htibosch:
Description
-----------
- FreeRTOS_DNS.c

Use `xReturn` and not `uxReturn` when the variable is signed.
Call `memset()` to clear the contents of a struct.

I have seen applications crash on such a statement:
~~~diff
-    struct xDNSBuffer xReceiveBuffer = { 0 };
+    struct xDNSBuffer xReceiveBuffer;
+    memset( xReceiveBuffer, 0, sizeof( xReceiveBuffer ) );
~~~
because the compile can be mistaken about the alignment.

- FreeRTOS_DNS_Networking.c

Check the return value of `FreeRTOS_recvfrom()` before assigning it to an unsigned variable:
~~~diff
-    pxReceiveBuffer->uxPayloadLength = FreeRTOS_recvfrom( xDNSSocket, ... );
-    pxReceiveBuffer->uxPayloadSize = pxReceiveBuffer->uxPayloadLength;
+    lResult = FreeRTOS_recvfrom( xDNSSocket, ... );
+    if( lResult > 0 )
+    {
+        pxReceiveBuffer->uxPayloadLength = ( size_t ) lResult;
+        pxReceiveBuffer->uxPayloadSize = pxReceiveBuffer->uxPayloadLength;
+    }
~~~
The result of `FreeRTOS_recvfrom()` must be tested because a negative value indicates an error. 

- FreeRTOS_IP_Utils.c

As I commented [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/527#discussion_r933723703), cast a `BaseType_t` variable to `int` when using `snprintf()`:
~~~diff
-( void ) snprintf( pcBuffer, uxLength, "Errno %d", xErrnum );
+( void ) snprintf( pcBuffer, uxLength, "Errno %d", ( int ) xErrnum );
~~~

- FreeRTOS_IP.h

Let `ipFR_TCP_VERSION_NUMBER` depend on the other 3 version macros:
~~~diff
-    #define ipFR_TCP_VERSION_NUMBER    "V2.3.999"
     #define ipFR_TCP_VERSION_MAJOR     2
     #define ipFR_TCP_VERSION_MINOR     3
     /* Development builds are always version 999. */
     #define ipFR_TCP_VERSION_BUILD     999
+    #define __VERSION( v1, v2, v3 )    "V" # v1 "." # v2 "." # v3
+    #define _VERSION( v1, v2, v3 )     __VERSION( v1, v2, v3 )
+
+    /* ipFR_TCP_VERSION_NUMBER makes a string like "V2.3.999". */
+    #define ipFR_TCP_VERSION_NUMBER \
+        _VERSION( ipFR_TCP_VERSION_MAJOR, ipFR_TCP_VERSION_MINOR, ipFR_TCP_VERSION_BUILD )
~~~
but I think that Aniruddha has a more elegant solution.

I compare today's mainline with the release on 7th of June. Beside a few minor things I think it is all good. Thanks.


source/FreeRTOS_DNS_Networking.c
```diff
@@ -131,15 +131,21 @@
...
+                                     0,
+                                     FREERTOS_ZERO_COPY,
+                                     xAddress,
+                                     &ulAddressLength );
+
+        if( lResult > 0 )
+        {
+            pxReceiveBuffer->uxPayloadLength = ( size_t ) lResult;
+            pxReceiveBuffer->uxPayloadSize = pxReceiveBuffer->uxPayloadLength;
+        }
```
AniruddhaKanhere, line 148: A very good catch Hein! Thank you.
In an else, maybe we should set both of the above fields to zero?
```suggestion
        if( lResult > 0 )
        {
            pxReceiveBuffer->uxPayloadLength = ( size_t ) lResult;
            pxReceiveBuffer->uxPayloadSize = pxReceiveBuffer->uxPayloadLength;
        }
        else
        {
            pxReceiveBuffer->uxPayloadLength = 0U;
            pxReceiveBuffer->uxPayloadSize = 0U;
        }
```

source/include/FreeRTOS_IP.h
```diff
@@ -44,11 +44,15 @@
...
-#define ipFR_TCP_VERSION_NUMBER    "V2.3.999"
-#define ipFR_TCP_VERSION_MAJOR     2
-#define ipFR_TCP_VERSION_MINOR     3
+#define ipFR_TCP_VERSION_MAJOR    2
+#define ipFR_TCP_VERSION_MINOR    3
 /* Development builds are always version 999. */
-#define ipFR_TCP_VERSION_BUILD     999
+#define ipFR_TCP_VERSION_BUILD    999
+#define __VERSION( v1, v2, v3 )    "V" # v1 "." # v2 "." # v3
+#define _VERSION( v1, v2, v3 )     __VERSION( v1, v2, v3 )
```
AniruddhaKanhere, line 52: MISRA doesn't like the use of `#` operator. Ref: Misra c-2012 rule 20.10 (advisory).

Also, these version numbers are updated together - and only when a release happens. Not completely sold on the value added by using the above expression to generate the `ipFR_TCP_VERSION_NUMBER` string.

AniruddhaKanhere (commented):
Thank you for this PR @htibosch. I have a couple of comments/suggestions. Can you please take a look at that?

source/include/FreeRTOS_IP.h
htibosch: I have cancelled the change to the macro `ipFR_TCP_VERSION_NUMBER`.

source/FreeRTOS_DNS_Networking.c
```diff
@@ -142,6 +142,7 @@
     {
         BaseType_t xReturn;
         uint32_t ulAddressLength = ( uint32_t ) sizeof( struct freertos_sockaddr );
+        int32_t lResult;
```
ActoryOu, line 145:
```suggestion
```

ActoryOu:
The logic added in DNS_ReadReply is not needed. Removed that part and update the branch.

ActoryOu (approved):

shubnil (approved):

PR #528, opened 2022-07-28
IPv6/Multi : (IPv6TestingBranch) Update folder structure

ActoryOu:
IPv6/Multi : IPv6TestingBranch Update folder structure

Description
-----------

Follow #466 to update folder structure in labs/ipv6_multi.
- Add source folder.
- move include, portable directories into source folder.
- move all source files into source folder.

Test Steps
-----------

Use following commands to make sure it can pass compilation.
- cmake -S test/unit-test -B test/unit-test/build
- make -C test/unit-test/build/ all

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #439, opened 2022-03-28
Unit-test TCP_IP.c file

AniruddhaKanhere:
Description
-----------

Test Steps
-----------

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #422, opened 2022-02-08
Split FreeRTOS_TCP_IP.c to 5 files.

xuelix:
Split FreeRTOS_TCP_IP.c to 5 files,
FreeRTOS_TCP_Reception.c
FreeRTOS_TCP_Transmission.c
FreeRTOS_TCP_State_Handling.c
FreeRTOS_TCP_Utils.c

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #532, opened 2022-08-02
IPv6/Multi : (IPv6TestingBranch) Decouple TCP layer

ActoryOu:
IPv6/Multi : (IPv6TestingBranch) Decouple TCP layer

Description
-----------

Based on #528, follow #422 and #439 to split TCP layers in labs/ipv6_multi.

Split FreeRTOS_TCP_IP.c/h to several files.
- FreeRTOS_TCP_IP.h/c
- FreeRTOS_TCP_Reception.h/c
- FreeRTOS_TCP_State_Handling.h/c
- FreeRTOS_TCP_Transmission.h/c
- FreeRTOS_TCP_Utils.h/c

Test Steps
-----------

Use following commands to make sure it can pass compilation.

- cmake -S test/unit-test -B test/unit-test/build
- make -C test/unit-test/build/ all

Related Issue
-----------


htibosch (commented):
When we started splitting up the source files into smaller parts, it is inevitable that users change their Make procedures. This one works pretty well:

`C_SRCS += $(wildcard $(PLUS_TCP_PATH)/source/*.c)`

because it includes all C-files.

However, it would be nice if users do not have to worry about including the new **header files**:

For instance, FreeRTOS_IP.h should include the following sub-files:

    FreeRTOS_IP_Private.h
    FreeRTOS_IP_Timers.h
    FreeRTOS_IP_Utils.h

You may want to check the IPv4 branch to see what #include statements are found in the header files.

Thank you  for all the cleaning up, like removing the cast macro's:
`ipDECL_CAST_CONST_PTR_FUNC_FOR_TYPE`.

As you know, there were many recent PR's about MISRA in the main branch. From now on there will be `coverity[]` comments in those places where a rule is ignored, along with a proper explanation that explains why the rule can be safely ignored.

It is very complicated to review this PR by looking at a diff. 

I just tested your 3 PR's: `decouple_ip`, `split_ip`, and `split_dns`. All tests were passed correctly. The only problem occurred during compilation because the new header files were not automatically included.

After you've added the appropriate `#include`s, I will approve this PR.

ActoryOu:
Thanks for comments. Let's note the conclusion here.
Most of new header files should be used inside FreeRTOS-Plus-TCP only.
So we don't need to include all of them to the oringinal header files for users.

Thanks.

PR #399, opened 2021-12-21
Dns decouple

alfred2g:
Decouple and refactor DNS

Description
-----------
Refactor DNS to be unit testable
add unit test

Test Steps
-----------
Unit Test passes

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #537, opened 2022-08-04
IPv6/Multi : (IPv6TestingBranch) Decouple DNS

ActoryOu:
IPv6/Multi : (IPv6TestingBranch) Decouple DNS

Description
-----------

Based on #535, follow #399, #406 and #431 to split DNS files in labs/ipv6_multi.

Split FreeRTOS_DNS.c/h to several files.
- FreeRTOS_DNS_Globals.h
- FreeRTOS_DNS_Cache.c/h
- FreeRTOS_DNS_Callback.c/h
- FreeRTOS_DNS_Netwroking.c/h
- FreeRTOS_DNS_Utils.c/h
- FreeRTOS_DNS.c/h

Test Steps
-----------

Use following commands to make sure it can pass compilation.
- cmake -S test/unit-test -B test/unit-test/build
- make -C test/unit-test/build/ all

Related Issue
-----------


htibosch (approved):
This PR proposes many changes, while the functionality of the library will not change at all.

@ActoryOu has taken the [IPv4/single branch](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP) as an example, and he has split-up the big source files in the same way.

Along the way he did some clean-ups, like removing the old `ipCAST_PTR_TO_TYPE_PTR` and `ipDECL_CAST_PTR_FUNC_FOR_TYPE` macros.

These are the new source files:

    /source/FreeRTOS_DNS_Cache.c
    /source/FreeRTOS_DNS_Callback.c
    /source/FreeRTOS_DNS_Networking.c
    /source/FreeRTOS_DNS_Parser.c
    /source/FreeRTOS_ICMP.c
    /source/FreeRTOS_IP_Timers.c
    /source/FreeRTOS_IP_Utils.c
    /source/FreeRTOS_TCP_Reception.c
    /source/FreeRTOS_TCP_State_Handling.c
    /source/FreeRTOS_TCP_Transmission.c
    /source/FreeRTOS_TCP_Utils.c

Yesterday I test-compiled projects in several configurations, both using IPv4 and IPv6. I checked if none of the code was broken.

These are the new header files:

    /source/include/FreeRTOS_DNS_Cache.h
    /source/include/FreeRTOS_DNS_Callback.h
    /source/include/FreeRTOS_DNS_Globals.h
    /source/include/FreeRTOS_DNS_Networking.h
    /source/include/FreeRTOS_DNS_Parser.h
    /source/include/FreeRTOS_ICMP.h
    /source/include/FreeRTOS_IP_Timers.h
    /source/include/FreeRTOS_IP_Utils.h
    /source/include/FreeRTOS_TCP_Reception.h
    /source/include/FreeRTOS_TCP_State_Handling.h
    /source/include/FreeRTOS_TCP_Transmission.h
    /source/include/FreeRTOS_TCP_Utils.h

None of the existing code should get broken because of the new header files. When needed, the new header files are included automatically.

The only thing that needs to change is the choice of source files, unless the make wildcard function is used:
~~~make
    C_SRCS += \
        $(wildcard $(PLUS_TCP_PATH)/*.c) \
        $(wildcard $(PLUS_TCP_PATH)/source/*.c)
~~~
The above macro works for both old and new libraries.

Also I have tested the code in real hardware ( STM32F4x ), both in an IPv4 as well as in an IPv6 project.

Thank you @ActoryOu

htibosch (approved):
Thank you @ActoryOu for this commit.

In the existing code, some of the incoming packets would get dropped because the definition of `ipFRAGMENT_OFFSET_BIT_MASK` within NetworkInterface.c was only valid for a big-endian platform.

In my case, DNS lookup's never got a reply because packets got dropped here:
~~~c
    /* Ensure that the incoming packet is not fragmented (only outgoing packets
     * can be fragmented) as these are the only handled IP frames currently. */
    if( ( pxIPHeader->usFragmentOffset & ipFRAGMENT_OFFSET_BIT_MASK ) != 0U )
    {
        return pdFALSE;
    }
~~~
This PR will make the definition of the `ipFRAGMENT` macros depend on `ipconfigBYTE_ORDER`, just like we did in the main branch.

I approve this commit.

source/FreeRTOS_DHCPv6.c
```diff
@@ -379,7 +379,7 @@
 
         /* Check for clashes. */
 
-        vIPReloadDHCP_RATimer( ( struct xNetworkEndPoint * ) pxEndPoint, EP_DHCPData.ulLeaseTime );
+        vDHCP_RATimerReload( ( struct xNetworkEndPoint * ) pxEndPoint, EP_DHCPData.ulLeaseTime );
```
htibosch, line 382: Oops, FreeRTOS_DHCPv6.c also needs to include FreeRTOS_IP_Timers.h, possibly here:

~~~diff
     #include "FreeRTOS_Sockets.h"
+    #include "FreeRTOS_IP_Timers.h"

     #include "FreeRTOS_BitConfig.h"
~~~

htibosch (commented):
And also, the function `prvPrepareLinkLayerIPLookUp()` needs to become available for both `FreeRTOS_DHCP.c` as well as `FreeRTOS_DHCPv6.c`.

I have put it at the bottom of `FreeRTOS_DHCP.h` like this:
~~~c
    #if ( ipconfigDHCP_FALL_BACK_AUTO_IP != 0 )
        struct xNetworkEndPoint;
/**
 * @brief When DHCP has failed, the code can assign a Link-Layer
 *        address, and check if another device already uses the IP-address.
 *
 * param[in] pxEndPoint: The end-point that wants to obtain a link-layer address.
 */
        void prvPrepareLinkLayerIPLookUp( struct xNetworkEndPoint * pxEndPoint );
    #endif
~~~
Thanks,

source/FreeRTOS_DHCPv6.c
ActoryOu: Thanks for checking. Added at commit [68db748](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/537/commits/68db74848c67189b6a62c432603f2a70d00145c8).

source/include/FreeRTOS_DHCP.h
```diff
@@ -133,6 +125,18 @@
...
+    #if ( ipconfigDHCP_FALL_BACK_AUTO_IP != 0 )
+        struct xNetworkEndPoint;
+
+/**
+ * @brief When DHCP has failed, the code can assign a Link-Layer
+ *        address, and check if another device already uses the IP-address.
+ *
+ * param[in] pxEndPoint: The end-point that wants to obtain a link-layer address.
+ */
+        void prvPrepareLinkLayerIPLookUp( NetworkEndPoint_t * pxEndPoint );
```
htibosch, line 137: Traditionally, "FreeRTOS_Routing.h" and "FreeRTOS_ND.h" are included as the last header files. That means that the earlier header files like "FreeRTOS_DHCP.h" can not refer to typedefs made in "FreeRTOS_Routing.h".

That is the reason that:
~~~diff
-void prvPrepareLinkLayerIPLookUp( NetworkEndPoint_t * pxEndPoint );
+void prvPrepareLinkLayerIPLookUp( struct xNetworkEndPoint * pxEndPoint );
~~~

ActoryOu: Fixed here: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/537/commits/75492057e5512ca002cec9bccc458c59522127af

htibosch (approved):
Hello @ActoryOu , once more I tested your [split_dns](https://github.com/ActoryOu/FreeRTOS-Plus-TCP/tree/split_dns) branch, and it looks all good to me:
- It compiles correctly.
- It runs well in my STM32F4x testing project.
- Also I ran the IPv6_Multi_WinSim_demo project, which passed all internal tests.

The test-set consisted of: DNS queries, ICMP/ping, HTTP client, and a query to NTP. All tests were done both using IPv4 as well as IPv6. See [pcCommandList](https://github.com/ActoryOu/FreeRTOS-Plus-TCP/blob/75492057e5512ca002cec9bccc458c59522127af/demos/IPv6_Multi_WinSim_demo/main.c#L518) in main.c.

As soon as this PR is merged, I will create a PR to realise a few changes, mostly related to DNS.

n9wxu (approved):

PR #544, opened 2022-08-31
New helper function: FreeRTOS_get_tx_base

htibosch:
Description
-----------
In a [post](https://forums.freertos.org/t/stm32h743-freertos-tcp-issue-with-zero-copy/15390)  on the FreeRTOS forum, [@arijav](https://forums.freertos.org/u/arijav) asked for a new function `FreeRTOS_get_tx_base()`.

The existing `FreeRTOS_get_tx_head()` can be used to obtain a pointer to the head position in the TX buffer. It implements "zero copy": after filling the buffer, `FreeRTOS_send()` can be called with the buffer address `NULL`. The memcpy will be skipped.

The new `FreeRTOS_get_tx_base()` will return a pointer to the TX buffer itself.

Note that both functions are classified as "advanced". For Javier it was very helpful: he could increase the performance.

Test Steps
-----------
I tested both functions by calling them before and after the first call to `FreeRTOS_send()`.

I decided that both functions should create the TX buffer in case it has not been created yet.


arijav:
Hi Hain, sorry for giving my feedback so late, the modifications work well. I just have to change for my application, as I use fill-in directly the FreeRTOS TCP buffer, some changes in freertos_sockets.c (line 1760):

                /* Round up to nearest MSS size */
                /* Modification */
                //ulNewValue = FreeRTOS_round_up( ulNewValue, ( uint32_t ) pxSocket->u.xTCP.usMSS );
                pxSocket->u.xTCP.uxTxStreamSize = ulNewValue;

Rounding up does not work for me.

BR,

Javier

shubnil:
Hi Hein, Do we still this need this change. If yes, it will be good if we can create a PR for dev/ipv6_integration branch.

source/include/FreeRTOS_Sockets.h
```diff
@@ -367,6 +367,13 @@
 /* For internal use only: return the connection status. */
         BaseType_t FreeRTOS_connstatus( ConstSocket_t xSocket );
 
+/* For advanced applications only:
+ * Get a direct pointer to the beginning of the circular transmit buffer.
+ * In case the buffer was not yet created, it will be created in
+ * this call.
+ */
+        uint8_t * FreeRTOS_get_tx_base( Socket_t xSocket );
```
ActoryOu, line 375: Should we wrap this with ipconfigUSE_TCP?

htibosch: It is already conditional on `ipconfigUSE_TCP`:
~~~
    #if ( ipconfigUSE_TCP == 1 )           // line 270
        #if ( ipconfigUSE_TCP == 1 )       // line 238
            BaseType_t FreeRTOS_ReleaseTCPPayloadBuffer( Socket_t xSocket,
                                                         void const * pvBuffer,
                                                         BaseType_t xByteCount );
        #endif /* ( ipconfigUSE_TCP == 1 ) */  // line 336
        uint8_t * FreeRTOS_get_tx_base( Socket_t xSocket );  // line 372
    #endif /* ( ipconfigUSE_TCP == 1 ) */      // line 393
~~~

I will remove the nested `#if/#endif` couple.

htibosch:
Before this PR we had already `FreeRTOS_get_rx_buf()`, which returns `xTCP.rxStream`. The new function `FreeRTOS_get_tx_base()` will return `xTCP.txStream`.

It would be nice if the two have comparable names, so I propose to call the new function `FreeRTOS_get_tx_buf()`.

The `_base()` version hasn't been released yet, it was only [a proposal](https://forums.freertos.org/t/stm32h743-freertos-tcp-issue-with-zero-copy/15390/51), so I don't mind mentioning `FreeRTOS_get_tx_base()` in a comment, so it can be found.

What do you think?

moninom1 (approved):

tony-josi-aws (approved):

shubnil (approved):

PR #545, opened 2022-08-31
IPv4/single: Avoid a hanging reuse socket

htibosch:
Description
-----------
In a [post](https://forums.freertos.org/t/sockets-end-in-state-eclose-wait) on the FreeRTOS forum, [Hubert Sack](https://forums.freertos.org/u/hs4FreeRtos) reported that a problem occurs when the SYN/ACK negotiation never finishes.

The orphaned socket will be marked as closed and deleted.

In case of a reuse socket, it should not be deleted because the application uses it as a server socket.

Solution: when a timeout is reached, the socket shall be put into `eTCP_LISTEN` mode again, so it can receive a new connection.

PS. A reuse socket is a TCP socket that turns itself into a client socket as soon as a connection is established. This method has two advantages: it uses less RAM, and it allows for a single connection only.

Test Steps
-----------
Have a client which only sends a SYN packet and then stop. After a timeout, the socket should be able to receive a new connection.

### DIFF
--- a/source/FreeRTOS_TCP_State_Handling.c
+++ b/source/FreeRTOS_TCP_State_Handling.c
@@ -191,9 +191,31 @@
                 /* ipconfigTCP_HANG_PROTECTION_TIME is in units of seconds. */
                 if( xAge > ( ( TickType_t ) ipconfigTCP_HANG_PROTECTION_TIME * ( TickType_t ) configTICK_RATE_HZ ) )
                 {
+                    BaseType_t xHandled = pdFALSE;
+
+                    if( pxSocket->u.xTCP.bits.bReuseSocket )
+                    {
+                        switch( pxSocket->u.xTCP.eTCPState )
+                        {
+                            case eSYN_FIRST:    /* 3 (server) Just created, must ACK the SYN request */
+                            case eSYN_RECEIVED: /* 4 (server) waiting for a confirming connection request */
+                                                /* acknowledgement after having both received and sent a connection request */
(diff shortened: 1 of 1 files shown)


htibosch:
Finally I have setup a situation that simulates the problem, and I could see that the server socket recovered well.
Note that this only applies to sockets that have the `FREERTOS_SO_REUSE_LISTEN_SOCKET` socket option.

A telnet client sent a SYN, it was answered by a SYN/ACK, and from then, all incoming packets were dropped.
When the `ipconfigTCP_HANG_PROTECTION` noticed that the socket was inactive for 30 seconds, it was put in the `eTCP_LISTEN` state again.
A next connection would succeed normally again.

htibosch:
As I am currently also working on [another issue](https://forums.freertos.org/t/freertos-connect-and-accept-functions-in-freertos-tcp-do-not-return-even-after-the-tcp-connection-closes/15787/11), I think it is better to close this PR for now, and integrate the it with the other [upcoming PR](https://github.com/htibosch/FreeRTOS-Plus-TCP/tree/IPv4_single_connect_returns_on_socket_closure).

A summary: the normal path of a TCP connection is either:
"eTCP_LISTEN, eSYN_FIRST, eSYN_RECEIVED, eESTABLISHED, eCLOSE_WAIT"
or
"eCONNECT_SYN, eESTABLISHED, eCLOSE_WAIT"

This and the other PR handle the situation in which the status goes directly from SYN to CLOSED.

To be continued. Hein

htibosch:
Now closing, to be continued in another PR.

htibosch:
In PR #559 I addressed the above issue.

Issue #543, opened 2022-08-28
[Feature Request] Add vTimerSetTimerID/pvTimerGetTimerID equivalents for sockets.


karver8:
**Is your feature request related to a problem? Please describe.**
There is no easy way to locate application data associated with a socket from a socket callback. 

Sockets allow the application to set callback functions for various events, e.g. `FREERTOS_SO_TCP_RECV_HANDLER` which will call a function with the prototype `BaseType_t (* FOnTCPReceive_t )( Socket_t xSocket, void * pData, size_t xLength )`. However if the same callback function is assigned to multiple sockets, there is no easy way to "identify" the socket using just its handle. You could define a data structure containing sockets paired with their associated objects and search through this, however this is not very efficient. 

**Describe the solution you'd like**
Add vTimerSetTimerID/pvTimerGetTimerID equivalents to the sockets API and add a `void * pvSocketID` member to the `IPTCPSocket_t` struct.

These could look like: 

* vSocketSetSocketID/pvSocketGetSocketID
* FreeRTOS_setsocketid/FreeRTOS_getsocketid (more in line with the libraries naming style)

**Describe alternatives you've considered**
A SOCKET_ID option could be added to `FreeRTOS_setsockopt()`, and the callback prototypes could be modified to include the socket ID. This however ruins backwards compatibility.

AniruddhaKanhere:
Hello @karver8!

That is quite an interesting feature request. Thank you for bringing this to out attention and thank you for also proposing a solution. The solution you suggest will seem like it would work.
We'll work on this and get back to you. Till that time, if you happen to have any more suggestions on how can we improve the FreeRTOS+TCP library, feel free to open more feature requests.

Thanks again,
Aniruddha

htibosch:
Hello [Jacob](https://github.com/karver8),

I agree with Aniruddha that this is an interesting request.

At one point, we already provide the possibility to add a personal object, in an asynchronous DNS lookup:
~~~
uint32_t FreeRTOS_gethostbyname_a( const char * pcHostName,
                                   FOnDNSEvent pCallback,
                                   void * pvSearchID,
                                   TickType_t uxTimeout );
~~~
When I use it, "pvSearchID" will point to an object in my source code.

I would not add an ID to `IPTCPSocket_t`, but to `FreeRTOS_Socket_t`, so that all sockets can have an ID.

What about :
~~~c
/* Setting an ID. */
CMySocket *socket = new CMySocket();
[...]

Issue #540, opened 2022-08-12
[Feature Request] STM32Hxx port: Duplicate MDIO Clock Config setup in HAL


Dweb2:
On the current version: ETH_MAC_MDIO_ClkConfig() https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/013b591124519a91fc81b52209474db6167fe49f/source/portable/NetworkInterface/STM32Hxx/stm32hxx_hal_eth.c#L2741 and HAL_ETH_SetMDIOClockRange() https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/013b591124519a91fc81b52209474db6167fe49f/source/portable/NetworkInterface/STM32Hxx/stm32hxx_hal_eth.c#L2180 are effectively identical functions which set the management clock. They get called from different contexts.

Maybe consider using just one function.

AniruddhaKanhere:
Hello @Dweb2, thank you for bringing this to our attention. We are looking into it and will get back to you soon

shubnil:
Hello @Dweb2, We missed this and will be looking into this now.

ActoryOu:
Hi @Dweb2,
Thanks for your contribution. We removed duplicate function HAL_ETH_SetMDIOClockRange by #711.
I'm going to close this issue.

Thanks.

PR #546, opened 2022-09-02
IPv4/Single: Add a SocketID to a socket

htibosch:
Description
-----------
In [issue #540](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/543), [Jabob Carver](https://github.com/karver8) is asking for the possibility to assign a unique ID to each socket.

We found that a good idea and this PR adds it, depending on the new macro `ipconfigUSE_SetSocketID`.

In many applications, sockets are embedded in an object, a class. The "SetSocketID" can be a pointer to the instance of that class.
These will be the new functions:
~~~c
BaseType_t xSocketSetSocketID( const Socket_t xSocket, void * pvSocketID );
void * pvSocketGetSocketID( const ConstSocket_t xSocket );
~~~

Test Steps
-----------
~~~c
/* Read the current value of the SocketID. */
void * id1 = pvSocketGetSocketID( xSocket );

/* Set value of the SocketID. */
int rc = xSocketSetSocketID( xSocket, ( void * ) 0x12345678 );

/* Read the current value of the SocketID. */
void * id2 = pvSocketGetSocketID( xSocket );

FreeRTOS_printf( ( "SocketSetSocketID %p -> %p rc = %d", id1, id2, ( int ) rc ) );
~~~

### DIFF
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -4780,32 +4780,27 @@ BaseType_t xSocketValid( const ConstSocket_t xSocket )
 #endif /* ipconfigUSE_TCP */
 /*-----------------------------------------------------------*/
 
-#if ( ipconfigUSE_SetSocketID != 0 )
-
 /**
  * @brief Set the value of the SocketID of a socket.
  * @param[in] xSocket: The socket whose ID should be set.
  * @param[in] pvSocketID: The new value for the SocketID.
  * @return Zero if the socket was valid, otherwise -EINVAL.
  */
-    BaseType_t xSocketSetSocketID( const Socket_t xSocket,
-                                   void * pvSocketID )
+BaseType_t xSocketSetSocketID( const Socket_t xSocket,
+                               void * pvSocketID )
+{
+    FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
+    BaseType_t xReturn = -pdFREERTOS_ERRNO_EINVAL;
+
+    if( xSocketValid( pxSocket ) )
     {
-        FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
-        BaseType_t xReturn = -pdFREERTOS_ERRNO_EINVAL;
-
-        if( xSocketValid( pxSocket ) )
-        {
-            xReturn = 0;
-            pxSocket->pvSocketID = pvSocketID;
-        }
-
-        return xReturn;
+        xReturn = 0;
+        pxSocket->pvSocketID = pvSocketID;
     }
-#endif /* ipconfigUSE_SetSocketID */
-/*-----------------------------------------------------------*/
 
-#if ( ipconfigUSE_SetSocketID != 0 )
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
 
 /**
  * @brief Retrieve the SocketID that is associated with a socket.
@@ -4814,19 +4809,18 @@ BaseType_t xSocketValid( const ConstSocket_t xSocket )
  *         the socket pointer is not valid or when the ID was not
  *         yet set.
  */
-    void * pvSocketGetSocketID( const ConstSocket_t xSocket )
+void * pvSocketGetSocketID( const ConstSocket_t xSocket )
+{
+    const FreeRTOS_Socket_t * pxSocket = ( const FreeRTOS_Socket_t * ) xSocket;
+    void * pvReturn = NULL;
+
+    if( xSocketValid( pxSocket ) )
     {
-        const FreeRTOS_Socket_t * pxSocket = ( const FreeRTOS_Socket_t * ) xSocket;
-        void * pvReturn = NULL;
-
-        if( xSocketValid( pxSocket ) )
-        {
-            pvReturn = pxSocket->pvSocketID;
-        }
-
-        return pvReturn;
+        pvReturn = pxSocket->pvSocketID;
     }
-#endif /* ipconfigUSE_SetSocketID */
+
+    return pvReturn;
+}
 /*-----------------------------------------------------------*/
 
 #if ( ( ipconfigHAS_PRINTF != 0 ) && ( ipconfigUSE_TCP == 1 ) )
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -4780,29 +4780,33 @@ BaseType_t xSocketValid( const ConstSocket_t xSocket )
 #endif /* ipconfigUSE_TCP */
 /*-----------------------------------------------------------*/
 
-#if( ipconfigUSE_SetSocketID != 0 )
+#if ( ipconfigUSE_SetSocketID != 0 )
+
 /**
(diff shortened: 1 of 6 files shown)


source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -1068,4 +1068,12 @@
     #define ipconfigPROCESS_CUSTOM_ETHERNET_FRAMES    0
 #endif
 
+/* This option adds the possibility to have a user-ID attached to a socket.
+ * The type of this ID is a void *.  Both UDP and TCP sockets have
+ * this ID. It has a default value of NULL.
+ */
+#ifndef ipconfigUSE_SetSocketID
```
AniruddhaKanhere, line 1075: Do we really need one more config?

source/FreeRTOS_Sockets.c
```diff
@@ -4780,6 +4780,55 @@ BaseType_t xSocketValid( const ConstSocket_t xSocket )
...
 
+#if ( ipconfigUSE_SetSocketID != 0 )
+
+/**
+ * @brief Set the value of the SocketID of a socket.
+ * @param[in] xSocket: The socket whose ID should be set.
+ * @param[in] pvSocketID: The new value for the SocketID.
+ * @return Zero if the socket was valid, otherwise -EINVAL.
+ */
+    BaseType_t xSocketSetSocketID( const Socket_t xSocket,
```
AniruddhaKanhere, line 4791: Seems like this is an API. Why not stick to the convention of starting the name with `FreeRTOS_`?
Something like `FreeRTOS_SetSocketID`?

source/FreeRTOS_Sockets.c
```diff
@@ -4780,6 +4780,55 @@ BaseType_t xSocketValid( const ConstSocket_t xSocket )
...
+#if ( ipconfigUSE_SetSocketID != 0 )
+
+/**
+ * @brief Retrieve the SocketID that is associated with a socket.
+ * @param[in] xSocket: The socket whose ID should be returned.
+ * @return The current value of pvSocketID, or NULL in case
+ *         the socket pointer is not valid or when the ID was not
+ *         yet set.
+ */
+    void * pvSocketGetSocketID( const ConstSocket_t xSocket )
```
AniruddhaKanhere, line 4817: Here too!

AniruddhaKanhere (changes requested):

source/include/FreeRTOSIPConfigDefaults.h
htibosch: Yes why not add it unconditionally. Linkers are smart enough to eliminate dead code. Only MISRA won't like it.
So I also vote to leave out the macro.

source/FreeRTOS_Sockets.c
htibosch: A disadvantage of `FreeRTOS_SetSocketID` would be that the name doesn't carry the type prefixes, like "x" and "pv".

source/include/FreeRTOSIPConfigDefaults.h
htibosch: > Yes why not add it unconditionally. Linkers are smart enough to eliminate dead code. Only MISRA won't like it. So I also vote to leave out the macro.

There is only one thing: all sockets will grown 4 of 8 bytes in size:
~~~
#if ( ipconfigUSE_SetSocketID != 0 )
    void * pvSocketID;
#endif
~~~
when the ID pointer is declared unconditionally.

AniruddhaKanhere: Yes, that is the cost of cleanliness in the code I suppose.
We already have far too many configuration options in my opinion. And the list just keeps growing.
I don't mind adding configs for things which affect the performance or certain characteristics (like timeouts etc) of the stack. But this seems like a bit of an overkill to me.
What do you think @htibosch?

htibosch:
About "[Removed the 'ipconfigUSE_SetSocketID' option](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/546/commits/1e5ac852614ff648e02aec056208dfac43c599be)":
We decided to add the Socket-ID functions unconditionally.

AniruddhaKanhere (approved):

test/unit-test/FreeRTOS_Sockets/FreeRTOS_Sockets_GenericAPI_utest.c
```diff
@@ -2903,6 +2903,89 @@ void test_FreeRTOS_maywrite_HappyPath( void )
...
+ */
+void test_xSocketSetSocketID_ValidSocket( void )
+{
+    BaseType_t xReturn;
+    FreeRTOS_Socket_t xSocket;
+    BaseType_t AnchorVariable;
+
+    memset( &xSocket, 0, sizeof( xSocket ) );
+
+    xReturn = xSocketSetSocketID( &xSocket, &AnchorVariable );
```
shubnil, line 2941: What will be the value of the anchor variable here.

source/FreeRTOS_Sockets.c
```diff
@@ -4797,6 +4797,49 @@ BaseType_t xSocketValid( const ConstSocket_t xSocket )
...
+ * @param[in] xSocket: The socket whose ID should be set.
+ * @param[in] pvSocketID: The new value for the SocketID.
+ * @return Zero if the socket was valid, otherwise -EINVAL.
+ */
+BaseType_t xSocketSetSocketID( const Socket_t xSocket,
+                               void * pvSocketID )
+{
+    FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
+    BaseType_t xReturn = -pdFREERTOS_ERRNO_EINVAL;
+
```
shubnil, line 4811: I have a query here. How are we getting the value of the socket_id. I mean, is it left to the test application to assign a value or will the TCP stack maintain a sequence.

shubnil (commented):
Hi, Overall I wanted to understand more on how we are generating the socket-id. Will this be assigned by the test application or will the TCP stack has a way to maintain and assign a socket id.

htibosch:
@shubnil: I will be one of the happy user of this new feature!

My applications encapsulates +TCP sockets into an object. This can be a struct or even a C++ class.

And as you know, the FreeRTOS+TCP library offers several different callbacks ( application hooks ) that are triggered on certain events: on connect, on read, on write etc.
Before this PR, I had to maintain a list of sockets, in order to be able to translate a socket pointer to its encapsulating object.

After this PR, that administration isn't necessary anymore.

The FreeRTOS+TCP library doesn't care about this pointer, it is just stores and reveals it.

> How are we getting the value of the socket_id. I mean, is it left to the test application to assign a value or will the TCP stack maintain a sequence.

As you will understand by now, the application builder will manage these pointers.

>> xReturn = xSocketSetSocketID( &xSocket, &AnchorVariable );
> What will be the value of the anchor variable here.

The new feature is storing a pointer. The value of the anchor value will probably an index or a pointer to an instance of a C++ class.

Does that make sense?

shubnil (approved):
Thanks, This makes sense.

PR #547, opened 2022-09-06
Set ipBUFFER_PADDING to 14 bytes by default on 64 bit targets

paulbartell:
Set ipBUFFER_PADDING to 14 bytes by default on 64 bit targets

Description
-----------
If ipconfigBUFFER_PADDING is not defined in FreeRTOSIPConfig.h, builds for 64 bit targets fail. Add a default value of 14 bytes ( 12U + ipconfigPACKET_FILLER_SIZE ) for ipBUFFER_PADDING to satisfy the alignment check in FreeRTOS_IP_Utils.c.

### DIFF
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -385,46 +385,52 @@ void prvProcessNetworkDownEvent( void )
  */
 void vPreCheckConfigs( void )
 {
+    size_t uxSize;
+
     /* This function should only be called once. */
     configASSERT( xIPIsNetworkTaskReady() == pdFALSE );
     configASSERT( xNetworkEventQueue == NULL );
     configASSERT( FreeRTOS_GetIPTaskHandle() == NULL );
 
     #if ( configASSERT_DEFINED == 1 )
-        {
-            volatile size_t uxSize = sizeof( uintptr_t );
+        #if ( UINTPTR_MAX > 0xFFFFFFFF )
 
-            if( uxSize == 8U )
-            {
-                /* This is a 64-bit platform, make sure there is enough space in
-                 * pucEthernetBuffer to store a pointer. */
-                configASSERT( ipconfigBUFFER_PADDING >= 14 );
-                /* But it must have this strange alignment: */
-                configASSERT( ( ( ( ipconfigBUFFER_PADDING ) + 2 ) % 4 ) == 0 );
-            }
+            /*
+             * This is a 64-bit platform, make sure there is enough space in
+             * pucEthernetBuffer to store a pointer.
+             */
+            configASSERT( ipBUFFER_PADDING >= 14U );
+        #else
+            /* This is a 32-bit platform. */
+            configASSERT( ipBUFFER_PADDING >= 10U );
+        #endif /* UINTPTR_MAX > 0xFFFFFFFF */
 
-            /* LCOV_EXCL_BR_START */
-            uxSize = ipconfigNETWORK_MTU;
-            /* Check if MTU is big enough. */
-            configASSERT( uxSize >= ( ipSIZE_OF_IPv4_HEADER + ipSIZE_OF_TCP_HEADER + ipconfigTCP_MSS ) );
+        /* And it must have this strange alignment: */
+        configASSERT( ( ( ( ipBUFFER_PADDING ) + 2U ) % 4U ) == 0 );
 
-            uxSize = sizeof( EthernetHeader_t );
-            /* Check structure packing is correct. */
-            configASSERT( uxSize == ipEXPECTED_EthernetHeader_t_SIZE );
+        /* LCOV_EXCL_BR_START */
+        uxSize = ipconfigNETWORK_MTU;
+        /* Check if MTU is big enough. */
+        configASSERT( uxSize >= ( ipSIZE_OF_IPv4_HEADER + ipSIZE_OF_TCP_HEADER + ipconfigTCP_MSS ) );
 
-            uxSize = sizeof( ARPHeader_t );
-            configASSERT( uxSize == ipEXPECTED_ARPHeader_t_SIZE );
+        uxSize = sizeof( EthernetHeader_t );
+        /* Check structure packing is correct. */
+        configASSERT( uxSize == ipEXPECTED_EthernetHeader_t_SIZE );
 
-            uxSize = sizeof( IPHeader_t );
-            configASSERT( uxSize == ipEXPECTED_IPHeader_t_SIZE );
+        uxSize = sizeof( ARPHeader_t );
+        configASSERT( uxSize == ipEXPECTED_ARPHeader_t_SIZE );
 
-            uxSize = sizeof( ICMPHeader_t );
-            configASSERT( uxSize == ipEXPECTED_ICMPHeader_t_SIZE );
+        uxSize = sizeof( IPHeader_t );
+        configASSERT( uxSize == ipEXPECTED_IPHeader_t_SIZE );
 
-            uxSize = sizeof( UDPHeader_t );
-            configASSERT( uxSize == ipEXPECTED_UDPHeader_t_SIZE );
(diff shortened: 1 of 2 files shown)


source/include/FreeRTOS_IP.h
```diff
@@ -98,22 +98,30 @@ extern uint32_t ulApplicationGetNextSequenceNumber( uint32_t ulSourceAddress,
...
- *  uchar_8 dest[6];
- *  uchar16_t type;
- *  << IP-header >>     // word-aligned
+ * +---------+-----------+---------+
+ * | Offset  | Alignment | Length  |
+ * | 32 | 64 | 32  | 64  | 32 | 64 |
+ * |----|----|-----|-----|----|----|
+ * | 0  | 0  | 4   | 8   | 4  | 8  | uchar_8 * pointer;
+ * | 4  | 8  | 4   | 8   | 2  | 6  | uchar_8   filler[];   // ipconfigPACKET_FILLER_SIZE
+ * | 6  | 14 | 4+2 | 8+6 | 7  | 7  | uchar_8   preamble[7] // Beginning of Ethernet header
```
paulbartell, line 110: I'm not sure if preamble should be word aligned or the ip-header.

AniruddhaKanhere: I think the IP header should be word aligned. The ethernet header is anyways described as an array of uint8_t. It's alignment should not matter I think.

@htibosch care to comment/clarify?

htibosch: @paulbartell wrote:

> I'm not sure if the ethernet preamble should be word aligned or the ip-header.

The Ethernet preamble is used by the hardware layer to establish the bit-speed of the connection. The first byte that is sent to the application (through MDIO or SPI) is the target hardware address. DMA understands the +2 offset from an aligned memory address.

Thank you Paul for the complete overview that compares 32- with 64-bits alignment. I made some small changes:

~~~
/*
 * +---------+-----------+---------+
 * | Offset  | Alignment | Length  |
 * | 32 | 64 | 32  | 64  | 32 | 64 |
 * |----|----|-----|-----|----|----|
 * | 0  | 0  | 4   | 8   | 4  | 8  | uchar_8 * pointer;     // Points to the 'NetworkBufferDescriptor_t'.
 * | 4  | 8  | 4   | 8   | 2  | 2  | uchar_8   filler[];    // To give the +2 byte offset.
 * | 6  | 10 | 4+2 | 8+2 | 6  | 6  | uchar_8   dest_mac[6]; // Destination address.
 * | 12 | 16 | 4   | 8   | 6  | 6  | uchar_8   src_mac[6];  // Source address.
 * | 18 | 22 | 4+2 | 8+6 | 2  | 2  | uchar16_t ethertype;
 * | 20 | 24 | 4   | 8   | ~  | ~  | << IP-header >>        // word-aligned, either 4 or 8 bytes.
 *  uint8_t ucVersionHeaderLength;
 *  etc
 */
~~~

I removed 'preamble' and 'delimiter' from the list.

@AniruddhaKanhere wrote:

> I think the IP header should be word aligned. The ethernet header is anyways described as an array of uint8_t. It's alignment should not matter I think.

That is right, we want the IP-header be aligned at least at 32-bits. There are no 64-bit fields in the library.
If a field is declared as `uint32_t`, it is assumed that the field is 32-bit aligned within the network buffer.
Otherwise a character array is used, as in the ARP header:
~~~
MACAddress_t xSenderHardwareAddress;  /**< Address of the sender     8 +  6 = 14 */
uint8_t ucSenderProtocolAddress[ 4 ]; /**< IP-address of sender     14 +  4 = 18  */
MACAddress_t xTargetHardwareAddress;  /**< Address of the receiver  18 +  6 = 24  */
~~~

source/FreeRTOS_IP_Utils.c
```diff
@@ -398,9 +398,9 @@ void vPreCheckConfigs( void )
             {
                 /* This is a 64-bit platform, make sure there is enough space in
                  * pucEthernetBuffer to store a pointer. */
-                configASSERT( ipconfigBUFFER_PADDING >= 14 );
+                configASSERT( ipBUFFER_PADDING >= 14 );
```
htibosch, line 401: To be complete, we might as well check `ipBUFFER_PADDING` in case of a 32-bit platform:
~~~
    #if( UINTPTR_MAX > 0xFFFFFFFF )
    {
        configASSERT( ipBUFFER_PADDING >= 14 );
    }
    #else
    {
        /* This is a 32-bit platform. */
        configASSERT( ipBUFFER_PADDING >= 10 );
    }
    #endif
    /* And it must have this strange alignment: */
    configASSERT( ( ( ( ipBUFFER_PADDING ) + 2 ) % 4 ) == 0 );
~~~

and use the static check for word size: `#if( UINTPTR_MAX > 0xFFFFFFFF )`

htibosch (commented):
Thank you Paul for this useful PR. It will give less confusion for the users.

I think that the table that you added to `FreeRTOS_IP.h` needs some changes, as I proposed.

I compiled and tested both a 32-bit ( STM32F7 ) and a 64-bit application ( UltraScale ).

source/FreeRTOS_IP_Utils.c
paulbartell: @htibosch I'm still a bit confused. How are the minimums of 10 and 14 bytes calculated?

htibosch: > I'm still a bit confused. How are the minimums of 10 and 14 bytes calculated?

That is a difficult story, have a look [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/dd42247cc0fd3fdef51b0fa902bb22923a525d83/source/include/FreeRTOS_IP_Private.h#L452).

`ipIP_TYPE_OFFSET` is the offset of 6 bytes before the actual packet where the IP version is written.
~~~
address -10 pointer to the owning NetworkBuffer
address  -9 pointer
address  -8 pointer
address  -7 pointer
address  -6 IP-type, either 0x40 or 0x60
address  -5 n.a.
address  -4 n.a.
address  -3 n.a.
address  -2 n.a.
address  -1 n.a.
address   0 Ethernet header
~~~

The IP-type is read by `pxUDPPayloadBuffer_to_NetworkBuffer()` :

The problem is that the library passes pointers to the UDP payload (used for zero-copy `recvfrom()`/`sendto()`).
At some point, we have a payload and we want to know the start of the packet. Therefore we need to know if it is an IPv4 or an IPv6 packet. The payload doesn't tell anything about this, it is just a user message.

For an IPv6 packet, the type is stored in the IP-header in the field `ucVersionTrafficClass`, which has the value 0x60. This field is located 48 bytes before the payload.
For an IPv4 packet, at a position of 48 bytes before the payload, the value `0x40` will be stored. That is a position **before** the Ethernet header, marked here above as "IP-type".

So that means that `ipBUFFER_PADDING` must be at least 10 bytes long. And as pointers on a 64-bit platform are 8 bytes long, the minimum length of `ipBUFFER_PADDING` is 14 or longer.

htibosch: Maybe a better way to visualise the message buffer header is as follows:
~~~c
#define ipTYPE_IPv4         ( 0x40U )
#define ipTYPE_IPv6         ( 0x60U )

struct NetworkBuffer {
    NetworkBufferDescriptor_t * pxOwner;
    uint8_t ucIPType;  /* Either 'ipTYPE_IPv4' or 'ipTYPE_IPv6' */
    uint8_t ucFiller[ 5 ];
    uint8_t ucEthernetHeader[ 14 ];
    uint8_t ucIPHeader[ 20 ];  /* Either 20 or 40 */
};
~~~
The field `ucIPHeader` must be 4-byte aligned.

moninom1:
Closing the PR as it is now taken care by https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1061

PR #551, opened 2022-09-09
IPv4/single: Call DNS callback routines while scheduler is not suspended

htibosch:
Description
-----------
This PR is quite similar to my earlier #250., which was merged with [ipv6_multi](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi).

When using `FreeRTOS_gethostbyname_a()`, an IP address will be looked up "asynchronously". The lookup will be started from the user task, and it will be finished by the IP-task.

When the lookup is ready, or after a time-out, the IP-task will call the DNS application hook, as indicated by the user:
~~~c
uint32_t FreeRTOS_gethostbyname_a(
                const char * pcHostName, // The name of the host to be looked up
                FOnDNSEvent pCallback,   // Application hook, called when ready
                void * pvSearchID,       // An ID that will be passed when calling the hook
                TickType_t uxTimeout );  // A time-out in clock ticks
~~~

While iterating through the list of outstanding DNS lookups, it is necessary to suspend the scheduler: it is possible that `List_t xCallbackList` is accessed from 2 tasks.

Until now, the function `pCallback` was called while the scheduler is suspended. That is not necessary, and it can be disturbing for user.

This PR will change that behaviour: the scheduler will be running while the application hook is called.

Test Steps
-----------
I tested by using `FreeRTOS_gethostbyname_a()` both for a successful lookup, and for a lookup that timed out.


source/FreeRTOS_DNS_Callback.c
```diff
@@ -157,34 +174,56 @@
...
-                 pxIterator != xEnd; )
+            vTaskSuspendAll();
             {
-                DNSCallback_t * pxCallback = ( ( DNSCallback_t * ) listGET_LIST_ITEM_OWNER( pxIterator ) );
-                /* Move to the next item because we might remove this item */
-                pxIterator = ( const ListItem_t * ) listGET_NEXT( pxIterator );
+                pxKeepCallback = NULL;
 
-                if( ( pvSearchID != NULL ) && ( pvSearchID == pxCallback->pvSearchID ) )
+                for( pxIterator = ( const ListItem_t * ) listGET_NEXT( xEnd );
```
AniruddhaKanhere, line 185: Why not remember the `pxIterator` value too?

This would allow us to continue from the last place where we were before we hit the `break` statement; making the code more fast and optimized. (think when the entries we want to remove are around the end of the list)

It is as simple as moving the initialization of `pxIterator` outside all the loops (just before the do-while loop begins).

AniruddhaKanhere (changes requested):

source/FreeRTOS_DNS_Callback.c
htibosch: I have thought of doing so, but the problem is that the `List_t` is shared with other tasks.
Possibly, we're OK keeping `pxIterator` and continuing were we were. That looks a lot better.
A user task may call `vDNSSetCallBack()`, which will add a new item at the end of the list.
The only problem I see is this function:
~~~c
        void FreeRTOS_gethostbyname_cancel( void * pvSearchID )
        {
            /* _HT_ Should better become a new API call to have the IP-task remove the callback */
            vDNSCheckCallBack( pvSearchID );
        }
~~~
My initials `_HT_` say that actually the IP-task should remove the callback.
I did so:
~~~c
    /* Let the IP-task remove the callback from the list. */
    IPStackEvent_t xRemoveEvent = { eCancelDNSCallbackEvent, pvSearchID };

    /* Send the network task the appropriate event. */
    xSendEventStructToIPTask( &( xRemoveEvent ), portMAX_DELAY );
~~~
and now it is safe to resume/suspend:
~~~c
    ( void ) xTaskResumeAll();
    {
        pCallbackFunction( pcName, pxCallback->pvSearchID, ulIPAddress );
    }
    vTaskSuspendAll();
~~~

AniruddhaKanhere (approved):

source/FreeRTOS_DNS_Callback.c
```diff
@@ -72,12 +72,17 @@
...
-                                                   ulIPAddress );
+                    FOnDNSEvent pCallbackFunction = pxCallback->pCallbackFunction;
+                    void * pvSearchID = pxCallback->pvSearchID;
                     ( void ) uxListRemove( &pxCallback->xListItem );
                     vPortFree( pxCallback );
 
+                    ( void ) xTaskResumeAll();
+                    {
+                        pCallbackFunction( pcName, pvSearchID, ulIPAddress );
+                    }
```
kstribrnAmzn, line 83: Naive question: Why does this call need to be in its own scope?

kstribrnAmzn (approved):

ActoryOu:
This function is merged in IPv6 dev branch and we're going to merge IPv6 back to main in the near future.
Close this PR in advance.

Issue #556, opened 2022-09-23
[Feature Request] Support for CMake


phelter:
**Is your feature request related to a problem? Please describe.**
As an integrator of FreeRTOS-Plus-TCP would like to support compilation via CMake.

**Describe the solution you'd like**

I would like to be able to integrate FreeRTOS-Plus-TCP via CMake to support a simpler build environment for native and cross-compilation.

I typically build my environments with an opt-out approach to all warnings and issues linters/compilers provide while compiling/building.  The primary compilers I use are GNU and Clang (sometimes MSVC).

I typically pull in repositories of this kind using (FetchContent)[https://cmake.org/cmake/help/latest/module/FetchContent.html?highlight=fetchcontent] functionality as it provides one key benefit over `git submodules` which is to allow a top level project to dictate the version being pulled, yet others `FetchContent` repos may request different versions.  Allowing the integrator of this project to decide on the version being pulled.  It also allows for a simple migration if/when a contributor decides to fork a repo (if required).

**Describe alternatives you've considered**
There are other build environments out there that support cross compilation (eg. meson, bazel); however, CMake is the most wide spread and widely used.  It is also the one that has support in the `FreeRTOS-Kernel` project as of v10.5.0.  Supporting other build environments in the future is not hindered by supporting CMake now.

**Additional context**
I will provide an example build of the existing code that I lay over the current `FreeRTOS-Plus-TCP` that allows me to compile for both GCC and CLang and provides hooks to compile with other compilers. 

With the addition of the compiler warning `opt-out` approach it will also identify documentation and other potential fixes that you may want to clean-up within the code, now that those compilers are defined with the strictest of measures.

phelter:
Here is the cmake builds.

Note: Outstanding items:
* Some of the build Network Interfaces for various platforms are not complete - didn't want to infer anything there.  Have not tested on all of them.
* Have only tested on Clang 13/GNU 7.5 - with POSIX.


I'm adding a patch file here since I don't have access to create a branch.
[cmake_build.patch.txt](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/9637607/cmake_build.patch.txt)

PR #557, opened 2022-09-25
#556 Initial Cmake Module definition.

phelter:
Support for CMake and compilation with strictest definitions for GNU/Clang.

Description
-----------
Added support for CMake compile of the source code.
- Excludes some Ports at the moment - due to lack of testing infrastructure.
- Excludes integrating/changing unit-tests and compiling:
  - `test/Coverity`
  - `test/cbmc`
  - `test/unit-tests`
  - Reason is these are projects of their own and do not know the interworkings of these - could integrate later.
- Includes build of the Various Build-configurations for testing - including a CUSTOM for externally defined configuration.
  - `DEFAULT_CONF`, `ENABLE_ALL`, `DISABLE_ALL`

Test Steps
-----------
- Updated CI - but tested locally with:
  - Ubuntu 18.04 with Clang 13 - ran all 3 build configuration tests
  - Ubuntu 18.04 with GNU 7.5 - ran all 3 build configuration tests

Related Issue
-----------
- #556 - Support for CMake.
- Note there are several Clang warnings that should be addressed in various files.  These will be fixed in subsequent PR's.


AniruddhaKanhere:
@phelter, thank you for raising the issue and creating this PR!
We are looking into it and will get back to you with changes (if any).

Thanks for your patience!

phelter:
@AniruddhaKanhere - Thanks for reviewing.  I hope it is along the lines of how you want `FreeRTOS-Plus-TCP` to be integrated into projects.  Feel free to ask why certain things are done a certain way as I know CMake isn't well used in the embedded firmware realm too much and is an unknown beast.

It would be great if some of the `ports` could be done by others following the same CMakeLists.txt format/guidelines as the other ports and then I can review those as needed.

htibosch:
> CMake isn't well used in the embedded firmware
> realm too much and is an unknown beast

That is probably why I know nothing about how to use cmake. I'm a GNU make user.

Like I just commented to your [PR #35](https://github.com/FreeRTOS/Lab-Project-FreeRTOS-FAT/pull/35), would you mind to write a few lines that explain how things can be tested?

And can you also update this PR with the main branch?

Also it needs formatting: you can start that to write a 1-line comment with the contents:
`/bot run uncrustify`
that should start reformatting. Don't forget to pull to your local branch after doing that.
Thanks

phelter:
@htibosch 

- For how things are tested:
Please see the changes to the `ci.yml` file here: https://github.com/phelter/FreeRTOS-Plus-TCP/blob/cf5aa7ff657ba650a6e62f30aa45a58a60173397/.github/workflows/ci.yml#L110

The cmake file now not only compiles just for these tests, but also provides the static libraries for others to include in their builds.

The user can also use these test builds to confirm that their specific `FreeRTOSConfig.h` and `FreeRTOSIPConfig.h` are also valid for the current versions of the Kernel and Plus-TCP components.

If you would rather just look at the results the last runs are here:
https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/3200788444/jobs/5268962000

- And can you also update this PR with the main branch?
Done

- Formatting
Done as mentioned. `/bot run uncrustify`

phelter:
@AniruddhaKanhere & @htibosch

I believe I've provided all of the requested info about this PR.  Would it be possible to add a few eyes on this PR in the near future?  

Thanks

README.md
```diff
@@ -31,7 +31,40 @@ Once python is downloaded and installed, you can verify the version from your te
 To run the script, you should switch to the FreeRTOS-Plus-TCP directory that was created using the [Cloning this repository](#cloning-this-repository) step above.
 And then run  `python <Path/to/the/script>/GenerateOriginalFiles.py`.
 
-## Cloning this repository
+## To consume FreeRTOS+FAT
```
archigup, line 34:
```suggestion
## To consume FreeRTOS+TCP
```

.github/workflows/ci.yml
```diff
@@ -111,23 +111,24 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Update submodules
-        run: git submodule update --init --checkout
+      # Using CMake FetchContent to pull FreeRTOS-Kernel now.
+      # - name: Update submodules
+      #   run: git submodule update --init --checkout
```
archigup, line 116:
```suggestion
```

tools/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+    tcp_utilities/include/tcp_dump_packets.h
+    tcp_utilities/include/tcp_mem_stats.h
+    tcp_utilities/include/tcp_netstat.h
+
+    tcp_utilities/tcp_dump_packets.c
+    tcp_utilities/tcp_mem_stats.c
+    tcp_utilities/tcp_netstat.c
+)
+
+# Note: Have to make system due to compiler warnings in header files.
```
archigup, line 14: Which warnings are these? Can the includes by fixed instead?

CMakeLists.txt
```diff
@@ -0,0 +1,279 @@
...
+            set(FREERTOS_PORT GCC_POSIX CACHE STRING "" FORCE)
+        elseif(MINGW)
+            set(FREERTOS_PORT MSVC_MINGW CACHE STRING "" FORCE)
+        endif()
+    endif()
+endif()
+
+########################################################################
+# Requirements
+set(CMAKE_C_STANDARD 99) # Note FreeRTOS-Kernel uses C99 constructs.
```
archigup, line 166: The kernel should compile with C89

source/portable/CMakeLists.txt
```diff
@@ -0,0 +1,223 @@
...
+if( FREERTOS_PLUS_TCP_NETWORK_IF STREQUAL "POSIX")
+    find_package(PCAP REQUIRED)
+    set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+    find_package(Threads)
+endif()
+
+target_include_directories( freertos_plus_tcp_network_if
+  PRIVATE
+    NetworkInterface
+    # $<$<STREQUAL:${FREERTOS_PLUS_TCP_NETWORK_IF},ATSAM4E:${CMAKE_CURRENT_SOURCE_DIR}/NetworkInterface/ATSAM4E> #TODO
```
archigup, line 177: Could you remove the commented out stuff?

tools/CMakeLists.txt
phelter: Yes - the plan here is to add the necessary CMake compilation with minimal changes to the C files themselves and Identify all of the compilation warnings first.
One of them that was identified is `-Wpadded (clang)` - this means that there is additional padding on particular structs etc.  which in most cases is okay, but points to possible improvements for better layout in memory and/or alignment.

A follow on PR for actually removing these warnings will happen after this PR is merged.  I didn't want to have too much in each of the PR's.  What is currently there has been used/consumed for the longest time so assuming none of these are major issues.

CMakeLists.txt
phelter: @archigup - IMO, that ship has sailed.  Unless there is a significant amount of work to remove all of the code that is using C99 constructs and operation is done in both the FreeRTOS-Kernel and FreeRTOS-Plus-TCP this is not viable at the moment.

One example is `declaration after execution` - i.e. declaring variables midway in a procedure.  There are many cases.

Is there a particular compiler that doesn't support C99 at this point in time? - For those users who wish to continue compiling with a compiler with no C99 support, I believe if they are not likely to update their compiler, what are the chances they will consume a new version of FreeRTOS-Plus-TCP as well?  We are talking about 20+ yr old compilers here.

phelter:
@archigup , @AniruddhaKanhere and @htibosch .

I've updated the PR and have fixed all of the requested issues identified.

Outstanding item: `C99` vs` C8`9 - please confirm you do or don't want this change at this moment. Note that the current FreeRTOS-Plus-TCP and FreeRTOS-Kernel both use `C99 ISO` constructs and would have to be re-worked to support `C89` without violating the ISO.

AniruddhaKanhere:
Hello @phelter,

Thank you for making the changes that we suggested!

We are discussing the changes that you have proposed here (and in FreeRTOS-Kernel repository). We don't want to rush accepting your PR without a discussion just to later overwrite your changes. To that end, we are trying to figure out a better way to make the changes that you have proposed.

It is taking time - as all good things do :)

I shall get back to you in a bit once we have figured out the best way to make certain changes and accept your PR (and also the one in FreeRTOS-kernel repository).

@paulbartell had some suggestions which he will add to the PR comments.

Thanks for your patience @phelter and for your efforts with your PRs,
Aniruddha

CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+########################################################################
+# Project Details
+project(FreeRTOS-Plus-TCP
+        VERSION 3.1.0
+        DESCRIPTION "FreeRTOS TCP/UDP Network Layer"
+        HOMEPAGE_URL https://freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/index.html
+        LANGUAGES C)
+
+# Do not allow in-source build.
+if( ${PROJECT_SOURCE_DIR} STREQUAL ${PROJECT_BINARY_DIR} )
```
paulbartell, line 17: why not?

CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+
+# Configuration
+# Override these at project level with:
+#   Optional: set(FREERTOS_PLUS_TCP_BUFFER_ALLOCATION "1" CACHE STRING "" FORCE)
+#   Optional: set(FREERTOS_PLUS_TCP_COMPILER "" CACHE STRING "" FORCE)
+#   Required: set(FREERTOS_PLUS_TCP_NETWORK_IF "POSIX" CACHE STRING "" FORCE)
+
+# Select the appropriate buffer allocaiton method.
+# See: https://freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/Embedded_Ethernet_Buffer_Management.html
+if (NOT FREERTOS_PLUS_TCP_BUFFER_ALLOCATION)
```
paulbartell, line 32: Could each buffer allocation be exposed as a separate static lib target?

CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+#   Keil                       | Yes      | ARMCC
+#   MSVC                       | Yes      | MSVC # Note only for MinGW
+#   Renesas                    | No       | ?TBD?
+# Will always a attempt to detect and if detectable double checks that the compiler is set correctly.
+set(FREERTOS_PLUS_TCP_COMPILER "" CACHE STRING "FreeRTOS Plus TCP Compiler Selection")
+
+
+# Select the appropriate network interface
+# This will fail the CMake preparation step if not set to one of those values.
+set(FREERTOS_PLUS_TCP_NETWORK_IF "" CACHE STRING "FreeRTOS Plus TCP Network Interface selection")
```
paulbartell, line 54: Could each of these be a separate static lib target?

CMakeLists.txt
paulbartell: @phelter : We have had a number of internal discussions on this topic and our current policy is ANSI C89 / ISO C90 compliance with a limited number of exceptions (stdbool and stdint mainly). I don't expect this policy to change in the near future until we can be 100% certain there are no compatibility breaks.

If there are specific C89/C90 compliance problems, they should be either:
1) Considered a bug <OR>
2) Added to the list of exceptions for that particular repository in the repository's README.md and masked for common compiler in the cmake or source files.

Note, this mainly applies to library code. I'ts a bit more flexible with demos or other examples that target a particular platform. In the case of FreeRTOS+TCP, the portable directory is excluded from this policy.

tools/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+add_library( freertos_plus_tcp_tools STATIC )
+
+target_sources( freertos_plus_tcp_tools
```
paulbartell, line 3: Could this been included in the default FreeRTOS+TCP target?

CMakeLists.txt
phelter: Ok - thanks - will assume all use of C99 as a bug then and identify them in the warnings for now - and fix them in a follow-on commit.

CMakeLists.txt
phelter: Since it is in other `CMakeList.txt` files in this current directory, I added the same restriction here.

See other examples: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/search?q=if%28+%24%7BPROJECT_SOURCE_DIR%7D+STREQUAL+%24%7BPROJECT_BINARY_DIR%7D+%29

CMakeLists.txt
phelter: I treated this similar to how the HEAP selection is performed in FreeRTOS-Kernel.  

Option 1: Each BufferAllocation be exposed as separate static lib target:
-  +ves Always compiled and available.
-  -ves Compiling code that the user does not use.  Additional static libraries generated.
Option 2: port specific - it is not port specific feature - a port may limit the BufferAllocation used, but they are independent configurations.

Will leave as is, unless a specific change is requested.

CMakeLists.txt
phelter: Yes - it makes more sense to separate these out as independent `CMakeLists.txt` in each of the respective portable/* directories, but I followed the way the FreeRTOS-Kernel does it instead.

That way the only selection (single case/if-else selection) is in the `portable/CMakeLists.txt` - as to which `add_subdirectory()` to use.

tools/CMakeLists.txt
phelter: Yes that `could` be done - if that is the case then the tools/* code `should` be placed in the same directory as the freertos_plus_tcp code.

Typically a separate directory means a separate library.  Having multiple libraries here is not a big deal.  The up-side to leaving as-is in the tcp_utilities directory allows one to take this and combine this static library in their testing area so they have the same code monitoring the counts inside and outside the `Unit under test`.  Otherwise one would need to expose a private set of API's in both.

Assuming leaving here as-is - for reasons provided.

CMakeLists.txt
phelter: FYI:

I've fixed all the C90 - ISO violations in the code and have separated out the cmakes so they are per Network IF port now instead of all one big lump with multiple `case` statements in the one.

I do not intend to test all of the other ports but have tested the linux and Zynq ports and have confirmed they work.

CMakeLists.txt
phelter: Yes - independent CMakeLists.txt now - which makes adding and maintaining the ports much simpler as they are now contained with their own CMakeLists.txt file.

CMakeLists.txt
phelter: @paulbartell - One other thing to note.  If you had example builds of each and every compiler that FreeRTOS-Plus-TCP (and FreeRTOS-Kernel) intends to support, then you could use those as a deciding factor as to whether you could increase to C99 or higher (eg C11).  The example builds could be easily provided in a CI/CD build by:
* Creating a docker image (eg. Linux Ubuntu) per compiler supported - along with installed BSP of each of the boards that is identified as supported.
* Provide a toolchain file per compiler
* Provide a toolchain file per board (optional) that includes the specific compile options for that board and includes the compiler toolchain (Dont-Repeat-Yourself DRY).  This is optional because if you only want to confirm the compiler works then you really only need one example and all others should be the same with the exception of the portable code that is compiled.  To reduce the amount of time checking all the different boards you could compile everything with one board BSP and then just compile the `freertos_plus_tcp_network_if` target for each of the other boards supported with that same compiler.
* Add the necessary build as a cross-compilation per compiler and board in the CI/CD flow - probably only necessary for release builds.

Then there is a guarantee that if you change something like the Required Compiler ISO - you know it won't break someone's build.

phelter:
@paulbartell , @AniruddhaKanhere , @archigup :

Any progress on this? - would like to not rely on my forks of FreeRTOS/* for projects that I am working on.

source/portable/NetworkInterface/STM32Fxx/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+add_library( freertos_plus_tcp_network_if STATIC )
+
+target_sources( freertos_plus_tcp_network_if
+  PRIVATE
+    NetworkInterface.c
+    # stm32f2xx_hal_eth.h
+    # stm32f4xx_hal_eth.h
+    stm32f7xx_hal_eth.h  #TODO: Only one of these ?
+    stm32fxx_hal_eth.c
+    stm32fxx_hal_eth.h>
```
AniruddhaKanhere, line 15:
```suggestion
    stm32fxx_hal_eth.h
```

AniruddhaKanhere (changes requested):
Just a small NIT pick

tools/CMakeLists.txt
phelter: If you want to see the warnings/changes - they are here:

* https://github.com/phelter/FreeRTOS-Plus-TCP/compare/feature/cmake-support...feature/fixing-clang-gnu-compiler-warnings

The big sets are:
* misuse of doxygen comments or out of date parameter names 
* uninitialized variables at the beginning - which have cases where they may be used before loaded/set - Undefined Behaviour.
* invalid use of unsigned vs. signed - missing casts.

Or you can compile as mentioned with clang-13 or 14 or above and check the warning lists yourselves.

phelter:
@paulbartell , @AniruddhaKanhere , @archigup :

Just a FYI,

This is the last post and update you will get from me for these changes.

I am no longer going to maintain this.  Please feel free to add or abandon these changes.

I am moving my efforts on the opensource community away from FreeRTOS as the turnaround times for changes and PR's is in excess of 5 months and it looks as though there is little effort in maintaining this code base.

I am assuming Amazon is winding down it's support for FreeRTOS since there has been a shift/migration to [Zephyr](https://zephyrproject.org/) .

AniruddhaKanhere:
Hello @phelter,

> I am moving my efforts on the opensource community away from FreeRTOS as the turnaround times for changes and PR's is in excess of 5 months

I am sorry to hear that and sincerely apologize for the delay. We did not use `CMake` in +TCP and the FreeRTOS-Kernel before and had many discussions about which way is the best to avoid rework and make the repository easier to use for our users. This led to all the discussions and comments we had on your PR.
Regardless, I share your view that the turnaround time on this PR was quite large. To fix that and to ensure that this is not repeated with any other PR, we are working on a mechanism which should help us focus on pending PRs amongst other things.

> and it looks as though there is little effort in maintaining this code base.

This might not be completely accurate. We were focused on adding and refactoring the changes in the IPv6 branch. IPv6 functionality is quite sought after and thus we were prioritizing that.

> I am assuming Amazon is winding down it's support for FreeRTOS

This is also not accurate. If anything, we are adding more and more to the FreeRTOS ecosystem.

Sincerely,
Aniruddha

AniruddhaKanhere (approved):
Thank you for this PR @phelter!

I am approving this PR. Once we have a second approval, we can merge it mainline

Thanks

phelter:
Thanks @AniruddhaKanhere for the review.

Please note I believe this repo has similar issues to the ones mentioned in this issue: 
https://github.com/aws/ota-for-aws-iot-embedded-sdk/issues/465

And this was mentioned in this issue here - https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/570#issuecomment-1312560393

It looks as though there is a fundamental misunderstanding in the requirements for making a `module and it's API` task/thread safe and robust in the various aws opensource projects.  I'd suggest adding a similar issue to this repo as well regarding a `State Access Audit`.

And to ensure these type of issues don't pop up with future modifications, consider incorporating both Address and Thread Sanitizer runtime unit tests. Where the unit tests stimulates the module as a `black/gray-box` (with all of the appropriate various threads and timers running in the typical threading organization).  These type of unit-tests can then also act as confirmation that the examples you provide are operational as they are more like example code rather than low-level unit-tests.

Why do I believe this is an issue in this repo?

I've been seeing various issues with FreeRTOS-Plus-TCP hanging when using the FTP server functionality.  It passes when it is compiled Debug mode, but when compiled for Release mode, the pings stop processing.  In my experience, these types of bugs are typically seen when there are issues with State Access between threads/tasks (or between ISR and thread contexts).  There were too many potential locations to list out for this particular repo and I was unable to convince @htibosch of the need to do this level of scrutiny without a specific error case that explicitly exposed this type of failure.

amazonKamath (approved):

cobusve:
@phelter for what it is worth I believe you are right about the State Access Audit. We have already done a lot of work on this and will continue with doing that. If you have more suggestions I would love to hear them.

tools/CMakeLists.txt
amazonKamath: Thanks @phelter.  CC: @kstribrnAmzn, @shubnil

PR #559, opened 2022-09-26
IPv4/single: connect() should return immediately after a protocol error and other things

htibosch:
Description
-----------
A TCP connection can be established in either two ways: accept() and connect().

When using connect(), the socket is created and owned by the application.

When using accept(), the new socket is created and temporarily owned by the IP-task, and in the comments it is called an "orphaned socket". Once the connection is established, the function accept() will pass the ownership to the application.

The stack has a mechanism that checks sockets for inactivity: `ipconfigTCP_HANG_PROTECTION`. After `ipconfigTCP_HANG_PROTECTION_TIME` seconds of inactivity an orphaned socket will be closed.

There were a few reasons to create this PR:

1) When a protocol error occurs during a `FreeRTOS_connect()`, the API should stop waiting and return `-pdFREERTOS_ERRNO_ENOTCONN` to the application. At this moment it will only return when the timeout has been reached.

2) When a protocol error occurs during a passive connection, the socket will hang until the "hang protection" will close the socket. When a fatal protocol error occurs the client socket should be closed immediately.

3) The re-use option `FREERTOS_SO_REUSE_LISTEN_SOCKET`: when the listening socket is configured to turn itself into a "client socket", it may not be closed after a protocol error or after "hang protection". The reusable socket will be put into the listening state again and it will be able to receive a new connection.

Many thanks to [Am Paschal](https://forums.freertos.org/u/AmPaschal), who wrote the post "FreeRTOS connect and accept functions (in FreeRTOS+TCP) do not return even after the TCP connection closes" on the [FreeRTOS forum](https://forums.freertos.org/t/freertos-connect-and-accept-functions-in-freertos-tcp-do-not-return-even-after-the-tcp-connection-closes).

Also thanks to [Alex "oberondoro"](https://forums.freertos.org/u/oberondoro) who wrote the post "Sockets end in state eCLOSE_WAIT" on the [FreeRTOS forum](https://forums.freertos.org/t/sockets-end-in-state-eclose-wait).

And thanks to [Hubert Sack](https://forums.freertos.org/u/hs4FreeRtos) who explained the problem which occurs when a reusable socket has a protocol error or hang in the SYN phase.

Test Steps
-----------

Testing this PR was a bit complicated because of the many possibilities:

1) Test an outgoing connection using connect()
2) Test an incoming connection with a normal listening socket
3) Test an incoming connection with a reusable socket

X. Induce a protocol error during the SYN phase
Y. Induce a time during the SYN phase and get a hang protection.

I did 6 tests in total: 1X, 2X, 3X, and 1Y, 2Y, and 3Y.

When the listening socket is reusable, it should return to its listening state and be able to receive a new connection.
Normal client sockets should be closed when a protocol error or a timeout occurs so that the resources become available again.
`FreeRTOS_connect()` should return immediately when a protocol error occurs, or when the socket stays inactive for a long time.

### DIFF
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -3145,13 +3145,13 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                                                             pdFALSE /*xWaitAllBits*/,
                                                             xRemainingTime );
 
-				if( ( uxEvents & eSOCKET_CLOSED ) != 0U )
-				{
-					FreeRTOS_printf( ( "connect: xEventGroupWaitBits: returns 0x%04X\n", ( unsigned ) uxEvents ) );
-					xResult = -pdFREERTOS_ERRNO_ENOTCONN;
-					FreeRTOS_printf( ( "FreeRTOS_connect() stopped due to an error\n" ) );
-					break;
-				}
+                if( ( uxEvents & eSOCKET_CLOSED ) != 0U )
+                {
+                    FreeRTOS_printf( ( "connect: xEventGroupWaitBits: returns 0x%04X\n", ( unsigned ) uxEvents ) );
+                    xResult = -pdFREERTOS_ERRNO_ENOTCONN;
+                    FreeRTOS_printf( ( "FreeRTOS_connect() stopped due to an error\n" ) );
+                    break;
+                }
             }
         }
 
@@ -3300,18 +3300,10 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
 
                 /* Go sleeping until we get any down-stream event */
                 EventBits_t uxEvents = xEventGroupWaitBits( pxSocket->xEventGroup,
-                                                            ( EventBits_t ) eSOCKET_ACCEPT | ( EventBits_t ) eSOCKET_CLOSED,
+                                                            ( EventBits_t ) eSOCKET_ACCEPT,
                                                             pdTRUE /*xClearOnExit*/,
                                                             pdFALSE /*xWaitAllBits*/,
                                                             xRemainingTime );
-
-				if( ( uxEvents & eSOCKET_CLOSED ) != 0U )
-				{
-					FreeRTOS_printf( ( "accept: xEventGroupWaitBits: %p returns 0x%04X\n", pxSocket, ( unsigned ) uxEvents ) );
-					pxClientSocket = NULL;
-					FreeRTOS_printf( ( "FreeRTOS_accept() stopped due to an error\n" ) );
-					break;
-				}
             }
         }
 
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -3102,6 +3102,8 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
             /* And wait for the result */
             for( ; ; )
             {
+                EventBits_t uxEvents;
+
                 if( xTimed == pdFALSE )
                 {
                     /* Only in the first round, check for non-blocking */
@@ -3139,11 +3141,11 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                 }
 
                 /* Go sleeping until we get any down-stream event */
-                EventBits_t uxEvents = xEventGroupWaitBits( pxSocket->xEventGroup,
-                                                            ( EventBits_t ) eSOCKET_CONNECT | ( EventBits_t ) eSOCKET_CLOSED,
-                                                            pdTRUE /*xClearOnExit*/,
-                                                            pdFALSE /*xWaitAllBits*/,
-                                                            xRemainingTime );
+                uxEvents = xEventGroupWaitBits( pxSocket->xEventGroup,
+                                                ( EventBits_t ) eSOCKET_CONNECT | ( EventBits_t ) eSOCKET_CLOSED,
+                                                pdTRUE /*xClearOnExit*/,
+                                                pdFALSE /*xWaitAllBits*/,
+                                                xRemainingTime );
 
                 if( ( uxEvents & eSOCKET_CLOSED ) != 0U )
                 {
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -3147,7 +3147,6 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
 
                 if( ( uxEvents & eSOCKET_CLOSED ) != 0U )
                 {
-                    FreeRTOS_printf( ( "connect: xEventGroupWaitBits: returns 0x%04X\n", ( unsigned ) uxEvents ) );
                     xResult = -pdFREERTOS_ERRNO_ENOTCONN;
                     FreeRTOS_printf( ( "FreeRTOS_connect() stopped due to an error\n" ) );
                     break;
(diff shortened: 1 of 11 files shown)


source/FreeRTOS_Sockets.c
```diff
@@ -3139,7 +3139,18 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                 }
 
                 /* Go sleeping until we get any down-stream event */
-                ( void ) xEventGroupWaitBits( pxSocket->xEventGroup, ( EventBits_t ) eSOCKET_CONNECT, pdTRUE /*xClearOnExit*/, pdFALSE /*xWaitAllBits*/, xRemainingTime );
+                EventBits_t uxEvents = xEventGroupWaitBits( pxSocket->xEventGroup,
```
AniruddhaKanhere, line 3142: Maybe declare the variable at top of the block? Maybe just after `for( ; ; )`?

source/FreeRTOS_Sockets.c
```diff
@@ -3287,7 +3298,11 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                 }
 
                 /* Go sleeping until we get any down-stream event */
```
AniruddhaKanhere, line 3300:
```suggestion
                /* Put the calling task to 'sleep' until a down-stream event is received. */
```

source/FreeRTOS_TCP_IP.c
```diff
@@ -272,39 +272,56 @@
...
+              ( xPreviousState == eSYN_FIRST ) ||
+              ( xPreviousState == eSYN_RECEIVED ) ) &&
+            ( eTCPState == eCLOSE_WAIT ) )
+        {
+            /* A socket was in the connecting phase but something
+             * went wrong and it should be closed. */
+            FreeRTOS_printf( ( "Move from %s to %s\n",
+                               FreeRTOS_GetTCPStateName( xPreviousState ),
+                               FreeRTOS_GetTCPStateName( eTCPState ) ) );
+            bBefore = pdTRUE;
```
AniruddhaKanhere, line 295:
```suggestion
            /* Set the flag to show that it was connected before and that the
             * status has changed now. This will cause the control flow to go 
             * in the below if condition.*/
            bBefore = pdTRUE;
```

source/FreeRTOS_TCP_IP.c
```diff
@@ -393,30 +413,48 @@
...
 
+        if( ( eTCPState == eCLOSE_WAIT ) && ( pxSocket->u.xTCP.bits.bReuseSocket == pdTRUE_UNSIGNED ) )
+        {
+            switch( xPreviousState )
+            {
+                case eSYN_FIRST:    /* 3 (server) Just created, must ACK the SYN request */
+                case eSYN_RECEIVED: /* 4 (server) waiting for a confirming connection request */
+                    FreeRTOS_printf( ( "Restoring a reuse socket port %u\n", pxSocket->usLocalPort ) );
+
+                    /* Go back into list mode. Set the TCP status to 'eCLOSED',
```
AniruddhaKanhere, line 446:
```suggestion
                    /* Go back into listening mode. Set the TCP status to 'eCLOSED',
```

source/FreeRTOS_TCP_IP.c
```diff
@@ -393,30 +413,48 @@
...
+                     * otherwise FreeRTOS_listen() will refuse the action. */
+                    pxSocket->u.xTCP.eTCPState = eCLOSED;
+                    ( void ) FreeRTOS_listen( ( Socket_t ) pxSocket, pxSocket->u.xTCP.usBacklog );
```
AniruddhaKanhere, line 449: MISRA will not like this.
`FreeRTOS_listen` -> `vTCPStateChange` -> `FreeRTOS_listen`.
How is this recursion stopped? What is to say that it won't keep calling itself?

AniruddhaKanhere (changes requested):

source/FreeRTOS_TCP_IP.c
htibosch: I was aware of it, but:

- When `vTCPStateChange()` calls `FreeRTOS_listen()`, the current TCP state is either `eSYN_FIRST` or `eSYN_RECEIVED`.
- `FreeRTOS_listen()` will call `vTCPStateChange()` while the current TCP state is `eCLOSED`. That can not lead to a new call to `listen()`.

Of course I could copy some of the listen() code into `vTCPStateChange()`, but I prefer to call the original API.

source/FreeRTOS_Sockets.c
htibosch: Done so, thanks

htibosch:
About "Avoid a recursive call to vTCPStateChange()" :

Earlier we introduced `vSocketCloseNextTime()` so that a call to `closesocket()` could be postponed.

Now in order to avoid a recursive call to `vTCPStateChange()`, I introduced `vSocketListenNextTime()`.

Just before the IP-task starts waiting for the next message, `FreeRTOS_listen()` may be called.

This may happen when a reusable listening socket gets a timeout or a protocol error before it is connected.
~~~c
static FreeRTOS_Socket_t * xSocketToListen = NULL;
void vSocketListenNextTime( FreeRTOS_Socket_t * pxSocket )
{
    if( ( xSocketToListen != NULL ) && ( xSocketToListen != pxSocket ) )
    {
        ( void ) FreeRTOS_listen( ( Socket_t ) xSocketToListen, xSocketToListen->u.xTCP.usBacklog );
    }

    xSocketToListen = pxSocket;
}
~~~
I tested the changes in all possible combinations:

1. By calling connect()
2. By calling accept() with a reusable socket
3. By calling accept() with a normal socket

There were 3 scenarios:

1. The connection went well
2. A protocol error occurred
3. A timeout occurred

The results were as expected, in situations 2.2 and 2.3, a delayed call to `FreeRTOS_listen()` was made, and the listening socket got restored.

AniruddhaKanhere (approved):

source/FreeRTOS_TCP_IP.c
```diff
@@ -272,39 +299,60 @@
...
         #endif
 
+        if( ( ( xPreviousState == eCONNECT_SYN ) ||
+              ( xPreviousState == eSYN_FIRST ) ||
+              ( xPreviousState == eSYN_RECEIVED ) ) &&
+            ( eTCPState == eCLOSE_WAIT ) )
+        {
+            /* A socket was in the connecting phase but something
+             * went wrong and it should be closed. */
+            FreeRTOS_printf( ( "Move from %s to %s\n",
```
cobusve, line 319: It seems to me that this should probably be a debug_printf ?

cobusve (dismissed):

AniruddhaKanhere (approved):

source/FreeRTOS_Sockets.c
```diff
@@ -3139,7 +3141,18 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
...
+                uxEvents = xEventGroupWaitBits( pxSocket->xEventGroup,
+                                                ( EventBits_t ) eSOCKET_CONNECT | ( EventBits_t ) eSOCKET_CLOSED,
+                                                pdTRUE /*xClearOnExit*/,
+                                                pdFALSE /*xWaitAllBits*/,
+                                                xRemainingTime );
+
+                if( ( uxEvents & eSOCKET_CLOSED ) != 0U )
+                {
+                    xResult = -pdFREERTOS_ERRNO_ENOTCONN;
+                    FreeRTOS_printf( ( "FreeRTOS_connect() stopped due to an error\n" ) );
```
cobusve, line 3153: I think this is another one?

source/FreeRTOS_TCP_IP.c
```diff
@@ -393,30 +444,51 @@
...
         /* Fill in the new state. */
         pxSocket->u.xTCP.eTCPState = eTCPState;
 
+        if( ( eTCPState == eCLOSE_WAIT ) && ( pxSocket->u.xTCP.bits.bReuseSocket == pdTRUE_UNSIGNED ) )
+        {
+            switch( xPreviousState )
+            {
+                case eSYN_FIRST:    /* 3 (server) Just created, must ACK the SYN request */
+                case eSYN_RECEIVED: /* 4 (server) waiting for a confirming connection request */
+                    FreeRTOS_printf( ( "Restoring a reuse socket port %u\n", pxSocket->usLocalPort ) );
```
cobusve, line 475: Is this another debug message?

source/FreeRTOS_Sockets.c
AniruddhaKanhere: done

source/FreeRTOS_TCP_IP.c
AniruddhaKanhere: done

n9wxu (approved):

AniruddhaKanhere (approved):

PR #536, opened 2022-08-03
Fix the build issues for STM32Fxx (#433)

ugo1983:
This was reported here - https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/411

Signed-off-by: Gaurav Aggarwal <aggarg@amazon.com>

Co-authored-by: Aniruddha Kanhere <60444055+AniruddhaKanhere@users.noreply.github.com>

Description
-----------

Test Steps
-----------

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #560, opened 2022-09-27
IPv6/multi: Made corrections to the DNS code after integration testing

htibosch:
Description
-----------
**FreeRTOS_ARP.c**

Removed debug logging

**FreeRTOS_DNS.c**

Added logging for IPv6 addresses

In `prvIncreaseDNS6Index()` and `prvIncreaseDNS4Index()`: when changing the current DNS server, skip zero addresses "::" and "0.0.0.0".

`xDNS_IP_Preference` determines what type of DNS contact will be made: IPv6 or IPv4.
Test and obey this preference when iterating through the IP addresses of DNS servers.

Set `sin_family` and `sin_len` also in case an IPv4 DNS server is chosen.

Within `prvGetHostByName()`, there was confusion about `xAddress` and `xRecvAddress`,
the latter was declared but never used.

When `DNS_ReadReply()` returns a negative number, this is not fatal in case of `-pdFREERTOS_ERRNO_EWOULDBLOCK`.

**FreeRTOS_DNS_Cache.c**

in `FreeRTOS_ProcessDNSCache()`, `prvInsertCacheEntry()` would also be called when `xLookUp` is true.
That is not correct, it would create an invalid entry.

Corrected a typo in the logging: "will be stored" in stead of "will been stored"

Added `vShowDNSCacheTable()` as a help in debugging. It shows a list of the complete table.

In `Prepare_CacheLookup()`, clear `xIPv46_Address` looking up the address.

**FreeRTOS_DNS_Parser.c**

In `prvParseDNS_ReadAnswers()`, clear a variable `xIP_Address` before checking the answers.

**FreeRTOS_IP.c**

In `prvAllowIPPacketIPv6()`, allow any public IPv6 address for now.

**FreeRTOS_IP.h**

When logging about timers and timeouts, tick-counts are printed.
In some logging, it is more convenient to log seconds or milliseconds.
For this `ipTICKS_TO_MS()` is introduced, the reverse of `pdMS_TO_TICKS()`

**FreeRTOS_IP_Timers.c**

Log the RA reload time in milliseconds in stead of clock ticks.

**FreeRTOS_ND.c**

When `vNDAgeCache()` calls `vNDSendNeighbourSolicitation()`, it should make sure that the end-point is known.

**FreeRTOS_RA.c**

In 3 locations, logging will show milliseconds in stead of clock-ticks.
In `vRAProcess()`, the default RA reload time will be set to 60 seconds in stead of 5 seconds, which was used for testing.

**FreeRTOS_Router.c**

Added a new function `xIPv6AddressIsPublic()` which returns true if a given IPv6 address is a global unicast address.

**FreeRTOS_UDP_IP.c**

In `prvStartLookup()`, moved some logging to a different place so that `eReturned` can be printed.

**NetworkInterface/WinPCap/NetworkInterface.c**

When an application write to the PCAP interface, it may receive an echo of each of those messages, a "bouncing back".
The function `xPacketBouncedBack()` will check the source MAC-address of each packet and see if it is "ours" or "theirs".
However, the return statement in that function was placed in the wrong location.

Test Steps
-----------
I used several boards and also the [WinWim demo](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi/demos/IPv6_Multi_WinSim_demo) to test the changes.
In my next PR, I would like to update that demo.

Related Issue
-----------
At this moment, DNS will probably only work for IPv4 addresses.

### DIFF
--- a/source/FreeRTOS_ARP.c
+++ b/source/FreeRTOS_ARP.c
@@ -680,11 +680,6 @@ eARPLookupResult_t eARPGetCacheEntry( uint32_t * pulIPAddress,
     *( ppxEndPoint ) = NULL;
     ulAddressToLookup = *pulIPAddress;
 
-    if( ( *pulIPAddress == 0x70040120U ) || ( *pulIPAddress == 0x80feU ) )
-    {
-        FreeRTOS_printf( ( "eARPGetCacheEntry %lxip\n", FreeRTOS_ntohl( *pulIPAddress ) ) );
-    }
-
     pxEndPoint = FreeRTOS_FindEndPointOnIP_IPv4( ulAddressToLookup, 0 );
 
     if( xIsIPv4Multicast( ulAddressToLookup ) != pdFALSE )
--- a/source/FreeRTOS_DNS.c
+++ b/source/FreeRTOS_DNS.c
@@ -720,7 +720,17 @@
 
                     if( ulIPAddress != 0UL )
                     {
-                        FreeRTOS_debug_printf( ( "prvPrepareLookup: found '%s' in cache: %lxip\n", pcHostName, ulIPAddress ) );
+                        #if ( ipconfigUSE_IPv6 != 0 )
+                            if( ( *ppxAddressInfo )->ai_family == FREERTOS_AF_INET6 )
+                            {
+                                FreeRTOS_printf( ( "prvPrepareLookup: found '%s' in cache: %pip\n",
+                                                   pcHostName, ( *ppxAddressInfo )->xPrivateStorage.sockaddr6.sin_addrv6.ucBytes ) );
+                            }
+                            else
+                        #endif
+                        {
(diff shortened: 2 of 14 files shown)


htibosch:
About my commit ["Include the correct HAL header file"](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/560/commits/21bd0ebbfe116088f7f95340563a2857c20c45bc):
~~~diff
-        #include "stm32f7xx_hal_def.h"
+        #include "stm32f7xx_hal.h"
~~~
There is a [similar change](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/536) in the pipeline.

htibosch:
A lot of code from this PR has been included in later PR's. Now closing this PR because it is running far behind today's [dev/IPv6_integration](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/dev/IPv6_integration).

PR #562, opened 2022-10-11
NetifSlirp: Fix compiler warnings and errors

paulbartell:
NetifSlirp: Fix compiler warnings and errors

Description
-----------
Fix some compiler warnings / errors on msvc. 
In particular, windows does not define an ssize_t type, so ensure we can use the slirp_ssize_t type defined by the latest revision of libslirp.

Test Steps
-----------
Built on latest gcc, clang, and msvc for x86.

### DIFF
--- a/source/portable/NetworkInterface/libslirp/MBuffNetifBackendLibslirp.c
+++ b/source/portable/NetworkInterface/libslirp/MBuffNetifBackendLibslirp.c
@@ -593,10 +593,10 @@ static int lSlirpGetREventsCallback( int lIdx,
     configASSERT( pxCtx );
 
     configASSERT( lIdx >= 0 );
-    configASSERT( lIdx < ( int ) pxCtx->nfds );
-    configASSERT( lIdx < ( int ) pxCtx->xPollFdArraySize );
 
     xIndex = ( nfds_t ) lIdx;
+    configASSERT( xIndex < pxCtx->nfds );
+    configASSERT( xIndex < pxCtx->xPollFdArraySize );
 
     rEvents = ( pxCtx->pxPollFdArray[ xIndex ].revents );
 
--- a/source/portable/NetworkInterface/libslirp/MBuffNetifBackendLibslirp.c
+++ b/source/portable/NetworkInterface/libslirp/MBuffNetifBackendLibslirp.c
@@ -27,6 +27,7 @@
 
 /* libc */
 #include <stdlib.h>
+#include <sys/types.h>
 
 /* QEMU Slirp Library */
 #include <libslirp.h>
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (dismissed):

shubnil (dismissed):

source/portable/NetworkInterface/libslirp/MBuffNetifBackendLibslirp.c
```diff
@@ -588,8 +593,8 @@ static int lSlirpGetREventsCallback( int lIdx,
     configASSERT( pxCtx );
 
     configASSERT( lIdx >= 0 );
-    configASSERT( lIdx < pxCtx->nfds );
-    configASSERT( lIdx < pxCtx->xPollFdArraySize );
+    configASSERT( lIdx < ( int ) pxCtx->nfds );
```
cobusve, line 596: Are these casts safe? When nfds or xPllFdArraySize  is larger than an int this cast will cause it to be truncated and a value larger than lIdx may evaluate to being smaller here?

The C standard only requires int to be at least 16 bits but size_t (in the second case) has to be large enough to hold the size of the largest allocatable object so it could be larger than int ...

paulbartell: I don't see a problem on first glance with this. It's probably better to move the asserts after the conversion to nfds_t though.

paulbartell, line 596:
```suggestion
```

source/portable/NetworkInterface/libslirp/MBuffNetifBackendLibslirp.c
```diff
@@ -588,8 +593,8 @@ static int lSlirpGetREventsCallback( int lIdx,
...
+    configASSERT( lIdx < ( int ) pxCtx->xPollFdArraySize );
```
paulbartell, line 597:
```suggestion
```

source/portable/NetworkInterface/libslirp/MBuffNetifBackendLibslirp.c
```diff
@@ -588,8 +593,8 @@ static int lSlirpGetREventsCallback( int lIdx,
...
 
     xIndex = ( nfds_t ) lIdx;
```
paulbartell, line 600:
```suggestion
    configASSERT( xIndex < pxCtx->nfds );
    configASSERT( xIndex < pxCtx->xPollFdArraySize );

```

AniruddhaKanhere (approved):

PR #566, opened 2022-10-21
Add IPv6 specific header files.

moninom1:
Description
-----------

- Split FreeRTOS_IP header files to IPv4 and IPv6 specific header files and add IPv6 specific structures.

> FreeRTOS_IPv4.h
> FreeRTOS_IPv4_Private.h
> FreeRTOS_IPv6.h
> FreeRTOS_IPv6_Private.h
> FreeRTOS_IP_Common.h

- Added header files based on ipconfigUSE_IPV4 and ipconfigUSE_IPV6 to ensure backward compatibility.

- Created a common IP address structure to handle both IPv4 and IPv6 address : 
      IP_Address_t xIPAddress;      /**< Source or destination IP address, depending on usage scenario. */

Test Steps
-----------
As of now compilation test is being done. *
```
cmake -S test/build-combination -B test/build-combination/build/ -DTEST_CONFIGURATION=ENABLE_ALL
```

Related Issue
-----------
As unit tests will be done after merging other changes as well. **It might break some test cases as of now which will be fixed in upcoming PRs.**

### DIFF
--- a/source/FreeRTOS_ARP.c
+++ b/source/FreeRTOS_ARP.c
@@ -936,7 +936,7 @@ void FreeRTOS_OutputARPRequest( uint32_t ulIPAddress )
 
     if( pxNetworkBuffer != NULL )
     {
-        pxNetworkBuffer->ulIPAddress = ulIPAddress;
+        pxNetworkBuffer->xIPAddress.xIP_IPv4 = ulIPAddress;
         vARPGenerateRequestPacket( pxNetworkBuffer );
 
         #if ( ipconfigETHERNET_MINIMUM_PACKET_BYTES > 0 )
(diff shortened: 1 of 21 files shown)


htibosch (approved):
About the IPv4 - IPv6 merge:
---
In the coming months, we will merge two branches of FreeRTOS+TCP: IPv4/single and IPv6/multi. The result can be found in [dev/IPv6_integration](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/dev/IPv6_integration).
Please don't use this branch for your projects.

The application builder will be able to choose between IPv4, IPv6 or both IP-versions.

Also, it will be possible to use multiple interfaces, and have multiple IP-addresses per interface.

Currently, the [IPv4/single](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP) is known as stable, well-tested, and it also has LTS releases.

The [IPv6/multi branch](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi) is still "in the labs". We aim at making a new release which has the best of both.

At the same time there will be a cleanup of sources: we will separated the protocols, e.g. do not combine IPv4 and IPv6 in a single file, neither combine TCP with IP. This should make the source code easier to understand, and easier to maintain.

As always, the new releases will be downward compatible. The new software will be a superset of it was. You can include the same header files as before, and the same configuration files.
Except for your makefiles, because new sources files will have to be included.

Users of make (cmake) are lucky: wildcards can be used:
~~~
    C_SRCS += \
        $(wildcard $(PLUS_TCP_PATH)/source/*.c)
~~~


Thank you @moninom1 for this first separation of protocols. The only thing I could not get is:

~~~diff
 typedef struct xNETWORK_BUFFER
 {
     ListItem_t xBufferListItem; /**< Used to reference the buffer form the free buffer list or a socket. */
-    IP_Address_t xIPAddress;    /**< Source or destination IP address, depending on usage scenario. */
+    uint32_t ulIPAddress;       /**< Source or destination IP address, depending on usage scenario. */
     ...
 }
~~~

**EDIT**: sorry I looked in the wrong direction. You are introducing `xIPAddress` which is a union of both IP-type addresses. All good

Or is this a thing "to-be-done"?

Thank you for this:
~~~diff
 #include "FreeRTOS_IP_Utils.h"
 
+#if ipconfigUSE_IPV4
+    #include "FreeRTOS_IPv4.h"
+#endif /* ipconfigUSE_IPV4 */
+#if ipconfigUSE_IPV6
+    #include "FreeRTOS_IPv6.h"
+#endif /* ipconfigUSE_IPV6 */
 
 /* *INDENT-OFF* */
 #ifdef __cplusplus
     } /* extern "C" */
 #endif
 /* *INDENT-ON* */
~~~

This will guarantee downward compatibility: all new header files are automatically included.

I approve this PR

PR #567, opened 2022-11-03
IPv4/single SAME70 emac race condition

htibosch:
Description
-----------

A user on the FreeRTOS forum, [@Maty](https://forums.freertos.org/u/maty) reported that on a SAME70 the transmission stops after a while.

The post can be [found here](https://forums.freertos.org/t/freertos-10-1-1-http-server-on-sam4e-xplained-stops-after-sending-one-file/7678/5).

We both found a solution to this problem, but Maty's solution looks more efficient.

The CPU seems faster than the peripheral. When it enqueues all TX descriptors, it may happen that the peripheral forgets to fire an interrupt.

In this solution, the driver will use all descriptors but one, at the same time. See also [my post](https://forums.freertos.org/t/freertos-10-1-1-http-server-on-sam4e-xplained-stops-after-sending-one-file/7678/22) a detailed description.

We are both still running tests to make sure that the solution works well.

Test Steps
-----------
I had the SAME70 send packets to a laptop using the IPERF3 protocol, using the -R option:
~~~
iperf3 -c 192.168.2.12 --port 5001 --bytes 1000G -R
~~~
( I had to adapt my iperf server in order to understand 64-bit numbers like `1000G` )

### DIFF
--- a/source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
+++ b/source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
@@ -193,7 +193,7 @@ static void hand_tx_errors( void );
 /*-----------------------------------------------------------*/
 
 /* A copy of PHY register 1: 'PHY_REG_01_BMSR' */
-static volatile BaseType_t xGMACSwitchRequired;
+static BaseType_t xGMACSwitchRequired;
 
 /* LLMNR multicast address. */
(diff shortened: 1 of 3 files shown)


htibosch:
That is strange, uncrustify reports a failure but all files have a "PASS".

AniruddhaKanhere (approved):

shubnil (approved):

PR #568, opened 2022-11-04
IPv4/single: SAME70 EMAC buffer sizes

htibosch:
Description
-----------

In the [DriverSAM](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/main/source/portable/NetworkInterface/DriverSAM) driver, the macro `GMAC_FRAME_LENTGH_MAX` is the maximum supported frame length, 1536 bytes. That is the maximum number of bytes in an Ethernet packet.

The macro `NETWORK_BUFFER_SIZE` was supposed to define size of a network buffer, which should be 1536 bytes **plus** a `ipBUFFER_PADDING` (10) hidden bytes .

This PR makes sure that the statically allocated buffers in `ucNetworkPackets[]` are large enough to accommodate the largest possible incoming packet.

To summarize this PR:
~~~diff
-#define NETWORK_BUFFER_SIZE    ( 1536 )
+#define NETWORK_BUFFER_SIZE    ( 1536+16 )
~~~

`ipBUFFER_PADDING` is rounded up to 16 to get a nice alignment.

### DIFF
--- a/source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
+++ b/source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
@@ -71,12 +71,6 @@
 #define EMAC_IF_ERR_EVENT    4UL
 #define EMAC_IF_ALL_EVENT    ( EMAC_IF_RX_EVENT | EMAC_IF_TX_EVENT | EMAC_IF_ERR_EVENT )
 
-/* 1536 bytes is more than needed, 1524 would be enough.
- * But 1536 is a multiple of 32, which gives a great alignment for
- * cached memories. */
-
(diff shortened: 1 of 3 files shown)


htibosch:
In "ICMP checksum calculated manually" I re-enabled the function `vGMACGenerateChecksum()`, but only for ICMP protocol checksum.
As for the SAME70, it does not have to calculate UDP or TCP protocol checksums.

AniruddhaKanhere (dismissed):

shubnil (dismissed):

AniruddhaKanhere (approved):

shubnil (approved):

PR #569, opened 2022-11-07
Add Header size functions for IPv6 and IPv4 

moninom1:
Description
-----------

- Add common Header size function for IPv6 and IPv4
- Fix rule 1.1, 10.3, 20.9 and suppress rule 20.5

Test Steps
-----------
As of now compilation test is being done.
```
cmake -S test/build-combination -B test/build-combination/build/ -DTEST_CONFIGURATION=ENABLE_ALL
```

Related Issue
-----------
As unit tests will be done after merging other changes as well. It might break some test cases as of now which will be fixed in upcoming PRs.

### DIFF
--- a/source/include/FreeRTOS_IP_Private.h
+++ b/source/include/FreeRTOS_IP_Private.h
@@ -215,7 +215,7 @@ typedef struct IP_TASK_COMMANDS
 struct xPacketSummary
 {
     BaseType_t xIsIPv6;                          /**< pdTRUE for IPv6 packets. */
-    #if ( ipconfigUSE_IPv6 != 0 )
+    #if ipconfigUSE_IPV6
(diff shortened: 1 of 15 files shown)


source/include/FreeRTOS_IP_Private.h
```diff
@@ -215,7 +215,7 @@ typedef struct IP_TASK_COMMANDS
 struct xPacketSummary
 {
     BaseType_t xIsIPv6;                          /**< pdTRUE for IPv6 packets. */
-    #if ( ipconfigUSE_IPv6 != 0 )
+    #if ipconfigUSE_IPV6
```
htibosch, line 218: @moninom1, although this PR is already merged, one minor remark: in all FreeRTOS sources, the following syntax is being used:
~~~c
#if ( configENABLE_MPU == 1 )
    ...
#endif
~~~
And you're right that we want to get rid of most of the `ipconfigUSE_IPv6` tests.

Thanks

PR #579, opened 2022-11-17
Update deprecated macros in network driver files

pete-pjb:
# Update deprecated macros

Description
-----------
Update the deprecated macros: 
```c
PHY_LS_HIGH_CHECK_TIME_MS
PHY_LS_LOW_CHECK_TIME_MS
```
with new versions:
```c
ipconfigPHY_LS_HIGH_CHECK_TIME_MS
ipconfigPHY_LS_LOW_CHECK_TIME_MS
```

A number of the network drivers have not been updated for these macro redefinitions in an earlier update, so I have updated the remaining ones.

Test Steps
-----------
N/A

Related Issue
-----------

N/A


Skptak:
Hey @pete-pjb thanks for bringing these deprecated macros to our attention! I've looked over your changes and only had one comment for your review at this point.

source/portable/NetworkInterface/RX/NetworkInterface.c
```diff
@@ -67,18 +67,23 @@
...
 
-#ifndef PHY_LS_HIGH_CHECK_TIME_MS
+#if defined( PHY_LS_HIGH_CHECK_TIME_MS ) || defined( PHY_LS_LOW_CHECK_TIME_MS )
+    #warning please use the new defines with 'ipconfig' prefix
+#endif
+
+#ifndef ipconfigPHY_LS_HIGH_CHECK_TIME_MS
 
-/* Check if the LinkSStatus in the PHY is still high after 2 seconds of not
+/* Check if the LinkStatus in the PHY is still high after 15 seconds of not
```
Skptak, line 76: Should this be left at the 2 seconds that it was previously set to?

pete-pjb: Hi @Skptak ,

Thank you for pointing this out, my bad, I have pushed another commit to fix it!

Kind Regards,

Pete

htibosch: This was indeed a thing to be done: give a proper prefix to these macro's.

Good to issue a warning in case the old macro's are still defined ( in `FreeRTOSIPConfig.h` ):
~~~
#if defined( PHY_LS_HIGH_CHECK_TIME_MS ) || defined( PHY_LS_LOW_CHECK_TIME_MS )
    #warning please use the new defines with 'ipconfig' prefix
#endif
~~~
although I have a slight preference for an `#error` here.

Thanks a lot @pete-pjb.

Skptak (approved):

source/portable/NetworkInterface/RX/NetworkInterface.c
Skptak: Hey @pete-pjb thanks for putting this change in and being part of the FreeRTOS Community!
I've asked one of our Bar Raisers to take a look at this review

pete-pjb: Hi @Skptak ,

Should we change the #warnings to a #errors as @htibosch prefers?

I will make the changes if required let me know what you think...

Kind Regards,

Pete

AniruddhaKanhere: Hello @pete-pjb,

Thank you for this PR and for your patience.

I also prefer `#error` over `#warning` here. The reason being, we don't want users to be puzzled why their config is not working (i am assuming that there are many warnings in a dev project which is not production ready). `#error` will make sure that the users fix the config to get the expected results.
Also, some platforms (or compilers) do not understand `#warning` - Visual Studio using windows compiler is an example. But `#error` is widely accepted.

Let me know if you'd like me to make that change for you or would you do that.

Thanks again,
Aniruddha

pete-pjb: Hi @AniruddhaKanhere ,

I have made an additional commit to include the #error directive in favour of the #warning directive for all files concerned as preferred.

cc @htibosch & @Skptak 

Kind Regards,

Pete

pete-pjb: Hi @Skptak , @AniruddhaKanhere ,

For some reason the CI is no failing on Doxygen... I can't see why that would be, please let me know if I have caused this.

Kind Regards,

Pete

AniruddhaKanhere: Hello @pete-pjb,

I shall fix that check :)
We recently updated the doxygen check and thus it is failing. Will update it right now.

AniruddhaKanhere (approved):

shubnil (approved):

PR #589, opened 2022-11-24
Minor warning fixes

ChristosZosi:
Description
-----------

While building the FreeRTOS-Plus-TCP library for the Xilinx UltraScale SoC, I came across some open warnings that would be triggered in particular cases.

Check the commit messages for more info about the warnings.

Test Steps
-----------

Compiler used to generate the warning in this PR:

```
arm-xilinx-eabi-gcc.exe (GCC) 10.2.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
``` 

Compiler options:
```
1: -pedantic -Wall -Wextra -O0
2: -pedantic -Wall -Wextra -O2
```

Related Issue
-----------
None

### DIFF
--- a/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
+++ b/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
@@ -47,7 +47,7 @@ void setup_isr( xemacpsif_s * xemacpsif )
     /*
      * Setup callbacks
      */
-    XEmacPs *xInstancePtr = &xemacpsif->emacps;
+    XEmacPs * xInstancePtr = &xemacpsif->emacps;
 
     xInstancePtr->SendHandler = ( XEmacPs_Handler ) emacps_send_handler;
     xInstancePtr->SendRef = ( void * ) xemacpsif;
--- a/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
+++ b/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
@@ -47,7 +47,7 @@ void setup_isr( xemacpsif_s * xemacpsif )
     /*
      * Setup callbacks
      */
-    XEmacPs * xInstancePtr = &xemacpsif->emacps;
+    XEmacPs * xInstancePtr = &( xemacpsif->emacps );
 
     xInstancePtr->SendHandler = ( XEmacPs_Handler ) emacps_send_handler;
     xInstancePtr->SendRef = ( void * ) xemacpsif;
--- a/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
+++ b/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
@@ -47,17 +47,16 @@ void setup_isr( xemacpsif_s * xemacpsif )
     /*
      * Setup callbacks
      */
-    XEmacPs_SetHandler( &xemacpsif->emacps, XEMACPS_HANDLER_DMASEND,
-                        ( void * ) emacps_send_handler,
-                        ( void * ) xemacpsif );
+    XEmacPs *xInstancePtr = &xemacpsif->emacps;
 
-    XEmacPs_SetHandler( &xemacpsif->emacps, XEMACPS_HANDLER_DMARECV,
-                        ( void * ) emacps_recv_handler,
-                        ( void * ) xemacpsif );
+    xInstancePtr->SendHandler = ( XEmacPs_Handler ) emacps_send_handler;
+    xInstancePtr->SendRef = ( void * ) xemacpsif;
 
-    XEmacPs_SetHandler( &xemacpsif->emacps, XEMACPS_HANDLER_ERROR,
-                        ( void * ) emacps_error_handler,
-                        ( void * ) xemacpsif );
+    xInstancePtr->RecvHandler = ( XEmacPs_Handler ) emacps_recv_handler;
+    xInstancePtr->RecvRef = ( void * ) xemacpsif;
+
+    xInstancePtr->ErrorHandler = ( XEmacPs_ErrHandler ) emacps_error_handler;
+    xInstancePtr->ErrorRef = ( void * ) xemacpsif;
 }
 
 void start_emacps( xemacpsif_s * xemacps )
--- a/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
+++ b/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
@@ -49,13 +49,13 @@ void setup_isr( xemacpsif_s * xemacpsif )
      */
     XEmacPs * xInstancePtr = &( xemacpsif->emacps );
 
-    xInstancePtr->SendHandler = ( XEmacPs_Handler ) emacps_send_handler;
+    xInstancePtr->SendHandler = emacps_send_handler;
     xInstancePtr->SendRef = ( void * ) xemacpsif;
 
-    xInstancePtr->RecvHandler = ( XEmacPs_Handler ) emacps_recv_handler;
+    xInstancePtr->RecvHandler = emacps_recv_handler;
     xInstancePtr->RecvRef = ( void * ) xemacpsif;
 
-    xInstancePtr->ErrorHandler = ( XEmacPs_ErrHandler ) emacps_error_handler;
+    xInstancePtr->ErrorHandler = emacps_error_handler;
     xInstancePtr->ErrorRef = ( void * ) xemacpsif;
 }
 
--- a/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_physpeed.c
+++ b/source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_physpeed.c
@@ -245,18 +245,10 @@ void my_sleep( uint32_t uxTicks )
 u32 phymapemac0[ 32 ];
 u32 phymapemac1[ 32 ];
 
-static uint16_t prvAR803x_debug_reg_read( XEmacPs * xemacpsp,
-                                          uint32_t phy_addr,
-                                          u16 reg );
 static uint16_t prvAR803x_debug_reg_write( XEmacPs * xemacpsp,
                                            uint32_t phy_addr,
                                            u16 reg,
                                            u16 value );
-static int prvAR803x_debug_reg_mask( XEmacPs * xemacpsp,
-                                     uint32_t phy_addr,
-                                     u16 reg,
-                                     u16 clear,
-                                     u16 set );
 static void prvSET_AR803x_TX_Timing( XEmacPs * xemacpsp,
                                      uint32_t phy_addr );
 
@@ -832,32 +824,8 @@ static uint32_t get_AR8035_phy_speed( XEmacPs * xemacpsp,
     }
 }
 
-static void ar8035Tick( XEmacPs * xemacpsp,
-                        uint32_t phy_addr )
-{
-    uint16_t value;
-    BaseType_t linkState;
-
-    /*Read basic status register */
-    value = XEmacPs_PhyRead2( xemacpsp, phy_addr, IEEE_STATUS_REG_OFFSET );
-    /*Retrieve current link state */
-    linkState = ( value & IEEE_STAT_LINK_STATUS ) ? TRUE : FALSE;
-
-    /* Just to prevent compiler warnings about unused variable. */
-    ( void ) linkState;
-}
-
 #define AR803X_DEBUG_ADDR    0x1D
 #define AR803X_DEBUG_DATA    0x1E
-static uint16_t prvAR803x_debug_reg_read( XEmacPs * xemacpsp,
-                                          uint32_t phy_addr,
-                                          u16 reg )
-{
-    XEmacPs_PhyWrite( xemacpsp, phy_addr, AR803X_DEBUG_ADDR, reg );
-
-    return XEmacPs_PhyRead2( xemacpsp, phy_addr, AR803X_DEBUG_DATA );
-}
-
 static uint16_t prvAR803x_debug_reg_write( XEmacPs * xemacpsp,
                                            uint32_t phy_addr,
                                            u16 reg,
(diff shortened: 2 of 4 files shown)


source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
```diff
@@ -47,17 +47,16 @@ void setup_isr( xemacpsif_s * xemacpsif )
     /*
      * Setup callbacks
      */
-    XEmacPs_SetHandler( &xemacpsif->emacps, XEMACPS_HANDLER_DMASEND,
-                        ( void * ) emacps_send_handler,
-                        ( void * ) xemacpsif );
+    XEmacPs * xInstancePtr = &xemacpsif->emacps;
```
paulbartell, line 50:
```suggestion
    XEmacPs * xInstancePtr = &( xemacpsif->emacps );
```

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_physpeed.c
```diff
@@ -1015,7 +1021,11 @@ static uint32_t get_IEEE_phy_speed_US( XEmacPs * xemacpsp,
     XEmacPs_PhyRead( xemacpsp, phy_addr, PHY_IDENTIFIER_1_REG,
                      &phy_identity );
 
-    FreeRTOS_printf( ( "Start %s PHY autonegotiation. ID = 0x%04X\n", pcGetPHIName( phy_identity ), phy_identity ) );
+    const char * pcPHINAme = pcGetPHIName( phy_identity );
```
paulbartell, line 1024:
```suggestion
    const char * pcPHYName = pcGetPHIName( phy_identity );
```

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_physpeed.c
```diff
@@ -1015,7 +1021,11 @@ static uint32_t get_IEEE_phy_speed_US( XEmacPs * xemacpsp,
...
+    FreeRTOS_printf( ( "Start %s PHY autonegotiation. ID = 0x%04X\n", pcPHINAme, phy_identity ) );
+
+    /* Just to prevent compiler warnings about unused variablies. */
+    ( void ) pcPHINAme;
```
paulbartell, line 1028:
```suggestion
    ( void ) pcPHYName;
```

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_physpeed.c
```diff
@@ -1015,7 +1021,11 @@ static uint32_t get_IEEE_phy_speed_US( XEmacPs * xemacpsp,
     XEmacPs_PhyRead( xemacpsp, phy_addr, PHY_IDENTIFIER_1_REG,
                      &phy_identity );
 
-    FreeRTOS_printf( ( "Start %s PHY autonegotiation. ID = 0x%04X\n", pcGetPHIName( phy_identity ), phy_identity ) );
+    const char * pcPHINAme = pcGetPHIName( phy_identity );
+    FreeRTOS_printf( ( "Start %s PHY autonegotiation. ID = 0x%04X\n", pcPHINAme, phy_identity ) );
```
paulbartell, line 1025:
```suggestion
    FreeRTOS_printf( ( "Start %s PHY autonegotiation. ID = 0x%04X\n", pcPHYName, phy_identity ) );
```

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
```diff
@@ -47,17 +47,16 @@ void setup_isr( xemacpsif_s * xemacpsif )
...
-    XEmacPs_SetHandler( &xemacpsif->emacps, XEMACPS_HANDLER_DMARECV,
-                        ( void * ) emacps_recv_handler,
-                        ( void * ) xemacpsif );
+    xInstancePtr->SendHandler = ( XEmacPs_Handler ) emacps_send_handler;
+    xInstancePtr->SendRef = ( void * ) xemacpsif;
 
-    XEmacPs_SetHandler( &xemacpsif->emacps, XEMACPS_HANDLER_ERROR,
-                        ( void * ) emacps_error_handler,
-                        ( void * ) xemacpsif );
+    xInstancePtr->RecvHandler = ( XEmacPs_Handler ) emacps_recv_handler;
```
paulbartell, line 55: Are these casts to XEmacPs_Handler and XEmacPs_ErrHandler necessary?

paulbartell: For reference, https://github.com/Xilinx/embeddedsw/blob/master/XilinxProcessorIPLib/drivers/emacps/src/xemacps.h#L531 seems to list XEmacPs_Handler and XEmacPs_ErrHandler types for these fields.

paulbartell:
@ChristosZosi : Thanks for the pull request. I added a few suggestions to your changes.

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
ChristosZosi: No, as far as I know the casts shouldn't be necessary. However I think it is a nice extra touch to do it them explicitly.

ChristosZosi:
There are still some warnings that I get when I try to build this port, however I was not sure how to resolve them. Maybe you could give me some suggestions:


1. Unused functions in `x_emacpsif_physpeed.c`
        - `static void ar8035Tick( XEmacPs * xemacpsp, uint32_t phy_addr )` seems to be not used anywhere
        - static int prvAR803x_debug_reg_mask( XEmacPs * xemacpsp, uint32_t phy_addr, u16 reg, u16 clear, u16 set )
        - static uint16_t prvAR803x_debug_reg_read( XEmacPs * xemacpsp, uint32_t phy_addr, u16 reg )

     Can these functions be removed or what would be the alternative to "fix" the warnings?

2. Warning ` #warning the use of Jumbo Frames has not been tested sufficiently yet.`
    
     Can we remove these warnings? Or how one should test the Jumbo frame "sufficiently" enough, so that the warnings can be removed

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
paulbartell: I think it would be preferable to remove the explicit casts so that a compiler warning is generated if the definitions do not match up.

paulbartell:
> * Unused functions in `x_emacpsif_physpeed.c`
>   - `static void ar8035Tick( XEmacPs * xemacpsp, uint32_t phy_addr )` seems to be not used anywhere
>   - static int prvAR803x_debug_reg_mask( XEmacPs * xemacpsp, uint32_t phy_addr, u16 reg, u16 clear, u16 set )
>   - static uint16_t prvAR803x_debug_reg_read( XEmacPs * xemacpsp, uint32_t phy_addr, u16 reg )
>   Can these functions be removed or what would be the alternative to "fix" the warnings?
@AniruddhaKanhere and @htibosch Do you have a preference on this?

> * Warning ` #warning the use of Jumbo Frames has not been tested sufficiently yet.`
>   Can we remove these warnings? Or how one should test the Jumbo frame "sufficiently" enough, so that the warnings can be removed

It sounds like you have validated Jumbo Frames, so the warning can probably be removed. @htibosch Do you have a particular set of tests in mind?

AniruddhaKanhere:
> There are still some warnings that I get when I try to build this port, however I was not sure how to resolve them. Maybe you could give me some suggestions:
> 
> 1. Unused functions in `x_emacpsif_physpeed.c`
>    - `static void ar8035Tick( XEmacPs * xemacpsp, uint32_t phy_addr )` seems to be not used anywhere
>    - static int prvAR803x_debug_reg_mask( XEmacPs * xemacpsp, uint32_t phy_addr, u16 reg, u16 clear, u16 set )
>    - static uint16_t prvAR803x_debug_reg_read( XEmacPs * xemacpsp, uint32_t phy_addr, u16 reg )
>    Can these functions be removed or what would be the alternative to "fix" the warnings?

This file wasn't created by us and I would like to refrain from modifying it. It seems to me that this file is quite old and, thus maybe a bit outdated. If possible, we should replace it with a newer version of the file.

> 2. Warning ` #warning the use of Jumbo Frames has not been tested sufficiently yet.`
>    Can we remove these warnings? Or how one should test the Jumbo frame "sufficiently" enough, so that the warnings can be removed

That was there to let the users know that the jumbo frame support is a new feature. And that they should use caution while using that as it might have some bugs. I agree that 'sufficiently' is not a measurable/quantifiable term, but it does convey the gist to the user :) .
Once we have tested Jumbo frame on Xilinx Ultrascale and have run various tests on it (such as sending many 'jumbo' packets, sending a packet bigger than jumbo frame, sending a mixture of both type of frames etc.), we shall remove that warning. But unless we are sure that it is safe to use and will not lead to some bad behavior of the code, I think we should leave that warning as is.

Does that seem apt @paulbartell and @htibosch?

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_physpeed.c
```diff
@@ -842,6 +842,9 @@ static void ar8035Tick( XEmacPs * xemacpsp,
     value = XEmacPs_PhyRead2( xemacpsp, phy_addr, IEEE_STATUS_REG_OFFSET );
```
htibosch, line 842: `ar8035Tick`, `prvAR803x_debug_reg_mask`, `prvAR803x_debug_reg_read` : if they're unused and static, I wouldn't mind if they are removed.

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
```diff
@@ -47,17 +47,16 @@ void setup_isr( xemacpsif_s * xemacpsif )
     /*
      * Setup callbacks
      */
-    XEmacPs_SetHandler( &xemacpsif->emacps, XEMACPS_HANDLER_DMASEND,
-                        ( void * ) emacps_send_handler,
-                        ( void * ) xemacpsif );
```
htibosch, line 52: I missed this change: why would you remove the function calls and replace them with direct assignments?
Was there a problem with the functions? Or did it introduce a dependency on sources?

htibosch (approved):
@ChristosZosi, I'm sorry to see that this PR lasts so long.

We are very busy with the IPv6/multi version of FreeRTOS+TCP.

About the warning and the Jumbo frames: I agree with what Aniruddha wrote about it. The stack has been written and tested with 1500-byte frames. I did a short test with Jumbo frames which went well. But I can't find the time yet to do extensive tests.

Ref FreeRTOS_Socket.c and FreeRTOS_ICMP.c : I hope that your changes will be preserved in the new stage if the library.

Here below two more remarks.
Thanks, Hein

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
AniruddhaKanhere: Hello @ChristosZosi,
I agree with Hein here. I am also not sure why was this removed?

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
AniruddhaKanhere: +1 to Paul's comment above.
If anytime in future if we change the typedef, I would like the compiler to issue a warning (which cannot happen with a cast).
Would you mind making this change please?

AniruddhaKanhere (changes requested):
Thank you for being patient with us @ChristosZosi.
Can you please take a look at the comments and make the changes? I can make them on your behalf if you like.
Since this PR has been open for so long due to the (much appreciated) discussion, I will go ahead and make these changes if we do not hear from you soon.

Thanks,
Aniruddha

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
ChristosZosi: No at all, I am going to address the comment in a coming commit.

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
ChristosZosi: In the commit message (9829705528a353d340b68fe901074105a3d3c257) I did add an explanation why the calls to the `XEmacPs_SetHandler` is problematic. In short, the compiler would output the pedantic warning `ISO C forbids conversion of object pointer to function pointer type`, when compiled with `-Wpedantic`. 

The reason for it has to do with the function prototype of  `XEmacPs_SetHandler`
```
LONG XEmacPs_SetHandler(XEmacPs *InstancePtr, u32 HandlerType,
			void *FuncPointer, void *CallBackRef);
``` 

Here you can see that the third parameter of the function is a `void`-pointer. However we pass to the function a function pointer. And here lies the issue, that function pointers (text) are not the same as void pointers (data). That is the reason for the compiler warning.

Here is also a nice discussion on stack overflow, which addresses this exact issue https://stackoverflow.com/questions/36645660/why-cant-i-cast-a-function-pointer-to-void 

I hope this explains it 😄.

ChristosZosi:
Hi @AniruddhaKanhere,

I did address all the open comments in the new commit (bd4c4e96c72ca41eee623ae0462c2c75e0d4aebe). However I am currently on holidays and unfortunately I do not have a setup to test the new changes. Although these are insignificant changes, I would still suggest for them to be verified. I can do it also, but in two weeks.

Thank you

AniruddhaKanhere (approved):

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
htibosch: Thank you for the URL and for the insights.

Actually I had the same problem when using `setsockopt()` which has a parameter `void *optval`. It was not allowed to pass a function pointer.

As a solution, I introduced a struct that holds a function pointer called [F_TCP_UDP_Handler_t](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/afcedead21c747cef64f07c7fedd50df75bcbd10/source/include/FreeRTOS_Sockets.h#L432).

htibosch (approved):
Thank for these changes, I approve them.

htibosch:
@AniruddhaKanhere wrote:

> unfortunately I do not have a setup to test the new changes

Today I retested an Ultrascale project by running [iperf3](https://github.com/htibosch/freertos_plus_projects/blob/master/plus/Common/Utilities/iperf_task_v3_0f.c) on this board:

image

All went well, as expected.

kstribrnAmzn (approved):

PR #592, opened 2022-11-28
Fix Windows thread calling vTaskSuspendAll / xTaskResumeAll.

jasonpcarroll:
Description
-----------

Fixed an issue where a Windows thread used to receive from WinPCap inadvertently called vTaskSuspendAll/xTaskResumeAll (added to uxStreamBufferAdd by https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/517) when adding to the circular buffer used to pass received data from WinPCap to a FreeRTOS task. This would cause WinSim to crash (very intermittently) in the case that the pcap callback suspends the scheduler while a Task running in WinSim is at point where it needs the scheduler to be in a running state (e.g. calling xQueueReceive, vTaskDelay, etc.).

Test Steps
-----------

Ran all WinSim demos from FreeRTOS/FreeRTOS.

Related Issue
-----------

### DIFF
--- a/source/portable/NetworkInterface/WinPCap/NetworkInterface.c
+++ b/source/portable/NetworkInterface/WinPCap/NetworkInterface.c
@@ -104,6 +104,20 @@ static void prvInterruptSimulatorTask( void * pvParameters );
 static void prvCreateThreadSafeBuffers( void );
 
 /*
+ * This function is equivalent to uxStreamBufferAdd from
+ * FreeRTOS_Stream_Buffer.c in the case that the stream buffer is being used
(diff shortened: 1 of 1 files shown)


htibosch (approved):
Hi Jason, we already tasked about the issue that was caused by [PR #517](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/517) for the WinSim driver.
I approve the PR that you are submitting, and thank you for it.

AniruddhaKanhere (approved):

amazonKamath (approved):

paulbartell (approved):

PR #596, opened 2022-12-06
Updated comments for FreeRTOS_select return value

tony-josi-aws:
Update function documentation for FreeRTOS_select API

Description
-----------
The documentation for return value of FreeRTOS_select API has been updated to fix the previous incorrect documentation. 

Test Steps
-----------
No additional tests performed as there is no code changes.

Related Issue
-----------
[FreeRTOS_select incorrect return value documentation](https://forums.freertos.org/t/freertos-plus-tcp-freertos-select-incorrect-return-value-documentation/16333)
Thanks [Igor_Petrov](https://forums.freertos.org/u/Igor_Petrov) for reporting this discrepancy.

### DIFF
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -704,7 +704,7 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
 
 /**
  * @brief The select() statement: wait for an event to occur on any of the sockets
- *        included in a socket set.
+ *        included in a socket set and return its event bits when the event occurs.
  *
  * @param[in] xSocketSet: The socket set including the sockets on which we are
  *                        waiting for an event to occur.
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -713,8 +713,8 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
  *                   indefinitely for an event to occur.
  *
  * @return The event bits (event flags) value for the socket set in which an
- *          event occurred. If any socket is signalled during the call, using 
- *          FreeRTOS_SignalSocket() or FreeRTOS_SignalSocketFromISR(), then eSELECT_INTR 
+ *          event occurred. If any socket is signalled during the call, using
+ *          FreeRTOS_SignalSocket() or FreeRTOS_SignalSocketFromISR(), then eSELECT_INTR
  *          is returned.
  *
  */
@@ -5137,9 +5137,9 @@ void * pvSocketGetSocketID( const ConstSocket_t xSocket )
  *        and return the value -pdFREERTOS_ERRNO_EINTR ( -4 ).
  *
  * @param[in] xSocket: The socket that will be signalled.
- * 
- * @return If xSocket is an invalid socket (NULL) or if the socket set is invalid (NULL) 
- *         and/or if event group is invalid/not created, then, -pdFREERTOS_ERRNO_EINVAL 
+ *
+ * @return If xSocket is an invalid socket (NULL) or if the socket set is invalid (NULL)
+ *         and/or if event group is invalid/not created, then, -pdFREERTOS_ERRNO_EINVAL
  *         is returned. Else 0 is returned.
  */
     BaseType_t FreeRTOS_SignalSocket( Socket_t xSocket )
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -713,8 +713,9 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
  *                   indefinitely for an event to occur.
  *
  * @return The event bits (event flags) value for the socket set in which an
- *          event occurred. During the call if any socket is signalled
- *          (using FreeRTOS_SignalSocket), then eSELECT_INTR is returned.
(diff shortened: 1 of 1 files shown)


source/FreeRTOS_Sockets.c
```diff
@@ -712,7 +712,8 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
  *                   If the value is 'portMAX_DELAY' then the function will wait
  *                   indefinitely for an event to occur.
  *
- * @return The socket which might have triggered the event bit.
+ * @return The event bits (event flags) value for the socket in which an event occurred.
```
AniruddhaKanhere, line 715:
```suggestion
 * @return The event bits (event flags) value for the socket set in which an
 *         event occurred. If any socket is signaled (using
 *         FreeRTOS_SignalSocket), then eSELECT_INTR is returned.
```

AniruddhaKanhere (changes requested):
We should also add the return value in this function's brief. https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/FreeRTOS_Sockets.c#L5137

I added a comment too about the description of return value. Would you mind taking a look?

source/FreeRTOS_Sockets.c
htibosch: Isn't `signaled` written with two l's?

Beside that, @tony-josi-aws, are using `ipconfigSUPPORT_SIGNALS`? Does it work well? Do you use it ti interrupt a call to `FreeRTOS_select()`?

tony-josi-aws: @htibosch fixed the typos. 

Havent used `ipconfigSUPPORT_SIGNALS` yet. Will update once I use it.

AniruddhaKanhere: @htibosch this PR is for a forum post here: https://forums.freertos.org/t/freertos-plus-tcp-freertos-select-incorrect-return-value-documentation/16333.
The post creator must be using select functions - I think they can provide you good feedback.

source/FreeRTOS_Sockets.c
```diff
@@ -704,15 +704,18 @@ Socket_t FreeRTOS_socket( BaseType_t xDomain,
...
  * @param[in] xSocketSet: The socket set including the sockets on which we are
  *                        waiting for an event to occur.
  * @param[in] xBlockTimeTicks: Maximum time ticks to wait for an event to occur.
  *                   If the value is 'portMAX_DELAY' then the function will wait
  *                   indefinitely for an event to occur.
  *
- * @return The socket which might have triggered the event bit.
+ * @return The event bits (event flags) value for the socket set in which an
+ *          event occurred. During the call if any socket is signalled
+ *          (using FreeRTOS_SignalSocket), then eSELECT_INTR is returned.
```
htibosch, line 717: I would suggest the following change:
~~~diff
-    During the call if any socket is signalled
-    (using FreeRTOS_SignalSocket), then eSELECT_INTR is returned.

+    If any socket is signalled during the call, using FreeRTOS_SignalSocket()
+    or FreeRTOS_SignalSocketFromISR(), then eSELECT_INTR is returned.
~~~

htibosch (approved):
@tony-josi-aws, thank you for this PR. I approve it.

source/FreeRTOS_Sockets.c
tony-josi-aws: updated

source/FreeRTOS_Sockets.c
```diff
@@ -5133,6 +5137,9 @@ void * pvSocketGetSocketID( const ConstSocket_t xSocket )
  *        and return the value -pdFREERTOS_ERRNO_EINTR ( -4 ).
  *
  * @param[in] xSocket: The socket that will be signalled.
+ * 
+ * @return If xSocket is an invalid socket then -pdFREERTOS_ERRNO_EINVAL 
+ *         is returned. Else 0 is returned.
```
AniruddhaKanhere, line 5142: This seems a bit incomplete. Here, the `-pdFREERTOS_ERRNO_EINVAL` is returned whenever there is an invalid input. that may be socket being invalid or the event group not being created and/or the socket set being NULL.

We would need to modify this comment to reflect that.

Thanks

AniruddhaKanhere (changes requested):

source/FreeRTOS_Sockets.c
```diff
@@ -5133,6 +5137,10 @@ void * pvSocketGetSocketID( const ConstSocket_t xSocket )
  *        and return the value -pdFREERTOS_ERRNO_EINTR ( -4 ).
  *
  * @param[in] xSocket: The socket that will be signalled.
+ *
+ * @return If xSocket is an invalid socket (NULL) or if the socket set is invalid (NULL)
+ *         and/or if event group is invalid/not created, then, -pdFREERTOS_ERRNO_EINVAL
+ *         is returned. Else 0 is returned.
```
AniruddhaKanhere, line 5143:
```suggestion
 *         is returned. On successful sending of a signal, 0 is returned.
```

AniruddhaKanhere (changes requested):

AniruddhaKanhere (approved):

aggarg (approved):

PR #604, opened 2022-12-12
Fixing MISRA violation for FreeRTOS_IP*, FreeRTOS_DHCP*

xuelix:
Description
-----------

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_DHCP.c
+++ b/source/FreeRTOS_DHCP.c
@@ -82,9 +82,14 @@
 /*-----------------------------------------------------------*/
 
 /*
+ * @brief The number of end-points that are making use of the UDP-socket.
+ */
+    static BaseType_t xDHCPSocketUserCount = 0;
+
+/*
  * Generate a DHCP discover message and send it on the DHCP socket.
  */
-    static BaseType_t prvSendDHCPDiscover( NetworkEndPoint_t * pxEndPoint );
+    static BaseType_t prvSendDHCPDiscover( const NetworkEndPoint_t * pxEndPoint );
 
 /*
  * Interpret message received on the DHCP socket.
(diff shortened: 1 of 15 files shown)


source/FreeRTOS_DHCP.c
```diff
@@ -312,8 +321,12 @@
                 }
             }
         }
+        else
+        {
+            /* do nothing, coverity happy */
+        }
 
-        if( ( pxEndPoint != NULL ) && ( xDoProcess != pdFALSE ) )
+        if( xDoProcess != pdFALSE )
```
moninom1, line 329: What is the reason for removing pxEndPoint NULL check?

source/FreeRTOS_IP.c
```diff
@@ -190,6 +178,10 @@ uint16_t usPacketIdentifier = 0U;
  * reference. */
 const MACAddress_t xBroadcastMACAddress = { { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff } };
 
+/* __XX__ temporary for coverity build */
+/** @brief Structure that stores the netmask, gateway address and DNS server addresses. */
+NetworkAddressingParameters_t xNetworkAddressing = { 0, 0, 0, 0, 0 };
+
```
moninom1, line 184: Why is this needed ?

htibosch: Do you mean why it is initialised, or what the function is of `xNetworkAddressing`?
It is initialised just to show it explicitly.
The struct will be dropped as soon as we have multiple interfaces. The struct will be replaced by end-points.

moninom1: Yes hein, as we have moved to end point based implementation , i removed it in previous PR, was just checking why was it needed again for fixing coverity.

moninom1 (approved):

PR #661, opened 2023-01-04
Fixed minor issue with vDNSCheckCallBack() on checking if temp list i…

tony-josi-aws:
Fixed minor issue with vDNSCheckCallBack() on checking if temp list is empty
-----------
Updated the conditional statement for checking if `xTempList` is empty.

Test Steps
-----------
Validated vDNSCheckCallBack() with unit tests for DNS_Callback from dev/IPv6 branch 

Related Issue
-----------
Check if list is empty was not correct.

### DIFF
--- a/source/FreeRTOS_DNS_Callback.c
+++ b/source/FreeRTOS_DNS_Callback.c
@@ -278,7 +278,7 @@
         }
         ( void ) xTaskResumeAll();
 
-        if( listLIST_IS_EMPTY( &xTempList ) != pdFALSE )
+        if( listLIST_IS_EMPTY( &xTempList ) == pdFALSE )
         {
             /* There is at least one item in xTempList which must be removed and deleted. */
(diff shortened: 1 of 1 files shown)


htibosch (approved):
When I wrote mentioned code, I got confused by the word `EMPTY` and `FALSE`.

But you are right, it should have been:
~~~
if( listLIST_IS_EMPTY( &xTempList ) != pdFALSE )
{
    /* Go through the list. */
}
~~~
Note that this only applies to the 'labs/ipv6_multi' branch. In the main branch it was done differently, and maybe that is [more elegant](https://github.com/htibosch/FreeRTOS-Plus-TCP/blob/IPv4_single_take_DNS_callback_out_of_suspended/source/FreeRTOS_DNS_Callback.c#L80).

Thanks a lot for this PR

moninom1 (approved):

shubnil (approved):

PR #663, opened 2023-01-05
Fix failing test cases for DNS unit tests 

tony-josi-aws:
Description
-----------
This PR fixes the failing test cases for DNS unit tests which were failing after the IPv6 integration changes. 

NOTE: The unit test: _`test_CreateSocket_bind_fail`_ from DNS_Networking unit test module has been removed as its not relevant to the `DNS_CreateSocket` function after IPv6 integration changes. The flow of binding is validated in a different test case (as part of the test cases for `FreeRTOS_gethostbyname`) instead.

Test Steps
-----------
```
cmake -S test/unit-test -B test/unit-test/build/ 
cd test/unit-test/build/
ctest -E system --output-on-failure --verbose
```

Related Issue
-----------
DNS unit test cases were failing due to the changes in the source code as part of the IPv6 integration


source/FreeRTOS_DNS.c
```diff
@@ -897,7 +897,25 @@
...
+            #endif
+        }
+        else 
+        {
+            xDNS_IP_Preference = xPreferenceIPv4;
+        }
+
         pxEndPoint = prvFillSockAddress( &xAddress, pcHostName );
+
+        xDNS_IP_Preference = xDNS_IP_Preference_Temp;
```
moninom1, line 917: why is this neeeded?

moninom1 (approved):

source/FreeRTOS_DNS.c
tony-josi-aws: Thought of restoring back the `xDNS_IP_Preference` (if it was something else before) after we are done with the function call? Maybe we can remove that as `xDNS_IP_Preference ` is not used anywhere else.

shubnil (approved):

moninom1 (approved):

PR #664, opened 2023-01-08
Reparations and recommendations for the dev/IPv6_integration branch

htibosch:
Description
-----------
Nost changes in this branch are related to DNS.

- Renamed pCallback to pCallbackFucntion for clarity

- Allows for multiple DNS server, hopping to the next one after a failure.

- Added FreeRTOS_getaddrinfo() and FreeRTOS_getaddrinfo_a() to get multiple find results and also to return IPv6 addresses.

- Started using the new unified 'struct freertos_sockaddr', which is downward compatible with the labs/IPv6_multi branch.

- A new function prvPrepare_ReadIPAddress() reads a string and see if it is a valid IP-address.

- Here is the new flow of functions in DNS:

~~~
FreeRTOS_getaddrinfo()
    prvPrepareLookup()
        prvPrepare_ReadIPAddress() ? resolve "192.168.2.1" to an IP-address
        Prepare_CacheLookup() ? try to find the host name in DNS cache
        If not resolve, call prvGetHostByName()
~~~	
- Added FreeRTOS_IPStart() and kept the old 'FreeRTOS_IPInit()' for backward compatibility in case 'ipconfigCOMPATIBLE_WITH_SINGLE' is defined.

- I tried to get rid of old macros `ipLOCAL_IP_ADDRESS_POINTER` and `ipLOCAL_MAC_ADDRESS` and replace them with references to end-points. Job still unfinished.

~~~diff
-    pvCopySource = ipLOCAL_MAC_ADDRESS;
+    pvCopySource = pxTargetEndPoint->xMACAddress.ucBytes;
~~~

Also I replaced these calls:

~~~diff
-    ( void ) xNetworkInterfaceOutput( pxNetworkBuffer, pdTRUE );
+    pxEndPoint->pxNetworkInterface->pfOutput( pxEndPoint->pxNetworkInterface, pxNetworkBuffer, pdTRUE );
~~~

Places some #idef's at the beginning of a module like e.g. :
~~~
/* *INDENT-OFF* */
#if( ipconfigUSE_IPv6 != 0 )
/* *INDENT-ON* */
    ...
/* *INDENT-OFF* */
#endif /* ipconfigUSE_IPv6 != 0 */
/* *INDENT-ON* */
~~~

Test Steps
-----------
I used the [IPv6_Multi_WinSim_demo demo](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/labs/ipv6_multi/demos/IPv6_Multi_WinSim_demo), but it might need some more adaptations.


source/FreeRTOS_UDP_IP.c
```diff
@@ -288,14 +281,13 @@ BaseType_t xProcessReceivedUDPPacket( NetworkBufferDescriptor_t * pxNetworkBuffe
 
     if( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE )
     {
-        xReturn = xProcessReceivedUDPPacket_IPv4( pxNetworkBuffer,
-                                                  usPort, pxIsWaitingForARPResolution );
+        xProcessReceivedUDPPacket_IPv4( pxNetworkBuffer,
```
tony-josi-aws, line 284: Can we return the return value of the corresponding the IPV4/V6 functions as well to enable testing?

source/FreeRTOS_UDP_IP.c
```diff
@@ -288,14 +281,13 @@ BaseType_t xProcessReceivedUDPPacket( NetworkBufferDescriptor_t * pxNetworkBuffe
...
+                                        usPort, pxIsWaitingForARPResolution );
     }
-
-    if( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv6_FRAME_TYPE )
+    else if( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv6_FRAME_TYPE )
     {
-        xReturn = xProcessReceivedUDPPacket_IPv6( pxNetworkBuffer,
-                                                  usPort, pxIsWaitingForARPResolution );
+        xProcessReceivedUDPPacket_IPv6( pxNetworkBuffer,
```
tony-josi-aws, line 289: Can we return the return value of the corresponding the IPV4/V6 functions as well to enable testing?

htibosch: All functions that handle incoming packets may either use the network buffer to return it, or return false, meaning : the buffer was not released.
I found that network buffers got lost because a "pdTRUE" was returned.

tony-josi-aws: In this case aren't we returning the packets using the network buffers?

The `xReturn`  in this case is set to `pdTRUE` always.

source/FreeRTOS_DNS_Callback.c
```diff
@@ -231,7 +231,7 @@
         if( listLIST_IS_EMPTY( &xTempList ) == pdFALSE )
         {
             /* There is at least one item in xTempList which must be removed and deleted. */
-            xEnd = listGET_END_MARKER( &xTempList );
+            xEnd = ( ( const ListItem_t * ) &( xTempList.xListEnd ) );
```
tony-josi-aws, line 234: Can we use the `listGET_END_MARKER` here, which will enable using mocks to manipulate the `xTempList` during unit testing

moninom1 (approved):

PR #689, opened 2023-01-30
Changes after testing all other protocols IPv4 and IPv6

htibosch:
Description
-----------
● UDP.c : add code to forward mDNS packet to the DNS module

● IP.c : Let prvProcessUDPPacket() return `prvProcessUDPPacket` by default, meaning the the caller must release the packet.

● Sockets.c : MSS must be 20 byte smaller when IPv6 is used

● Sockets.c : set `bIsIPv6` as soon as possible, based on `xDomain`

● Sockets.c : The Ipv4 field `pxSocket->u.xTCP.xRemoteIP.ulIP_IPv4` was set unconditionally, disturbing IPv6 communication

● DNS_Parser.c : Also allow AAAA (IPv6) records

● ARP.c : send an ARP request to all IPv4 end-points that have been initialised ( by DHCP or static ).

● DHCP.c : I had 2 end-points with a different MAC-address, both doing the DHCP procedure. Made changes to make this possible.

● DHCP.c : `FreeRTOS_recvfrom()` was called after the socket `xDHCPv4Socket` was closed, issuing an error code

● DHCP.c : The statement `xDHCPSocketUserCount++;` was never reached due to an erroneous if/else clause.

● DHCP.c : replaced more `ipLOCAL_MAC_ADDRESS` macro's with `pxEndPoint->xMACAddress`.

### DIFF
--- a/source/FreeRTOS_DHCP.c
+++ b/source/FreeRTOS_DHCP.c
@@ -199,6 +199,10 @@
     {
         BaseType_t xDoProcess = pdTRUE;
 
+        /* The function is called by the IP-task, so pxEndPoint
+         * should be non-NULL. */
+        configASSERT( pxEndPoint != NULL );
+
         /* Is DHCP starting over? */
         if( xReset != pdFALSE )
         {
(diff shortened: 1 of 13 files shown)


source/FreeRTOS_DHCP.c
```diff
@@ -867,7 +870,7 @@
                 /* Bind to the standard DHCP client port. */
                 xAddress.sin_port = ( uint16_t ) dhcpCLIENT_PORT_IPv4;
                 xReturn = vSocketBind( xDHCPv4Socket, &xAddress, sizeof( xAddress ), pdFALSE );
-                /*configASSERT( xReturn == 0 ); */
+                configASSERT( xReturn == 0 );
```
moninom1, line 873: @htibosch  Why is this needed? xReturn is properly handled at line 879

moninom1 (approved):

PR #698, opened 2023-02-02
Use vTaskDelay for sleep in the network-interface of xilinx_ultrascale

ChristosZosi:
Description
-----------

The issue here is that, the FreeRTOS IP-task would block all other tasks during PHY-link speed negotiations, as it was using busy waiting. However this is not really ideal. A much more suitable function for such a task would be `vTaskDelay`.

Test Steps
-----------
This PR was tested within our application with real hardware and the behaviour of the IP-task was as expected, meaning it would no longer block the other the tasks of the application during link speed negotiations. 

Related Issue
-----------
NA


n9wxu (approved):

AniruddhaKanhere (approved):
Thank you @ChristosZosi for this PR.

It looks good to me. I am approving this. @htibosch can you also look?

PR #699, opened 2023-02-03
Fix unit test build issues and test failures for IP Utils

tony-josi-aws:
Description
-----------
This PR fixes unit test build issues and test failures for IP utils unit tests which were failing after the latest IPv6 integration changes.

Test Steps
-----------
```
cmake -S test/unit-test -B test/unit-test/build/ 
cd test/unit-test/build/
ctest -E system --output-on-failure --verbose
```

Related Issue
-----------
IP utils unit test cases were failing due to the changes in the source code as part of the IPv6 integration.

### DIFF
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -186,6 +186,8 @@ NetworkBufferDescriptor_t * pxDuplicateNetworkBufferWithDescriptor( const Networ
 
     if( pxNewBuffer != NULL )
     {
+        configASSERT( pxNewBuffer->pucEthernetBuffer != NULL );
+
         /* Get the minimum of both values to copy the data. */
         if( uxLengthToCopy > pxNetworkBuffer->xDataLength )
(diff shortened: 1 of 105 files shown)


source/FreeRTOS_IP_Utils.c
```diff
@@ -550,7 +554,7 @@ static void prvChecksumProtocolSetChecksum( BaseType_t xOutgoingPacket,
         *( pxSet->pusChecksum ) = pxSet->usChecksum;
     }
 
-    #if ( ipconfigHAS_DEBUG_PRINTF != 0 )
+    #if ( ipconfigHAS_DEBUG_PRINTF == 0 )
```
moninom1, line 557: As suggested offline, can you update this with proper return, instead of == 0 {which was a workaround}

tony-josi-aws: sure, updated

moninom1 (approved):

PR #702, opened 2023-02-07
A few changes after testing Xueli's branch, mostly TCP_Transmission

htibosch:
Here is a summary of the changes in this PR:
-----------
FreeRTOS_TCP_Transmission_IPV6.c

● In prvTCPReturnPacket_IPV6: use 'TCPPacket_IPv6_t' in stead of 'TCPPacket_t'

● When it has a socket, copy address from socket
● When it has no socket, swap the destination and source IP-addresses

● In prvTCPPrepareConnect_IPV6(): use `TCPPacket_IPv6_t` and `IPHeader_IPv6_t` in stead of the IPv4 variants.

FreeRTOS_TCP_Transmission_IPV4.c

● When it has a socket, copy address from socket
● When it has no socket, swap the destination and source IP-addresses

FreeRTOS_TCP_Transmission.c

● Let prvTCPSendSpecialPacketHelper() return 'xReturn' in stead of const 'pdFAIL'

FreeRTOS_TCP_IP.c

● In xProcessReceivedTCPPacket() : no need to set `xResult` if it will be overwritten unconditionally.

FreeRTOS_Sockets.c:

● In prvAcceptWaitClient(), set `*pxAddressLength` only in 1 location

FreeRTOS_IP.c:

● Removed the unused `xInterfaces[ 1 ]`

FreeRTOS_DNS_Parser.c :

● `prepareReplyDNSMessage(0) should call `vReturnEthernetFrame()` to reply to a request

FreeRTOS_DNS.c

● Changed a comment: from `dnsTYPE_AAA_HOST` to `dnsTYPE_AAAA_HOST`.

FreeRTOS_DHCP.c

● `vDHCPProcess()`: it should only be called to work on an IPv4 endpoint. Assrt if the wrong type is received.

The problem with TCP.close will be solved in my next PR, not this one.

### DIFF
--- a/source/FreeRTOS_TCP_Transmission_IPV6.c
+++ b/source/FreeRTOS_TCP_Transmission_IPV6.c
@@ -89,7 +89,7 @@
                                   uint32_t ulLen,
                                   BaseType_t xReleaseAfterSend )
     {
-        TCPPacket_t * pxTCPPacket = NULL;
+        TCPPacket_IPv6_t * pxTCPPacket = NULL;
         ProtocolHeaders_t * pxProtocolHeaders = NULL;
         IPHeader_IPv6_t * pxIPHeader = NULL;
         BaseType_t xDoRelease = xReleaseAfterSend;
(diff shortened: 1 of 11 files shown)


source/FreeRTOS_TCP_Transmission_IPV6.c
```diff
@@ -89,7 +89,7 @@
                                   uint32_t ulLen,
                                   BaseType_t xReleaseAfterSend )
     {
-        TCPPacket_t * pxTCPPacket = NULL;
+        TCPPacket_IPv6_t * pxTCPPacket = NULL;
```
shubnil, line 92: Can we retain the older structure "TCPPacket_t" as this is going to break the unit tests.

shubnil (approved):

PR #708, opened 2023-02-10
Change in UDP process flow to drop packet in case of invalid packet length

kar-rahul-aws:
Change in UDP packet process flow to drop packet in case of invalid packet length

Description
-----------
File changed:
FreeRTOS_IP.c

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -1607,12 +1607,7 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
     /* Note the header values required prior to the checksum
      * generation as the checksum pseudo header may clobber some of
      * these values. */
-    if( usLength > ( FreeRTOS_ntohs( pxIPHeader->usLength ) - uxIPHeaderSizePacket( pxNetworkBuffer ) ) )
-    {
-        /* The UDP packet is bigger than the IP-payload. Something is wrong, drop the packet. */
-        eReturn = eReleaseBuffer;
-    }
-    else if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
+    if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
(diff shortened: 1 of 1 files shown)


shubnil (approved):

source/FreeRTOS_IP.c
```diff
@@ -1817,8 +1817,20 @@ static eFrameProcessingResult_t prvProcessIPPacket( const IPPacket_t * pxIPPacke
 
                     case ipPROTOCOL_UDP:
                         /* The IP packet contained a UDP frame. */
-
-                        eReturn = prvProcessUDPPacket( pxNetworkBuffer );
+                        {
+                            const UDPPacket_t * pxUDPPacket = ( ( const UDPPacket_t * ) pxNetworkBuffer->pucEthernetBuffer );
+                            uint16_t usLength;
+                            usLength = FreeRTOS_ntohs( pxUDPPacket->xUDPHeader.usLength );
+                            if( usLength > ( FreeRTOS_ntohs( pxIPHeader->usLength ) - uxIPHeaderSizePacket( pxNetworkBuffer ) ) )
```
shubnil, line 1824: pxIPHeader can be obtained from NetworkBufferDescriptor_t pointer and we are passing the same to prvProcessUDPPacket.

shubnil (changes requested):

source/FreeRTOS_IP.c
htibosch: MISRA doesn't like the cast from `pucEthernetBuffer` to any of the structs. If a function needs such a cast, I tend to do so one time only.
For instance, `UDPPacket_t` already has a pointer to `IPHeader_t`, so you don't need the extra cast from `pucEthernetBuffer`.

It looks like this function should be compatible with IPv6, and that will need some adaptations.
The IPv6 header has `usPayloadLength`, which is defined as the "number of bytes after the IPv6 header".

So the patch works well for IPv4, it has an undefined behaviour for IPv6 packets.

PR #714, opened 2023-02-14
Fix compilation error in FreeRTOS_BitConfig.h

moninom1:
The sizes of array variables in C must be known at compile time. Here sizeof result is a constant. However, if we assign it to a constant variable it is still a variable to C.
Hence updating the array size parameter.


chinglee-iot (approved):
To support visual studio, const variable can't be used in array length.
This PR fixes this problem.
Reference the following page for more information.
https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2057?view=msvc-170

shubnil (approved):

PR #715, opened 2023-02-14
Change in UDP process flow to drop packet in case of invalid payload for IPv4

kar-rahul-aws:
**Title**
Change in UDP process flow to drop packet in case of invalid payload for IPv4

**Description**
Change in prvProcessUDPPacket function, add condition for Release buffer 
Checking the frame type only for IPv4, since the same check will give unwanted results for IPv6

Test Steps
-----------
Code is building with the changes

Related Issue
-----------


shubnil (changes requested):
There are lot of unwanted changes which is probably because of wrong base branch. Please rebase the local branch and push the PR again.

PR #716, opened 2023-02-14
Change in UDP process flow to drop packet for invalid IPv4 payload length

kar-rahul-aws:
Change in UDP process flow to drop packet for invalid IPv4 payload length
Description
-----------
Change in prvProcessUDPPacket function, add condition for Release buffer
Checking the frame type only for IPv4, since the same check will give unwanted results for IPv6

Test Steps
-----------
Code is building with the changes

Related Issue
-----------


shubnil (changes requested):
There are lot of unwanted changes which is probably because of wrong base branch. You will have to rebase the local branch and push the PR again. The issue is same as https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/715

shubnil:
My bad, uncrustify is adding the extra indentations

shubnil (approved):

moninom1 (approved):

Issue #700, opened 2023-02-05
[BUG] Invalid reset packet (incorrect ack sequence number) when connection backlog is full


t123yh:
**Describe the bug**
If you set connection backlog to a low limit, and connection count starts to reach the limit so new connections cannot be handled, it is expected that FreeRTOS+TCP will send RST packet upon receiving a SYN request it cannot handle (see [FreeRTOS_TCP_State_Handling.c#L963](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/a9f7c9a317c37499de17014216c550a559bc505c/source/FreeRTOS_TCP_State_Handling.c#L963)).

However, the RST packets sent here have a wrong ACK value, as you can see in the following screenshot. 192.168.2.102 (client, Linux) is trying to connect to 192.168.1.115 (server running FreeRTOS+TCP). The client sent a SYN packet with sequence number 615815458, and server replied the same ACK number. 

![image](https://user-images.githubusercontent.com/8972552/216835793-68cd9945-9789-42d3-abbd-0a246550f65d.png)

This behavior will make TCP stack of the client ignore this RST packet, because as per RFC793, Page 69, when a segment arrives for a connection in a state other than LISTEN, it is at first checked whether the sequence number falls into the receiver window, and if it doesn't, it's dropped ([ref](https://superuser.com/a/616052)). When client sent SYN with seq=n, it is expected that the RST packet comes back with ack=n+1 (you can check this behavior on any normal Linux PC). In the screenshot below, 192.168.1.1 is a normal Linux server:

![image](https://user-images.githubusercontent.com/8972552/216836164-8e06379e-8343-4d85-97fd-875f8f1f150e.png)

**Target**
- Development board: rp2040+w5500
- Instruction Set Architecture: ARM Cortex m0+
- IDE and version: CLion
- Toolchain and version: arm-none-eabi-gcc (GNU Arm Embedded Toolchain 10.3-2021.10) 10.3.1 

**Host**
- Host OS: Linux
- Version: Debian 11

**To Reproduce**
- Create a simple TCP server with FreeRTOS+TCP that listens but never accepts connections. Set the backlog to 3.
- Connect to this server with TCP clients on your PC. Make 4 connections simultaneously.

**Expected behavior**
An RST packet with correct ACK number is sent from FreeRTOS+TCP. The fourth TCP client reports `Connection refused`.

** Actual behaviour **
[...]

PR #724, opened 2023-02-16
Main/TCP4 : ACK number in TCP RESET reply to SYN packet

htibosch:
Description
-----------

[@t123yh](https://github.com/t123yh) reported in [issue #700](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/700) that when a device responds with a RST packet to a SYN request, the ACK number is not increased by 1. A SYN packet has a symbolical length of 1 data byte and should be acknowledged.

That is a very useful change, before the TCP client would reach a time-out, as show here:
~~~
Connecting to 192.168.2.15 ...
TCP connection timeout
~~~

After this change, the client sees a valid ACK in the RST reply, and it can report an error:

~~~
Connecting to 192.168.2.15 ...
TCP connection error :10061
~~~
10061 : No connection could be made because the target machine actively refused it

@t123yh, thank you for reporting this bug.

I will make sure that the change will also be made to the development branch [IPv6_integration](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/dev/IPv6_integration)

### DIFF
--- a/source/FreeRTOS_TCP_Transmission.c
+++ b/source/FreeRTOS_TCP_Transmission.c
@@ -1469,7 +1469,7 @@
 
                 if( ( ucFlagsReceived & tcpTCP_FLAG_SYN ) != 0U )
                 {
-                    /* A synchronise packet is received. It counts as 1 pseudo byte of data,
+                    /* A synchronize packet is received. It counts as 1 pseudo byte of data,
                      * so increase the variable with 1. Before sending a reply, the values of
                      * 'ulSequenceNumber' and 'ulAckNr' will be swapped. */
                     uint32_t ulSequenceNumber = FreeRTOS_ntohl( pxTCPPacket->xTCPHeader.ulSequenceNumber );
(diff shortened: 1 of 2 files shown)


AniruddhaKanhere (approved):

AniruddhaKanhere (approved):

source/FreeRTOS_TCP_Transmission.c
```diff
@@ -1474,7 +1474,7 @@
                      * 'ulSequenceNumber' and 'ulAckNr' will be swapped. */
                     uint32_t ulSequenceNumber = FreeRTOS_ntohl( pxTCPPacket->xTCPHeader.ulSequenceNumber );
                     ulSequenceNumber++;
-                    pxTCPPacket->xTCPHeader.ulSequenceNumber = FreeRTOS_ntohl( ulSequenceNumber );
+                    pxTCPPacket->xTCPHeader.ulSequenceNumber = FreeRTOS_htonl( ulSequenceNumber );
```
htibosch, line 1477: `ntohl` versus `htonl` : as you know, their math is the same, but the semantics are different.
Thank you!

AniruddhaKanhere: Yes, internally, both the functions are same. But, for the sake of consistency, I changed this as we are putting data into the network buffer which should be in network format.
Thank you @htibosch for this PR.

tony-josi-aws (approved):

PR #731, opened 2023-02-17
 Fix for assert in vNDAgeCache function

shubnil:
Fix for assert in vNDAgeCache function

Description
-----------
Add a condition so that NDageCache functions just returns gracefully when invoked for IPv4 cases.

Test Steps
-----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_ND.c
+++ b/source/FreeRTOS_ND.c
@@ -572,81 +572,81 @@
         MACAddress_t xMultiCastMacAddress;
         NetworkBufferDescriptor_t * pxDescriptor = pxNetworkBuffer;
 
-        configASSERT( pxEndPoint != NULL );
-        configASSERT( pxEndPoint->bits.bIPv6 != pdFALSE_UNSIGNED );
-
-        uxNeededSize = ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + sizeof( ICMPHeader_IPv6_t );
-
-        if( pxDescriptor->xDataLength < uxNeededSize )
+        if( ( pxEndPoint != NULL ) && ( pxEndPoint->bits.bIPv6 != pdFALSE_UNSIGNED ) )
         {
-            pxDescriptor = pxDuplicateNetworkBufferWithDescriptor( pxDescriptor, uxNeededSize );
-        }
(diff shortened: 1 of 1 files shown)


tony-josi-aws (approved):

kar-rahul-aws (approved):

source/FreeRTOS_ND.c
```diff
@@ -550,81 +550,81 @@
...
         NetworkBufferDescriptor_t * pxDescriptor = pxNetworkBuffer;
 
-        configASSERT( pxEndPoint != NULL );
-        configASSERT( pxEndPoint->bits.bIPv6 != pdFALSE_UNSIGNED );
-
-        uxNeededSize = ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + sizeof( ICMPHeader_IPv6_t );
-
-        if( pxDescriptor->xDataLength < uxNeededSize )
+        if( ( pxEndPoint != NULL ) && ( pxEndPoint->bits.bIPv6 != pdFALSE_UNSIGNED ) )
         {
```
htibosch, line 554: Hello Shub, thanks for this PR. The proposed change consists of replacing the two assert statement with an if clause:

~~~diff
     void vNDSendNeighbourSolicitation( NetworkBufferDescriptor_t * const pxNetworkBuffer,
                                        const IPv6_Address_t * pxIPAddress )
     {
 
-        configASSERT( pxEndPoint != NULL );
-        configASSERT( pxEndPoint->bits.bIPv6 != pdFALSE_UNSIGNED );
+        if( ( pxEndPoint != NULL ) && ( pxEndPoint->bits.bIPv6 != pdFALSE_UNSIGNED ) )
+        {
             uxNeededSize = ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + sizeof( ICMPHeader_IPv6_t );
             ...

+        }
~~~

That will surely help avoiding an assert, but, as I commented last Friday, I'm curious to see from where it is called with improper parameters.

I couldn't reproduce the error, but it looks like `vNDAgeCache()` calls `vNDSendNeighbourSolicitation()` with invalid parameters.

So maybe someone can do the same test and put some checks or asserts into the function `vNDRefreshCacheEntry()`?

shubnil: Hi Hein, I agree that vNDAgeCache should not enter here. However, irrespective of that the check is valid and is helping avoid a protocol test issue. We would like to merge the same and continue to check the function vNDRefreshCacheEntry.

PR #737, opened 2023-02-21
Fix build errors and warnings for IAR compiler

tony-josi-aws:
Description
-----------
This PR fixes the build errors and compilation warnings for the IAR compiler. 

Test Steps
-----------
Build succeeds without errors or warnings other than unreferenced static functions.

Checklist:
----------

Related Issue
-----------
Build issues and warnings in IAR build.

### DIFF
--- a/source/FreeRTOS_DHCP.c
+++ b/source/FreeRTOS_DHCP.c
@@ -89,7 +89,7 @@
 /*
  * Generate a DHCP discover message and send it on the DHCP socket.
  */
-    static BaseType_t prvSendDHCPDiscover( const NetworkEndPoint_t * pxEndPoint );
+    static BaseType_t prvSendDHCPDiscover( NetworkEndPoint_t * pxEndPoint );
 
 /*
  * Interpret message received on the DHCP socket.
@@ -100,7 +100,7 @@
 /*
  * Generate a DHCP request packet, and send it on the DHCP socket.
  */
-    static BaseType_t prvSendDHCPRequest( const NetworkEndPoint_t * pxEndPoint );
+    static BaseType_t prvSendDHCPRequest( NetworkEndPoint_t * pxEndPoint );
 
 /*
  * Prepare to start a DHCP transaction.  This initialises some state variables
@@ -1450,7 +1450,7 @@
  *
  * param[in] pxEndPoint: The end-point for which the request will be sent.
  */
-    static BaseType_t prvSendDHCPRequest( const NetworkEndPoint_t * pxEndPoint )
+    static BaseType_t prvSendDHCPRequest( NetworkEndPoint_t * pxEndPoint )
     {
         BaseType_t xResult = pdFAIL;
         uint8_t * pucUDPPayloadBuffer;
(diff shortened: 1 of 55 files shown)


source/FreeRTOS_DHCP.c
```diff
@@ -100,7 +100,7 @@
 /*
  * Generate a DHCP request packet, and send it on the DHCP socket.
  */
-    static BaseType_t prvSendDHCPRequest( const NetworkEndPoint_t * pxEndPoint );
+    static BaseType_t prvSendDHCPRequest( NetworkEndPoint_t * pxEndPoint );
```
moninom1, line 103: Why are we removing const? They might have been added for MISRA fixes

source/FreeRTOS_IP_Utils.c
```diff
@@ -829,10 +829,8 @@ void vPreCheckConfigs( void )
             if( uxSize == 8U )
             {
                 /* This is a 64-bit platform, make sure there is enough space in
-                 * pucEthernetBuffer to store a pointer. */
-                configASSERT( ipconfigBUFFER_PADDING >= 14 );
-                /* But it must have this strange alignment: */
-                configASSERT( ( ( ( ipconfigBUFFER_PADDING ) + 2 ) % 4 ) == 0 );
+                 * pucEthernetBuffer to store a pointer and it must have this strange alignment. */
+                configASSERT( ( ipconfigBUFFER_PADDING >= 14 ) && ( ( ( ( ipconfigBUFFER_PADDING ) + 2 ) % 4 ) == 0 ) );
```
moninom1, line 833: It should be ||

source/FreeRTOS_TCP_Transmission_IPV4.c
```diff
@@ -349,6 +349,7 @@
         BaseType_t xReturn = pdTRUE;
         uint32_t ulInitialSequenceNumber = 0;
         NetworkEndPoint_t * pxEndPoint = pxSocket->pxEndPoint;
+        ( void ) pxEndPoint;
```
moninom1, line 352: This is also a MISRA fix

source/FreeRTOS_DHCP.c
tony-josi-aws: Thats becuase in the same function we are assigning it to a non const pointer (`xDHCPv4Socket->pxEndPoint = pxEndPoint;`) and IAR considers this as an error. I think MISRA shouldn't be complaining about that in this case.

source/FreeRTOS_IP_Utils.c
tony-josi-aws: We need to assert if either `( ipconfigBUFFER_PADDING >= 14 )` or `( ( ( ( ipconfigBUFFER_PADDING ) + 2 ) % 4 ) == 0 )` is false, as `configASSERT` will be triggered if the output of the conditional expression is 0 I think we might need `&&`.

moninom1: Yes if one condition is 0 and other is 1, we should ASSERT, so && make sense.

PR #743, opened 2023-02-27
Zero copy fix

tony-josi-aws:
Description
-----------
This PR fixes the zero copy functionality on the dev/IPv6_Integration branch.

Test Steps
-----------
Tested with IPv4.

Checklist:
----------

Related Issue
-----------
The IP type of the generated buffer was not set correctly

### DIFF
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -786,9 +786,12 @@ void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
                                      TickType_t uxBlockTimeTicks,
                                      uint8_t ucIPType )
 {
+
     NetworkBufferDescriptor_t * pxNetworkBuffer;
     void * pvReturn;
     TickType_t uxBlockTime = uxBlockTimeTicks;
+    size_t uxPayloadOffset = 0;
+    BaseType_t xPayloadIPTypeOffset;
 
     /* Cap the block time.  The reason for this is explained where
      * ipconfigUDP_MAX_SEND_BLOCK_TIME_TICKS is defined (assuming an official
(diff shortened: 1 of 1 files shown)


source/FreeRTOS_IP.c
```diff
@@ -798,34 +799,37 @@ void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
...
-            uint8_t * pucIPType;
-            size_t uxIPHeaderLength;
+        /* Obtain a network buffer with the required amount of storage. */
+        pxNetworkBuffer = pxGetNetworkBufferWithDescriptor( uxPayloadOffset + uxRequestedSizeBytes, uxBlockTime );
 
-            /* Calculate the distance between the beginning of
-             * UDP payload until the hidden byte that reflects
-             * the IP-type: either ipTYPE_IPv4 or ipTYPE_IPv6.
-             */
+        if( pxNetworkBuffer != 0 )
```
shubnil, line 821: Please suggest why we are not checking against NULL. This should be if ( pxNetworkBuffer != NULL )

shubnil (changes requested):

source/FreeRTOS_IP.c
tony-josi-aws: Fixed, thanks

source/FreeRTOS_IP.c
```diff
@@ -839,10 +904,7 @@ void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
         }
         else
         {
-            /* Set the actual packet size in case a bigger buffer was returned. */
-            pxNetworkBuffer->xDataLength = sizeof( UDPPacket_t ) + uxRequestedSizeBytes;
-            /* Skip 3 headers. */
-            pvReturn = &( pxNetworkBuffer->pucEthernetBuffer[ sizeof( UDPPacket_t ) ] );
+            pvReturn = NULL;
```
htibosch, line 907: We tend to initialise the return variable to the most likely outcome:
~~~
    void * pvReturn = NULL;
~~~

This avoids the need for llines like this:
~~~
            *pucIPType = ucIPType;
        }
        else
        {
            pvReturn = NULL;
        }
    }
    else
    {
        pvReturn = NULL;
    }
~~~
and also it avoids that `pvReturn` will be used without initialising it.

source/include/FreeRTOS_IP.h
```diff
@@ -349,6 +349,9 @@ void FreeRTOS_SetEndPointConfiguration( const uint32_t * pulIPAddress,
 TaskHandle_t FreeRTOS_GetIPTaskHandle( void );
 
 void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
+                                     TickType_t uxBlockTimeTicks );
+
+void * FreeRTOS_GetUDPPayloadBuffer_ByIPType( size_t uxRequestedSizeBytes,
                                      TickType_t uxBlockTimeTicks,
                                      uint8_t ucIPType );
```
htibosch, line 356: Please add a parameter `ucIPType` to `FreeRTOS_GetUDPPayloadBuffer()`, and do not add a new function.

In labs/ipv6 we already had this function:
~~~
#if ( ipconfigUSE_IPv6 == 1 )
void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
                                     TickType_t uxBlockTimeTicks,
                                     uint8_t ucIPType )
~~~

htibosch (approved):
Hi @tony-josi-aws , thank you for this PR.

It looks OK to me, but why did you add a new function and the change signature of `FreeRTOS_GetUDPPayloadBuffer()` ?

Let us stay with this single existing function:
~~~
void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
                                     TickType_t uxBlockTimeTicks,
                                     uint8_t ucIPType );
~~~
and give it the implementation that you propose.

That is also what existing labs/IPv4-multi applications are already using.

Thank you

source/include/FreeRTOS_IP.h
tony-josi-aws: Hello Hein, Thanks for the review.

The reason for adding 2 functions is to maintain backward compatibility to projects that use the main branch. So the behavior  `FreeRTOS_GetUDPPayloadBuffer` is similar to main branch and the projects already using it can continue to use it without modifications, but `FreeRTOS_GetUDPPayloadBuffer_ByIPType` can be used for getting IPv6/IPv4 buffers.

source/FreeRTOS_IP.c
tony-josi-aws: Thanks, will update that.

tony-josi-aws:
> Member

done

tony-josi-aws:
@moninom1 done

source/include/FreeRTOS_IP.h
htibosch: Ok, if you want to offer thius backward compatibility, please don't write the logic two times, but have

tony-josi-aws: Sure, I have made the changes. The comments for `FreeRTOS_GetUDPPayloadBuffer` have been updated to notify the user that its kept only for backward compatibility and the user should use `FreeRTOS_GetUDPPayloadBuffer_ByIPType` to allocate a IPv4 or IPv6 buffer.

htibosch: Sorry no, I prefer to keep `FreeRTOS_GetUDPPayloadBuffer()` only, along with the `ucIPType` parameter.
Normally I am also in favour of keeping functions backward compatible. But not in this case, because now the application writer must become aware that there are two types of UDP payload buffers: IPv4 and IPv6.

In labs/IPv6 we had this:

~~~
#if ( ipconfigUSE_IPv6 != 0 )
    void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
                                         TickType_t uxBlockTimeTicks,
                                         uint8_t ucIPType );
#else
    void * FreeRTOS_GetUDPPayloadBuffer( size_t uxRequestedSizeBytes,
                                         TickType_t uxBlockTimeTicks );
#endif
~~~

But as we stopped using the macro `ipconfigUSE_IPv6`, only the first function is left.

shubnil: Thanks Tony and Hein. I think it's best if we retain the original version of the API and create a new one for IPv4 and IPv6 scenario. The reason being : 1. We are following the same approach for the new API and we would like to be consistent 2. When the new code is merged with mainline, the existing IPv4 users will still be able to continue with the existing application.
However, I do agree that we can have a common logic in a single function and have 2 abstraction APIs to handle the 2 scenarios.

shubnil (approved):

PR #747, opened 2023-02-27
Merge buffer allocation methods

holden-zenith:
Merge Buffer Allocations methods

Description
-----------
The two files had fallen out of sync but can share a significant amount of code. A config macro should be defined to determine which method to use.

Checklist:
----------

### DIFF
--- a/.github/workflows/uncrustify.yml
+++ b/.github/workflows/uncrustify.yml
@@ -8,7 +8,7 @@ jobs:
   Uncrustify:
     name: Run_Uncrustify
     if: ${{ github.event.issue.pull_request && github.event.comment.body == '/bot run uncrustify' }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Dump GitHub context
         env:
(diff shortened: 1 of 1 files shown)


htibosch:
Thank you @holden-zenith for this welcome initiative.

`BufferAllocation_1.c` indeed has much more safety checks. That is the module that is mostly used.

Two problems can emerge:

- A leak of buffers: they get allocated but not freed.
- Buffers get released twice.

In my private copy I have even added two hidden fields: the places where a buffer was allocated, and where it was released. I don't think that this is useful as a general feature.

Functions like `bIsValidNetworkDescriptor()` are very useful and it would be nice if such functions are available in both files.

Please ping me if you want to have the changes tested.

holden-zenith:
@htibosch if you would like to test it that would be great. I've run them both on my STM32F7 so far. If you have any suggestions or changes let me know.

amazonKamath:
> @htibosch if you would like to test it that would be great. I've run them both on my STM32F7 so far. If you have any suggestions or changes let me know.

amazonKamath:
@holden-zenith we will review your PR next week and provide further feedback. Apologies for the delay, we are in the middle of pushing a release candidate on [FreeRTOS:dev/IPv6_integration](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/dev/IPv6_integration)

PR #754, opened 2023-03-06
Allow use of loopback addresses in IP stack (127.0.0.0/8)

adam-stamand:
Allow use of loopback addresses in IP stack (127.0.0.0/8)

Description
-----------
The use of loopback addresses (defined as IP range 127.0.0.0/8) can be useful for development, testing, and even general use.

Currently the FreeRTOS IP stack filters out these addresses. This PR modifies the filtering such that:
* the IP filter will allow addresses in the loopback range
* ARP will return cache hit if the destination address is in loopback range
* Checking if ARP resolution is required will return false if destination address is in loopback range

Test Steps
-----------
Unit tests added to IP and ARP

Checklist:
----------

Related Issue
-----------
N/A

### DIFF
--- a/test/unit-test/FreeRTOS_IP/FreeRTOS_IP_utest.c
+++ b/test/unit-test/FreeRTOS_IP/FreeRTOS_IP_utest.c
@@ -2249,6 +2249,41 @@ void test_prvAllowIPPacket_IncorrectProtocolChecksum( void )
     TEST_ASSERT_EQUAL( eReleaseBuffer, eResult );
 }
 
+void test_prvAllowIPPacket_LoopbackDest( void )
+{
+    eFrameProcessingResult_t eResult;
+    IPPacket_t * pxIPPacket;
+    NetworkBufferDescriptor_t * pxNetworkBuffer, xNetworkBuffer;
+    UBaseType_t uxHeaderLength = 0;
+    uint8_t ucEthBuffer[ ipconfigTCP_MSS ];
+    IPHeader_t * pxIPHeader;
+
+    memset( ucEthBuffer, 0, ipconfigTCP_MSS );
+
+    pxNetworkBuffer = &xNetworkBuffer;
+    pxNetworkBuffer->pucEthernetBuffer = ucEthBuffer;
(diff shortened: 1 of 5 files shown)


test/unit-test/FreeRTOS_IP/FreeRTOS_IP_utest.c
```diff
@@ -2249,6 +2249,41 @@ void test_prvAllowIPPacket_IncorrectProtocolChecksum( void )
...
+    pxNetworkBuffer = &xNetworkBuffer;
+    pxNetworkBuffer->pucEthernetBuffer = ucEthBuffer;
+    pxIPPacket = ( IPPacket_t * ) pxNetworkBuffer->pucEthernetBuffer;
+    pxIPHeader = &( pxIPPacket->xIPHeader );
+
+    *ipLOCAL_IP_ADDRESS_POINTER = 0xFFFFFFFF;
+
+    pxIPHeader->ucVersionHeaderLength = 0x45;
+
+    pxIPHeader->ulDestinationIPAddress = ipLOOPBACK_ADDRESS;
```
Skptak, line 2272: Small nit, but the majority of these test cases set this to `pxIPHeader->ulDestinationIPAddress = *ipLOCAL_IP_ADDRESS_POINTER;` maybe consider assigning `ipLOCAL_IP_ADDRESS_POINTER` to  `ipLOOPBACK_ADDRESS` so this line can remain the same?

adam-stamand: Thanks for the feedback!

I get what you mean. The reason I'm learning towards how I have it is because `ipLOCAL_IP_ADDRESS_POINTER` refers to the host IP address, while `ipLOOPBACK_ADDRESS` refers specifically to the destination loopback address.

Put another way, hosts will never have an assigned address of `ipLOOPBACK_ADDRESS` (reserved by standard), so I think testing with a local address different from the loopback address makes sense. That said, totally fine changing it if you think best; I've run it both ways and the test passes.

Skptak (approved):

n9wxu (approved):

PR #765, opened 2023-03-09
Read mac address using FreeRTOS_GetMACAddress() rather than using the…

sayyadumar:
Read mac address using FreeRTOS_GetMACAddress()

Description
-----------
Read the mac address after it has been initilized by the FreeRTOS IP Stack, rather than from defines as the mac address is usually read from the EEPROM, and it might be different to the mac address in the defines, especially in production environments


htibosch (approved):
Makes sense to me, I approve it.
Thank you

paulbartell (approved):

AniruddhaKanhere (approved):
Hello @sayyadumar,

Thank you for this PR, it makes sense to use the MAC address with which the IP stack is configured.

I am approving this.

Cheers

PR #774, opened 2023-03-10
Add Nxp1060 network interface

AniruddhaKanhere:
Description
-----------
Add a FreeRTOS+TCP compatible network interface for this board: https://github.com/FreeRTOS/iot-reference-nxp-rt1060

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

Log
------
[DHCP_Check_With_Cable_Disconnects](https://user-images.githubusercontent.com/60444055/225434073-61802d0d-d93b-411f-a638-8844216327e6.jpeg)

### DIFF
--- a/source/portable/NetworkInterface/NXP1060/NetworkInterface.c
+++ b/source/portable/NetworkInterface/NXP1060/NetworkInterface.c
@@ -224,11 +224,15 @@ static void ethernet_callback( ENET_Type * base,
 
 static void prvProcessFrame( int length );
 
-static status_t xSetupPHY( phy_config_t * pxConfig, EMACState_t * peEMACState );
+static status_t xSetupPHY( phy_config_t * pxConfig,
+                           EMACState_t * peEMACState );
 
-static status_t xWaitPHY( phy_config_t * pxConfig, EMACState_t * peEMACState );
+static status_t xWaitPHY( phy_config_t * pxConfig,
+                          EMACState_t * peEMACState );
 
-static status_t xEMACInit( EMACState_t * peEMACState, phy_speed_t speed, phy_duplex_t duplex );
+static status_t xEMACInit( EMACState_t * peEMACState,
+                           phy_speed_t speed,
+                           phy_duplex_t duplex );
 
 BaseType_t xNetworkInterfaceInitialise( void )
 {
@@ -250,13 +254,13 @@ BaseType_t xNetworkInterfaceInitialise( void )
     switch( eEMACState )
     {
         case xEMAC_SetupPHY:
-        	xStatus = xSetupPHY( &xConfig, &eEMACState );
+            xStatus = xSetupPHY( &xConfig, &eEMACState );
 
-        	if( xStatus != kStatus_Success )
-        	{
-        		eEMACState = xEMAC_Fatal;
-        		break;
-        	}
+            if( xStatus != kStatus_Success )
+            {
+                eEMACState = xEMAC_Fatal;
+                break;
+            }
 
         /* Fall through. */
         case xEMAC_WaitPHY:
@@ -270,10 +274,10 @@ BaseType_t xNetworkInterfaceInitialise( void )
             }
 
             if( xStatus != kStatus_Success )
-			{
-            	eEMACState = xEMAC_Fatal;
-				break;
-			}
+            {
+                eEMACState = xEMAC_Fatal;
+                break;
+            }
 
         /* Fall through. */
         case xEMAC_Init:
@@ -282,48 +286,48 @@ BaseType_t xNetworkInterfaceInitialise( void )
 
             if( xFirstCall == pdTRUE )
             {
-				if( xStatus == kStatus_Success )
-				{
-					/* The link is now up. */
-					bGlobalLinkStatus = true;
+                if( xStatus == kStatus_Success )
+                {
+                    /* The link is now up. */
+                    bGlobalLinkStatus = true;
 
-					/* The handler task is created at the highest possible priority to
-					 * ensure the interrupt handler can return directly to it. */
-					xTaskCreated = xTaskCreate( prvEMACHandlerTask,
-										   "EMAC-Handler",
-										   configMINIMAL_STACK_SIZE * 3,
-										   NULL,
-										   configMAX_PRIORITIES - 1,
-										   &receiveTaskHandle );
+                    /* The handler task is created at the highest possible priority to
+                     * ensure the interrupt handler can return directly to it. */
+                    xTaskCreated = xTaskCreate( prvEMACHandlerTask,
+                                                "EMAC-Handler",
+                                                configMINIMAL_STACK_SIZE * 3,
+                                                NULL,
+                                                configMAX_PRIORITIES - 1,
+                                                &receiveTaskHandle );
 
-					if( ( receiveTaskHandle == NULL ) || ( xTaskCreated != pdPASS ) )
-					{
-						FreeRTOS_printf( ( "Failed to create the handler task." ) );
-						eEMACState = xEMAC_Fatal;
-						break;
-					}
+                    if( ( receiveTaskHandle == NULL ) || ( xTaskCreated != pdPASS ) )
+                    {
+                        FreeRTOS_printf( ( "Failed to create the handler task." ) );
+                        eEMACState = xEMAC_Fatal;
+                        break;
+                    }
 
-					/* Enable the interrupt and set its priority to the minimum
-					 * interrupt priority.  */
-					NVIC_SetPriority( ENET_IRQn, ENET_PRIORITY );
-					NVIC_EnableIRQ( ENET_IRQn );
+                    /* Enable the interrupt and set its priority to the minimum
+                     * interrupt priority.  */
+                    NVIC_SetPriority( ENET_IRQn, ENET_PRIORITY );
+                    NVIC_EnableIRQ( ENET_IRQn );
 
-					FreeRTOS_printf( ( "Driver ready for use." ) );
+                    FreeRTOS_printf( ( "Driver ready for use." ) );
 
-					eEMACState = xEMAC_Ready;
+                    eEMACState = xEMAC_Ready;
 
-					/* After this, the task should not be created. */
-					xFirstCall = pdFALSE;
-				}
-				else
-				{
-					eEMACState = xEMAC_Fatal;
-					break;
-				}
+                    /* After this, the task should not be created. */
+                    xFirstCall = pdFALSE;
+                }
+                else
+                {
+                    eEMACState = xEMAC_Fatal;
+                    break;
+                }
             }
             else
             {
-            	eEMACState = xEMAC_Ready;
+                eEMACState = xEMAC_Ready;
             }
 
         /* Fall through. */
@@ -347,9 +351,9 @@ BaseType_t xNetworkInterfaceOutput( NetworkBufferDescriptor_t * const pxNetworkB
 
     if( bGlobalLinkStatus == true )
     {
-    	/* ENET_SendFrame copies the data before sending it. Therefore, the network buffer can
-    	 * be released without worrying about the buffer memory being used by the ENET_SendFrame
-    	 * function. */
+        /* ENET_SendFrame copies the data before sending it. Therefore, the network buffer can
+         * be released without worrying about the buffer memory being used by the ENET_SendFrame
+         * function. */
         result = ENET_SendFrame( ethernetifLocal->base,
                                  &ethernetifLocal->handle,
                                  pxNetworkBuffer->pucEthernetBuffer,
@@ -395,21 +399,21 @@ static void prvEMACHandlerTask( void * parameter )
             if( readStatus == kStatus_Success )
             {
                 if( LinkUp == pdFALSE )
-				{
-					/* The link is down. */
-					bGlobalLinkStatus = false;
-					/* We need to setup the PHY again. */
-					eEMACState = xEMAC_SetupPHY;
+                {
+                    /* The link is down. */
+                    bGlobalLinkStatus = false;
+                    /* We need to setup the PHY again. */
+                    eEMACState = xEMAC_SetupPHY;
 
-					FreeRTOS_printf( ( "Link down!" ) );
+                    FreeRTOS_printf( ( "Link down!" ) );
 
-					xSendEventStructToIPTask( &xNetworkEventDown, 0U );
-				}
-				else
-				{
-					/* The link is still up. */
-					bGlobalLinkStatus = true;
-				}
+                    xSendEventStructToIPTask( &xNetworkEventDown, 0U );
+                }
+                else
+                {
+                    /* The link is still up. */
+                    bGlobalLinkStatus = true;
+                }
             }
         }
         else
@@ -540,44 +544,45 @@ static void prvProcessFrame( int length )
  *
  * @return kStatus_Success if the PHY was initialized; error code otherwise.
  */
-static status_t xSetupPHY( phy_config_t * pxConfig, EMACState_t * peEMACState )
+static status_t xSetupPHY( phy_config_t * pxConfig,
+                           EMACState_t * peEMACState )
 {
-	status_t xStatus;
+    status_t xStatus;
 
-	/* Make sure that current state is correct. */
-	configASSERT( *peEMACState == xEMAC_SetupPHY );
+    /* Make sure that current state is correct. */
+    configASSERT( *peEMACState == xEMAC_SetupPHY );
 
-	ethernetifLocal->RxBuffDescrip = &( rxBuffDescrip_0[ 0 ] );
-	ethernetifLocal->TxBuffDescrip = &( txBuffDescrip_0[ 0 ] );
-	ethernetifLocal->RxDataBuff = &( rxDataBuff_0[ 0 ] );
-	ethernetifLocal->TxDataBuff = &( txDataBuff_0[ 0 ] );
-	ethernetifLocal->base = ( void * ) ENET_BASE;
+    ethernetifLocal->RxBuffDescrip = &( rxBuffDescrip_0[ 0 ] );
+    ethernetifLocal->TxBuffDescrip = &( txBuffDescrip_0[ 0 ] );
+    ethernetifLocal->RxDataBuff = &( rxDataBuff_0[ 0 ] );
+    ethernetifLocal->TxDataBuff = &( txDataBuff_0[ 0 ] );
+    ethernetifLocal->base = ( void * ) ENET_BASE;
 
-	/* Set the clock frequency. */
-	mdioHandle.resource.csrClock_Hz = EXAMPLE_CLOCK_FREQ;
-	mdioHandle.resource.base = ( void * ) ENET_BASE;
+    /* Set the clock frequency. */
+    mdioHandle.resource.csrClock_Hz = EXAMPLE_CLOCK_FREQ;
+    mdioHandle.resource.base = ( void * ) ENET_BASE;
 
-	/* Set the configuration to auto-negotiate; set the phy to full duplex mode; the phy's address; and the speed. */
-	pxConfig->autoNeg = pdTRUE;
-	pxConfig->duplex = kPHY_FullDuplex;
-	pxConfig->phyAddr = BOARD_ENET0_PHY_ADDRESS;
-	pxConfig->speed = kPHY_Speed100M;
+    /* Set the configuration to auto-negotiate; set the phy to full duplex mode; the phy's address; and the speed. */
+    pxConfig->autoNeg = pdTRUE;
+    pxConfig->duplex = kPHY_FullDuplex;
+    pxConfig->phyAddr = BOARD_ENET0_PHY_ADDRESS;
+    pxConfig->speed = kPHY_Speed100M;
 
-	FreeRTOS_printf( ( "Starting PHY initialization.\r\n" ) );
+    FreeRTOS_printf( ( "Starting PHY initialization.\r\n" ) );
 
-	xStatus = PHY_Init( &phyHandle, pxConfig );
+    xStatus = PHY_Init( &phyHandle, pxConfig );
 
-	if( xStatus != kStatus_Success )
-	{
-		FreeRTOS_printf( ( "Failed to initialize the PHY..." ) );
-		*peEMACState = xEMAC_Fatal;
-	}
-	else
-	{
-		*peEMACState = xEMAC_WaitPHY;
-	}
+    if( xStatus != kStatus_Success )
+    {
+        FreeRTOS_printf( ( "Failed to initialize the PHY..." ) );
+        *peEMACState = xEMAC_Fatal;
+    }
+    else
+    {
+        *peEMACState = xEMAC_WaitPHY;
+    }
 
-	return xStatus;
+    return xStatus;
 }
 
 /*
@@ -592,99 +597,100 @@ static status_t xSetupPHY( phy_config_t * pxConfig, EMACState_t * peEMACState )
  *
  * @return kStatus_Success if the PHY was initialized; error code otherwise.
  */
-static status_t xWaitPHY( phy_config_t * pxConfig, EMACState_t * peEMACState )
+static status_t xWaitPHY( phy_config_t * pxConfig,
+                          EMACState_t * peEMACState )
 {
-	status_t xStatus;
-	bool LinkUp;
-	bool autoNegotiationComplete;
-	uint8_t ucCounter = 0;
+    status_t xStatus;
+    bool LinkUp;
+    bool autoNegotiationComplete;
+    uint8_t ucCounter = 0;
 
-	/* Make sure that current state is correct. */
-	configASSERT( *peEMACState == xEMAC_WaitPHY );
+    /* Make sure that current state is correct. */
+    configASSERT( *peEMACState == xEMAC_WaitPHY );
 
-	do
-	{
-		xStatus = PHY_GetAutoNegotiationStatus( &phyHandle, &autoNegotiationComplete );
+    do
+    {
+        xStatus = PHY_GetAutoNegotiationStatus( &phyHandle, &autoNegotiationComplete );
 
-		if( autoNegotiationComplete == true )
-		{
-			break;
-		}
+        if( autoNegotiationComplete == true )
+        {
+            break;
+        }
 
-		/* Try for only a limited number of times. */
-		if( ucCounter++ > MAX_AUTONEG_FAILURE_COUNT )
-		{
-			break;
-		}
+        /* Try for only a limited number of times. */
+        if( ucCounter++ > MAX_AUTONEG_FAILURE_COUNT )
+        {
+            break;
+        }
 
-		vTaskDelay( pdMS_TO_TICKS( 10 ) );
-	}
-	while( xStatus == kStatus_Success );
+        vTaskDelay( pdMS_TO_TICKS( 10 ) );
+    }
+    while( xStatus == kStatus_Success );
 
-	if( autoNegotiationComplete == false )
-	{
-		FreeRTOS_printf( ( "Failed to complete auto-negotiation. Using default values." ) );
+    if( autoNegotiationComplete == false )
+    {
+        FreeRTOS_printf( ( "Failed to complete auto-negotiation. Using default values." ) );
 
-		pxConfig->autoNeg = pdFALSE;
-		pxConfig->duplex = kPHY_FullDuplex;
-		pxConfig->phyAddr = BOARD_ENET0_PHY_ADDRESS;
-		pxConfig->speed = kPHY_Speed100M;
+        pxConfig->autoNeg = pdFALSE;
+        pxConfig->duplex = kPHY_FullDuplex;
+        pxConfig->phyAddr = BOARD_ENET0_PHY_ADDRESS;
+        pxConfig->speed = kPHY_Speed100M;
 
-		xStatus = PHY_Init( &phyHandle, pxConfig );
+        xStatus = PHY_Init( &phyHandle, pxConfig );
 
-		if( xStatus != kStatus_Success )
-		{
-			FreeRTOS_printf( ( "Failed to re-initialize the PHY with default configs!" ) );
-			*peEMACState = xEMAC_Fatal;
-		}
-		else
-		{
-			FreeRTOS_printf( ( "PHY re-initialization successful." ) );
-		}
-	}
-	else
-	{
-		FreeRTOS_printf( ( "Auto-negotiation complete.\r\n" ) );
-	}
+        if( xStatus != kStatus_Success )
+        {
+            FreeRTOS_printf( ( "Failed to re-initialize the PHY with default configs!" ) );
+            *peEMACState = xEMAC_Fatal;
+        }
+        else
+        {
+            FreeRTOS_printf( ( "PHY re-initialization successful." ) );
+        }
+    }
+    else
+    {
+        FreeRTOS_printf( ( "Auto-negotiation complete.\r\n" ) );
+    }
 
-	if( xStatus == kStatus_Success )
-	{
-		/* Reset the counter for next use. */
-		ucCounter = 0;
+    if( xStatus == kStatus_Success )
+    {
+        /* Reset the counter for next use. */
+        ucCounter = 0;
 
-		do
-		{
-			xStatus = PHY_GetLinkStatus( &phyHandle, &LinkUp );
+        do
+        {
+            xStatus = PHY_GetLinkStatus( &phyHandle, &LinkUp );
 
-			if( LinkUp == true )
-			{
-				break;
-			}
+            if( LinkUp == true )
+            {
+                break;
+            }
 
-			vTaskDelay( pdMS_TO_TICKS( 1000 ) );
+            vTaskDelay( pdMS_TO_TICKS( 1000 ) );
 
-			/* Try for only a limited number of times. */
-			if( ucCounter++ > MAX_AUTONEG_FAILURE_COUNT )
-			{
-				break;
-			}
-		}
-		while( xStatus == kStatus_Success );
+            /* Try for only a limited number of times. */
+            if( ucCounter++ > MAX_AUTONEG_FAILURE_COUNT )
+            {
+                break;
+            }
+        }
+        while( xStatus == kStatus_Success );
 
-		if( LinkUp == false )
-		{
-			FreeRTOS_printf( ( "Failed to get the link up." ) );
-			xStatus = kStatus_Fail;
-		}
-		else
-		{
-			/* Success in auto-negotiation and the link is up. */
-			FreeRTOS_printf( ( "Link up." ) );
-			*peEMACState = xEMAC_Init;
-		}
-	}
+        if( LinkUp == false )
+        {
+            FreeRTOS_printf( ( "Failed to get the link up." ) );
+            xStatus = kStatus_Fail;
+        }
+        else
+        {
+            /* Success in auto-negotiation and the link is up. */
+            FreeRTOS_printf( ( "Link up." ) );
+            *peEMACState = xEMAC_Init;
+        }
+    }
 
-	return xStatus;
+    return xStatus;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+                               enet_frame_info_t *frameInfo,
+                               void *userData );
+
+static void prvProcessFrame( int length );
+
+BaseType_t xNetworkInterfaceInitialise( void )
+{
+	status_t xStatus;
+	phy_config_t xConfig = { 0 };
+	BaseType_t xResult = pdPASS;
```
htibosch, line 227: I would choose `pdFAIL` as a default value, because only one branch leads to `pdPASS`.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+    bool autoNegotiationComplete = false;
+    bool LinkUp = false;
+
+    configASSERT( FSL_FEATURE_ENET_QUEUE == 1 );
+
+    #if defined( ENET_ENHANCEDBUFFERDESCRIPTOR_MODE )
+        configASSERT( pdFALSE );
+	#endif
+
+    ethernetifLocal->RxBuffDescrip = &( rxBuffDescrip_0[ 0 ] );
```
htibosch, line 245: Mind you that `xNetworkInterfaceInitialise()` may get called many times, while this initialisation only needs to be done a single time. When it fails, `xNetworkInterfaceInitialise()` should return `pdFAIL` ever after.
I like [this example](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c#L197) in the driver for Ultrascale, it passes through these stages:
~~~
typedef enum xEMAC_STATE
{
    xEMAC_Init,
    xEMAC_SetupPHY,
    xEMAC_WaitPHY,
    xEMAC_Ready,
    xEMAC_Fatal,
} EMACState_t;
~~~
After `xEMAC_Init`, the new stage will either be `xEMAC_SetupPHY` or `xEMAC_Fatal`.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+
+			vTaskDelay( pdMS_TO_TICKS( 10 ) );
+		}
+		while( readStatus == kStatus_Success );
+	}
+
+
+	if( xResult == pdPASS )
+	{
+		if( autoNegotiationComplete == false )
```
htibosch, line 298: Were you able to this this scenario? Auto-negotiation fails and a manual configuration succeeds?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+		do
+		{
+			readStatus = PHY_GetLinkStatus( &phyHandle, &LinkUp );
+
+			if( LinkUp == true )
+			{
+				break;
+			}
+
+			vTaskDelay( pdMS_TO_TICKS( 1000 ) );
```
htibosch, line 330: Calling vTaskDelay() is no problem, as there is only one interface. In the /multi library you could return pdFAIL and wait to be called back again.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+	}
+
+	if( xResult != pdFAIL )
+	{
+	    /* We were able to get the link up. Set up all other required things - such as DMA,
+		 * network output etc. */
+
+		/* The handler task is created at the highest possible priority to
+		 * ensure the interrupt handler can return directly to it. */
+		xResult = xTaskCreate( prvEMACHandlerTask,
```
htibosch, line 369: The program might get here more than one time and create a new instance of the task, don't you think?
I would recommend to postpone creating the task until the moment that the link-status is high, and all has been initialise. If not, the task starts running (at a higher priority than the IP-task) and work with the EMAC.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+			    				 pxNetworkBuffer->pucEthernetBuffer,
+				    			 pxNetworkBuffer->xDataLength,
+					    		 0,
+						    	 false,
+							     NULL );
+
+		switch( result )
+		{
+		    case kStatus_ENET_TxFrameBusy:
+		    	configPRINTF( ( "Failed to send the frame - driver busy!\r\n" ) );
```
htibosch, line 478: We tend to use `FreeRTOS_printf()` for logging in the +TCP library.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+	BaseType_t xReturn = pdFAIL;
+	bool LinkUp = false;
+	status_t readStatus;
+
+	readStatus = PHY_GetLinkStatus( &phyHandle, &LinkUp );
+
+	if( ( readStatus == kStatus_Success ) &&
+		( LinkUp == true ) )
+	{
+		result = ENET_SendFrame( ethernetifLocal->base,
```
htibosch, line 467: Does `ENET_SendFrame()` wait until the frame has been sent?
Or if not, does it send a copy of the frame?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,620 @@
...
+
+BaseType_t xNetworkInterfaceOutput( NetworkBufferDescriptor_t * const pxNetworkBuffer,
+                                    BaseType_t xReleaseAfterSend )
+{
+	status_t result;
+	BaseType_t xReturn = pdFAIL;
+	bool LinkUp = false;
+	status_t readStatus;
+
+	readStatus = PHY_GetLinkStatus( &phyHandle, &LinkUp );
```
htibosch, line 462: The function `PHY_GetLinkStatus()` is not visible in this source, but it looks like it is polling the PHY. It would be better to check the PHY at a quiet moment and here use a variable that tells if the link status is still high.
Polling the PHY, accessing the MII interface can be disturbing when receiving frames.
And also: `prvEMACHandlerTask()` is the owner of the EMAC and the PHY. I'm not sure if a function like `PHY_GetLinkStatus()` thread-proof?

htibosch (commented):
Hi @AniruddhaKanhere, thank you for this new driver. Have you already done some testing? Does it work?
Here below some comments. Most importantly about turning `xNetworkInterfaceInitialise()` into a state-machine.

AniruddhaKanhere:
> Hi @AniruddhaKanhere, thank you for this new driver. Have you already done some testing? Does it work?

Hello @htibosch, yes it does work.
I bombarded the device with TCP packets from my computer. I also parallelly ran a script which would ping the device a 1000 times with 1400 bytes of data. And the device handled all of it very well.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,666 @@
...
+            xConfig.phyAddr = BOARD_ENET0_PHY_ADDRESS;
+            xConfig.speed = kPHY_Speed100M;
+
+            FreeRTOS_printf( ( "Starting PHY initialization.\r\n" ) );
+
+            xStatus = PHY_Init( &phyHandle, &xConfig );
+
+            if( xStatus != kStatus_Success )
+            {
+                FreeRTOS_printf( ( "Failed to initialize the PHY..." ) );
```
htibosch, line 280: Don't you think this is a fatal error? Or can it succeed a second time?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,666 @@
...
+            {
+                readStatus = PHY_GetAutoNegotiationStatus( &phyHandle, &autoNegotiationComplete );
+
+                if( autoNegotiationComplete == true )
+                {
+                    break;
+                }
+
+                /* Try for only a limited number of times. */
+                if( ucCounter++ > MAX_AUTONEG_FAILURE_COUNT )
```
htibosch, line 300: `MAX_AUTONEG_FAILURE_COUNT = 5`
When you find the time it would be interesting to see how many loops are actually needed.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,666 @@
...
+            do
+            {
+                readStatus = PHY_GetLinkStatus( &phyHandle, &LinkUp );
+
+                if( LinkUp == true )
+                {
+                    break;
+                }
+
+                vTaskDelay( pdMS_TO_TICKS( 1000 ) );
```
htibosch, line 347: Make it a bit shorter than a whole second? Maybe 50 ms and doing more loops?

htibosch (approved):
That looks very good to me, and I approve it.
iperf3 can be a good tool to test it.
I'm also curious of the behaviour when you unplug it for e.g. 1 second and then reconnect it to the LAN.
Thank you @AniruddhaKanhere

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,807 @@
...
+
+            if( xStatus != kStatus_Success )
+            {
+                eEMACState = xEMAC_Fatal;
+                break;
+            }
+
+        /* Fall through. */
+        case xEMAC_WaitPHY:
+            FreeRTOS_printf( ( "Configuration successful. Waiting for auto-negotiation to complete.\r\n" ) );
```
htibosch, line 267: I like these changes, it is very easy to understand the main logic.
Let's do some tests with DHCP enabled and unplug and plug the DUT.

htibosch (dismissed):
Thank you @AniruddhaKanhere

I approve the PR, it all looks good to me.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,807 @@
...
+    {
+        xStatus = PHY_GetAutoNegotiationStatus( &phyHandle, &autoNegotiationComplete );
+
+        if( autoNegotiationComplete == true )
+        {
+            break;
+        }
+
+        /* Try for only a limited number of times. */
+        if( ucCounter++ > MAX_AUTONEG_FAILURE_COUNT )
```
paulbartell, line 621: MAX_AUTONEG_FAILURE_COUNT is 5, this runs 5 time.. So auto-negotiation times out it 50ms?

AniruddhaKanhere: FIXED

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: It copies the frame

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: fixed

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Fixed.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: I removed this. Because this would force a default speed if the cable is disconnected and reconnected at just the right moment.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Yes, I switched to using that.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Makes sense, fixed.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: I removed the fatal error state. The init function will be stuck in the last successful state. This allows things like cable disconnects and reconnects to have no impact on the driver.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: I increased this to 40 and the timeout to 100ms. It seems that 4s is the cap on the time required to complete auto-negotiation in many stacks. I followed suit there.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Fixed too

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: I tried and they succeeded. I have attached the log in the PR description.

AniruddhaKanhere:
Thank you @htibosch and @paulbartell for the reviews. I have made all the required changes. Can you please take a look again?

I attached the log to the PR description where I was disconnecting and reconnecting the network cable.

htibosch (dismissed):
( see image at intro PR )
That looks good: unplug, try auto-nego, time-out, try auto-nego, success, new DHCP request.

I approve it. Let's start working with it.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+static phy_handle_t phyHandle = { .phyAddr = 0x00, .mdioHandle = &mdioHandle, .ops = &EXAMPLE_PHY_OPS };
+
+/**
+ * The task-handle for deferred interrupt handler task that processes
+ * incoming packets.
+ */
+static TaskHandle_t receiveTaskHandle = NULL;
+
+static struct ethernetif EthernetInterface1;
+struct ethernetif * ethernetifLocal = &EthernetInterface1;
```
tony-josi-aws, line 213: Is `ethernetifLocal` used by NXP ethernet driver or anywhere outside, if not, can we make it `static`?

shubnil (dismissed):

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
paulbartell: Why copy it to a local variable?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+    if( xStatus != kStatus_Success )
+    {
+        FreeRTOS_printf( ( "Failed to initialize the PHY..." ) );
+    }
+
+    return xStatus;
+}
+
+/**
+ * @brief This function is used wait on the auto-negotiation completion. In case auto-negotiation
```
paulbartell, line 606: Comment seems out of date.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+            NVIC_SetPriority( enetTxIrqId[ instance ], ENET_PRIORITY );
+            break;
+        }
+    }
+
+    configASSERT( instance != ARRAY_SIZE( enetBases ) );
+
+    if( instance == ARRAY_SIZE( enetBases ) )
+    {
+        /**peEMACState = xEMAC_Fatal; */
```
paulbartell, line 766: Comment left over.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+static void prvProcessFrame( int length );
+
+static status_t xSetupPHY( phy_config_t * pxConfig );
+
+static status_t xWaitPHY( phy_config_t * pxConfig );
+
+static status_t xEMACInit( phy_speed_t speed,
+                           phy_duplex_t duplex );
+
+BaseType_t xNetworkInterfaceInitialise( void )
```
paulbartell, line 241: As we chatted about in person, we should make sure to improve this api significantly when supporting multiple interfaces.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+ *                   The state will be updated to the next state depending on the status of
+ *                   the PHY initialization.
+ *
+ * @return kStatus_Success if the PHY was initialized; error code otherwise.
+ */
+static status_t xSetupPHY( phy_config_t * pxConfig )
+{
+    status_t xStatus;
+
+    ethernetifLocal->RxBuffDescrip = &( rxBuffDescrip_0[ 0 ] );
```
paulbartell, line 577: This is configuring the ethernet mac.. not phy.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+        else
+        {
+            FreeRTOS_printf( ( "RX Event not to be considered\n" ) );
+            vReleaseNetworkBufferAndDescriptor( pxBufferDescriptor );
+            /* Not sure if a trace is required.  The stack did not want this message */
+        }
+    }
+    else
+    {
+        FreeRTOS_printf( ( "RX No Buffer Available\n" ) );
```
paulbartell, line 555: "Dropping RX frame from xx:xx:xx:xx:xx:xx" or something similar would be a more useful error message.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+            if( xSendEventStructToIPTask( &xRxEvent, 0 ) == pdFALSE )
+            {
+                vReleaseNetworkBufferAndDescriptor( pxBufferDescriptor );
+                iptraceETHERNET_RX_EVENT_LOST();
+                FreeRTOS_printf( ( "RX Event Lost\n" ) );
+            }
+        }
+        else
+        {
+            FreeRTOS_printf( ( "RX Event not to be considered\n" ) );
```
paulbartell, line 548: "Dropping RX frame from xx:xx:xx:xx:xx:xx to: xx:xx:xx:xx:xx:xx, invalid target mac" or something similar would be a more useful error message here.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+
+    /* Only get interrupt for incoming messages. */
+    config.interrupt = kENET_RxFrameInterrupt;
+    config.callback = ethernet_callback;
+
+    for( instance = 0; instance < ARRAY_SIZE( enetBases ); instance++ )
+    {
+        if( enetBases[ instance ] == ethernetifLocal->base )
+        {
+            NVIC_SetPriority( enetRxIrqId[ instance ], ENET_PRIORITY );
```
paulbartell, line 756: Perhaps an NVIC_SetVector would be useful here?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+    ( void ) base;
+    ( void ) handle;
+    ( void ) frameInfo;
+    ( void ) userData;
+
+    switch( event )
+    {
+        case kENET_RxEvent:
+            vTaskNotifyGiveFromISR( receiveTaskHandle, &needsToYield );
+            portEND_SWITCHING_ISR( needsToYield );
```
paulbartell, line 508: is taskYIELD_FROM_ISR preferred over portEND_SWITCHING_ISR?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+ * since ENET driver requires the starting buffer addresses to be aligned.
+ */
+#if ETH_PAD_SIZE != 0
+    #error "ETH_PAD_SIZE != 0"
+#endif /* ETH_PAD_SIZE != 0 */
+
+/*******************************************************************************
+ * Definitions
+ ******************************************************************************/
+#define ENET_RING_NUM    ( 1 )
```
paulbartell, line 76: Only a single incoming buffer could lead to dropping packets.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+
+/* The length or RX buffer. */
+#ifndef ENET_RXBUFF_SIZE
+    #define ENET_RXBUFF_SIZE    ( ENET_FRAME_MAX_FRAMELEN )
+#endif
+
+/* ENET IRQ priority. Used in FreeRTOS. */
+/* Interrupt priorities. */
+#ifdef __CA7_REV
```
paulbartell, line 85: IMXRT1060 is always a Cortex-M7.. Perhaps this driver should be name something more generic? Perhaps fsl_enet?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+    #ifndef ENET_PRIORITY
+        #define ENET_PRIORITY         ( 6U )
+    #endif
+    #ifndef ENET_1588_PRIORITY
+        #define ENET_1588_PRIORITY    ( 5U )
+    #endif
+#endif /* ifdef __CA7_REV */
+
+/* The number of ENET buffers needed to receive frame of maximum length. */
+#define MAX_BUFFERS_PER_FRAME \
```
paulbartell, line 102: This should always be 1.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+#endif
+
+/* The number of RX buffers. ENET_RXBD_NUM is always held by ENET driver,
+ * so a couple more are needed to pass zero-copy data into lwIP. */
+#ifndef ENET_RXBUFF_NUM
+    #define ENET_RXBUFF_NUM    ( ENET_RXBD_NUM * 2 )
+#endif
+
+/* At least ENET_RXBD_NUM number of buffers is always held by ENET driver for RX.
+ * Some additional buffers are needed to pass at least one frame zero-copy data to lwIP. */
```
paulbartell, line 127: lwIP?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
paulbartell: Nevermind. I see this is the number of rings not number of rx buffers.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+static status_t xWaitPHY( phy_config_t * pxConfig )
+{
+    status_t xStatus;
+    bool LinkUp;
+    bool autoNegotiationComplete;
+    uint8_t ucCounter = 0;
+
+    do
+    {
+        xStatus = PHY_GetAutoNegotiationStatus( &phyHandle, &autoNegotiationComplete );
```
paulbartell, line 626: As we discussed in person yesterday.. This loop should really be reversed.. 

First checking PHY_GetLinkStatus() in case auto-negotiation is disabled. 

Then you could optionally poll PHY_GetAutoNegotiationStatus() if auto-negotiation is enabled for the purpose of providing log messages to make the state of the system clear. Ultimately, the thing that drives this state machine should be only the call to PHY_GetLinkStatus().

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+
+    ( void ) base;
+    ( void ) handle;
+    ( void ) frameInfo;
+    ( void ) userData;
+
+    switch( event )
+    {
+        case kENET_RxEvent:
+            vTaskNotifyGiveFromISR( receiveTaskHandle, &needsToYield );
```
paulbartell, line 507: In a zero-copy driver, this would be the best place to pass the incoming packet buffer to the tcp stack.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+    static const IRQn_Type enetTxIrqId[] = ENET_Transmit_IRQS;
+    /*! @brief Pointers to enet receive IRQ number for each instance. */
+    static const IRQn_Type enetRxIrqId[] = ENET_Receive_IRQS;
+    int i;
+
+    /* Make sure that the channel is full duplex. */
+    configASSERT( duplex == kPHY_FullDuplex );
+
+    /* prepare the buffer configuration. */
+    buffCfg[ 0 ].rxBdNumber = ENET_RXBD_NUM;                  /* Receive buffer descriptor number. */
```
paulbartell, line 722:
```suggestion
    buffCfg[ 0 ].rxBdNumber = ENET_RXBD_NUM;                  /* Number of RX buffer descriptors */
```

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+    BaseType_t xResult = pdFAIL;
+    phy_speed_t speed;
+    phy_duplex_t duplex;
+    BaseType_t xTaskCreated;
+    static BaseType_t xFirstCall = pdTRUE;
+
+    configASSERT( FSL_FEATURE_ENET_QUEUE == 1 );
+
+    #if defined( ENET_ENHANCEDBUFFERDESCRIPTOR_MODE )
+        /* This driver is not to be used with enhanced buffer descriptor mode. */
```
paulbartell, line 254:
```suggestion
        #error "ENET_ENHANCEDBUFFERDESCRIPTOR_MODE is not supported by this driver"
```

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+        configASSERT( pdFALSE == pdTRUE );
```
paulbartell, line 255:
```suggestion
```

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: The vector handler is set inside the call to `ENET_Init`.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Is that macro deprecated by any chance? I cannot find its reference anywhere.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Yes, but currently it is not a zero copy driver and thus we need to invoke a task which can allocate the memory and process the frame.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: I think we can do that later on. Because a generic driver will need more testing and unless that is done, I am hesitant to make it more generic.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: It will be.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Fixed.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+/* A bigger value is chosen so that the previous notification values are
+ * not interfering with the driver ready notifications. */
+#define DRIVER_READY    ( 0x40 )
+#define DRIVER_FATAL    ( DRIVER_READY << 1 )
+
+typedef uint8_t   rx_buffer_t[ SDK_SIZEALIGN( ENET_RXBUFF_SIZE, FSL_ENET_BUFF_ALIGNMENT ) ];
+typedef uint8_t   tx_buffer_t[ SDK_SIZEALIGN( ENET_TXBUFF_SIZE, FSL_ENET_BUFF_ALIGNMENT ) ];
+
+/**
+ * @brief Used to wrap received data in a pbuf to be passed into lwIP
```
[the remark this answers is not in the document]
AniruddhaKanhere: Oops. Fixed.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Fixed. Moved to top and out of this function.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Fixed.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Updated

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: We cannot decipher the MAC as we do not have a free buffer in which to read the frame.
Thus, we just can say that we had to drop the frame.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
paulbartell: s/taskYIELD_FROM_ISR/portYIELD_FROM_ISR/

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,797 @@
...
+                readStatus = PHY_GetLinkStatus( &phyHandle, &LinkUp );
+
+                if( readStatus == kStatus_Success )
+                {
+                    if( LinkUp == pdFALSE )
+                    {
+                        /* The link is down. */
+                        bGlobalLinkStatus = false;
+                        /* We need to setup the PHY again. */
+                        eEMACState = xEMAC_SetupPHY;
```
paulbartell, line 419: There's no need to reconfigure the phy.

```suggestion
                        eEMACState = xEMAC_WaitPHY;
```

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,813 @@
...
+    status_t xStatus;
+    uint32_t instance;
+    static ENET_Type * const enetBases[] = ENET_BASE_PTRS;
+    static const IRQn_Type enetTxIrqId[] = ENET_Transmit_IRQS;
+    /*! @brief Pointers to enet receive IRQ number for each instance. */
+    static const IRQn_Type enetRxIrqId[] = ENET_Receive_IRQS;
+    int i;
+
+    /* Make sure that the channel is full duplex. */
+    configASSERT( duplex == kPHY_FullDuplex );
```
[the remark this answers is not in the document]
AniruddhaKanhere: It is, but the driver is not written with that in mind.
It requires some magic sauce to work :)

Thus, unless we add that, I would like to keep the configASSERT in place.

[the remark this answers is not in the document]
AniruddhaKanhere: Ignore that, it seems that I was wrong :)
Have fixed that.

paulbartell (approved):
Still a bit overly complicated in the xEMACInit function, but good enough for now i guess.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,810 @@
...
+    {
+        vReleaseNetworkBufferAndDescriptor( pxNetworkBuffer );
+    }
+
+    return xReturn;
+}
+
+static void prvEMACHandlerTask( void * parameter )
+{
+    bool LinkUp = false;
```
kstribrnAmzn, line 412: Nit-pick: Why not `bLinkUp`?

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,810 @@
...
+                if( readStatus == kStatus_Success )
+                {
+                    if( LinkUp == pdFALSE )
+                    {
+                        /* The link is down. */
+                        bGlobalLinkStatus = false;
+                        /* We need to setup the PHY again. */
+                        eEMACState = xEMAC_WaitPHY;
+
+                        FreeRTOS_printf( ( "Link down!" ) );
```
kstribrnAmzn, line 438: Q: Should this instead be a log statement?

kstribrnAmzn (approved):

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: In FreeRTOS+TCP, there is no `log_print` or similar logging macros. In +TCP, we just use `FreeRTOS_printf` and `FreeRTOS_debug_printf`.

Out of those, I chose the non-debug version since a link going down seems to me like a major event which the user should know about.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,810 @@
...
+#if ENET_RXBUFF_NUM < ( ENET_RXBD_NUM + MAX_BUFFERS_PER_FRAME )
+    #error "ENET_RXBUFF_NUM < (ENET_RXBD_NUM + MAX_BUFFERS_PER_FRAME)"
+#endif
+
+/* The number of buffer descriptors in ENET TX ring. */
+#ifndef ENET_TXBD_NUM
+    #define ENET_TXBD_NUM    ( 3 )
+#endif
+
+/* Set the timeout values such that the total timeout adds up to 4000ms. */
```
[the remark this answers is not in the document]
AniruddhaKanhere: I think it is just a de-facto standard. UNIX uses this and thus I copied this value :)

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,810 @@
...
+#define ENET_RING_NUM    ( 1 )
+
+/* The length or RX buffer. */
+#ifndef ENET_RXBUFF_SIZE
+    #define ENET_RXBUFF_SIZE    ( ENET_FRAME_MAX_FRAMELEN )
+#endif
+
+/* ENET IRQ priority. Used in FreeRTOS. */
+/* Interrupt priorities. */
+#ifdef __CA7_REV
```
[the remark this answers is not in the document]
AniruddhaKanhere: I believe it signals ARM Cortex A7.
In this (NXP MIMXRT1060) case however, the processor is Cortex M7

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
```diff
@@ -0,0 +1,810 @@
...
+
+/* ENET clock frequency. */
+#define EXAMPLE_CLOCK_FREQ    CLOCK_GetFreq( kCLOCK_IpgClk )
+
+/*
+ * Padding of ethernet frames has to be disabled for zero-copy functionality
+ * since ENET driver requires the starting buffer addresses to be aligned.
+ */
+#if ETH_PAD_SIZE != 0
+    #error "ETH_PAD_SIZE != 0"
```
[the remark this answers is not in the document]
AniruddhaKanhere: Makes sense, I will add this as a follow up PR.

source/portable/NetworkInterface/NXP1060/NetworkInterface.c
AniruddhaKanhere: Oops, makes sense, I will add this as a follow up PR.

Issue #771, opened 2023-03-09
[BUG] Able to define ipconfigUSE_IPv4 & ipconfigUSE_IPv6 to 0


holden-zenith:
There should be an error generated if both of these are defined to 0. https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/0b67e4551f31529a41e7c3f37a5c281bb1db7da6/source/include/FreeRTOSIPConfigDefaults.h#L235

paulbartell:
Looks like this Issue was addressed. Feel free to reopen if not.

PR #776, opened 2023-03-10
Enable both ipconfigUSE_IPv6 ipconfigUSE_IPv4 while building

tony-josi-aws:
Description
-----------
This PR makes it mandatory to enable ipconfigUSE_IPv4 and ipconfigUSE_IPv6 flags to be enabled for build to progress, as separation for both IPv4 and IPv6 is work in progress.

Test Steps
-----------
Unit tests  are validated.

Checklist:
----------

Related Issue
-----------
Issue #771 

### DIFF
--- a/source/include/FreeRTOSIPConfigDefaults.h
+++ b/source/include/FreeRTOSIPConfigDefaults.h
@@ -241,6 +241,11 @@
     #define ipconfigUSE_IPv6    ( 1 )
 #endif
 
+#if (ipconfigUSE_IPv4 != 1) || (ipconfigUSE_IPv6 != 1)
+    #error "Build separation for both IPv4 and IPv6 is work in progress. \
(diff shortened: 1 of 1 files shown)


amazonKamath (approved):

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -241,6 +241,11 @@
     #define ipconfigUSE_IPv6    ( 1 )
 #endif
 
+#if ( ipconfigUSE_IPv4 != 1 ) || ( ipconfigUSE_IPv6 != 1 )
```
sirnish, line 244: What is the default value for these in case one want to use it with older applications? Does it mean that library will not be backward compatible if one does not define these?

shubnil (approved):

moninom1 (approved):

Issue #558, opened 2022-09-25
[BUG] Fix Clang/GCC compiler warnings.


phelter:
**Describe the bug**
Cleanup the Clang and GNU compiler warnings when `-Wall`, `-Weverything`, and `-Wpedantic` are used.

Start using clang for checking doxygen and other C based linting violations - can use in gitub-actions.

**Target**
- Development board: Linux
- Instruction Set Architecture: Linux
- IDE and version: Visual Code 1.71.1
- Toolchain and version: Clang-13 and GCC 7.5.0

**Host**
- Host OS: Linux
- Version: Ubuntu 18.04

**To Reproduce**
- Based on - https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/557
- Run on Ubuntu 18.04 with Clang 14 and Remove the `target_compile_options(...` could observe failures when it compiles

**Expected behavior**
Code cleaned up and a CI call to Clang and GCC to ensure code quality is adhered to.

**Screenshots**
N/A

**Wireshark logs**
N/A

**Additional context**
The primary reason for doing this is to use Clang and GCC to provide better and more descriptive code.

kstribrnAmzn:
Dumping the warnings I found using the QEMU emulator demo compiled with arm-none-eabi-gcc. This is lazy and I'll need to check further, this is just the quickest way I could get to producing warnings.

Setup:
* Mac Pro, I7, Ventura 13.2
* arm-none-eabi-gcc (Arm GNU Toolchain 11.3.Rel1) 11.3.1 20220712
* Flags used - `-Wall -Wextra -Wpedantic`
```
startup.c: In function 'prvGetRegistersFromStack':
startup.c:83:23: error: variable 'psr' set but not used [-Werror=unused-but-set-variable]
   83 |     volatile uint32_t psr; /* Program status register. */
      |                       ^~~
startup.c:82:23: error: variable 'pc' set but not used [-Werror=unused-but-set-variable]
   82 |     volatile uint32_t pc;  /* Program counter. */
      |                       ^~
startup.c:81:23: error: variable 'lr' set but not used [-Werror=unused-but-set-variable]
   81 |     volatile uint32_t lr;  /* Link register. */
      |                       ^~
startup.c:80:23: error: variable 'r12' set but not used [-Werror=unused-but-set-variable]
   80 |     volatile uint32_t r12;
      |                       ^~~
[...]

PR #798, opened 2023-03-15
Correct GCC warnings

kstribrnAmzn:
Description
-----------
Corrects warnings with current GCC flags
for GCC 7.5.0. The only suppressed warning pertains to function to object pointer conversion which is
required and common for socket callbacks.

Test Steps
-----------
Performed all three variations of the `build-check` builds on an 18.04 ubuntu machine

Checklist:
----------

Will need to sync with library owners. I'm not familiar with running additional tests.

Related Issue
-----------
https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/558

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -1,5 +1,7 @@
 add_library( freertos_plus_tcp STATIC )
 
+set_property(TARGET freertos_plus_tcp PROPERTY C_STANDARD 90)
+
 target_sources( freertos_plus_tcp
   PRIVATE
       include/FreeRTOS_ARP.h
@@ -74,10 +76,7 @@ target_compile_options( freertos_plus_tcp
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-shorten-64-to-32>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-sign-conversion>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-unused-macros>
-    $<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-unused-but-set-variable>
-    $<$<COMPILE_LANG_AND_ID:C,Clang,GNU>:-Wno-unused-parameter>
-    $<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-unused-variable>
-    $<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-pedantic>
+    $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-unused-parameter>
 )
 
 target_link_libraries( freertos_plus_tcp
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -1987,10 +1987,7 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                         /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-111 */
                         /* coverity[misra_c_2012_rule_11_8_violation] */
                         /* coverity[misra_c_2012_rule_11_1_violation] */
-                        #pragma GCC diagnostic push
-                        #pragma GCC diagnostic ignored "-Wpedantic"
                         pxSocket->pxUserWakeCallback = ( SocketWakeupCallback_t ) pvOptionValue;
-                        #pragma GCC diagnostic pop
                         xReturn = 0;
                         break;
                 #endif /* ipconfigSOCKET_HAS_USER_WAKE_CALLBACK */
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -1987,7 +1987,10 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                         /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-111 */
                         /* coverity[misra_c_2012_rule_11_8_violation] */
                         /* coverity[misra_c_2012_rule_11_1_violation] */
+                        #pragma GCC diagnostic push
+                        #pragma GCC diagnostic ignored "-Wpedantic"
                         pxSocket->pxUserWakeCallback = ( SocketWakeupCallback_t ) pvOptionValue;
+                        #pragma GCC diagnostic pop
                         xReturn = 0;
                         break;
                 #endif /* ipconfigSOCKET_HAS_USER_WAKE_CALLBACK */
--- a/source/FreeRTOS_TCP_IP.c
+++ b/source/FreeRTOS_TCP_IP.c
@@ -634,27 +634,36 @@
         /* Function might modify the parameter. */
         NetworkBufferDescriptor_t * pxNetworkBuffer = pxDescriptor;
 
+        /* Map the buffer onto a ProtocolHeaders_t struct for easy access to the fields. */
+
+        ProtocolHeaders_t * pxProtocolHeaders;
+        FreeRTOS_Socket_t * pxSocket;
+        uint16_t ucTCPFlags;
+        uint32_t ulLocalIP;
+        uint16_t usLocalPort;
+        uint16_t usRemotePort;
+        uint32_t ulRemoteIP;
+        uint32_t ulSequenceNumber;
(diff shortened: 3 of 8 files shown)


source/FreeRTOS_Sockets.c
```diff
@@ -1987,7 +1987,10 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                         /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-111 */
                         /* coverity[misra_c_2012_rule_11_8_violation] */
                         /* coverity[misra_c_2012_rule_11_1_violation] */
+                        #pragma GCC diagnostic push
+                        #pragma GCC diagnostic ignored "-Wpedantic"
                         pxSocket->pxUserWakeCallback = ( SocketWakeupCallback_t ) pvOptionValue;
+                        #pragma GCC diagnostic pop
```
kstribrnAmzn, line 1993: Not sure I'm a big fan of this but I had to do this to resolve the final pedantic warning which was around this cast.

I'm open to removing this and disabling pedantic warnings.

Skptak: Wait but it looks like the wpedantic option was removed in the CMakeLists.txt, is it still hitting this?
And I know other compilers will flag pragmas as warnings? Where feel like this might just be trading this warning for another one?

source/FreeRTOS_TCP_IP.c
```diff
@@ -634,28 +634,37 @@
...
+        /* coverity[misra_c_2012_rule_11_3_violation] */
+        pxProtocolHeaders = ( ( ProtocolHeaders_t * )
+                              &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER + xIPHeaderSize( pxNetworkBuffer ) ] ) );
+
+        ucTCPFlags = pxProtocolHeaders->xTCPHeader.ucTCPFlags;
+        usLocalPort = FreeRTOS_htons( pxProtocolHeaders->xTCPHeader.usDestinationPort );
+        usRemotePort = FreeRTOS_htons( pxProtocolHeaders->xTCPHeader.usSourcePort );
+        ulSequenceNumber = FreeRTOS_ntohl( pxProtocolHeaders->xTCPHeader.ulSequenceNumber );
+        ulAckNumber = FreeRTOS_ntohl( pxProtocolHeaders->xTCPHeader.ulAckNr );
+        xResult = pdPASS;
```
AniruddhaKanhere, line 666: All others seem good to me, but can we make this as this was before?
As in declare and set in the same location: `BaseType_t xResult = pdPASS;`

My rationale is that this kind of flag setting is easy to miss when making changes, especially when it is done along with the setting of many variables. And therefore, this might lead to errors in the future.

test/build-combination/Common/FreeRTOSConfig.h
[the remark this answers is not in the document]
AniruddhaKanhere: NVM, found it: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/dcd6e58cec5676231c329e9645590d90eb4c5cba/test/build-combination/Common/main.c#L223 :)

source/CMakeLists.txt
```diff
@@ -74,10 +76,7 @@ target_compile_options( freertos_plus_tcp
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-shorten-64-to-32>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-sign-conversion>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-unused-macros>
-    $<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-unused-but-set-variable>
-    $<$<COMPILE_LANG_AND_ID:C,Clang,GNU>:-Wno-unused-parameter>
-    $<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-unused-variable>
```
AniruddhaKanhere, line 79: Why are we removing these? These seem like good warnings to have (talking about 77 and 79)

source/FreeRTOS_UDP_IP.c
```diff
@@ -310,6 +310,7 @@ BaseType_t xProcessReceivedUDPPacket( NetworkBufferDescriptor_t * pxNetworkBuffe
 {
     BaseType_t xReturn = pdPASS;
     FreeRTOS_Socket_t * pxSocket;
+    UDPPacket_t * pxUDPPacket;
```
AniruddhaKanhere, line 313:
```suggestion
    const UDPPacket_t * pxUDPPacket;
```

source/CMakeLists.txt
AniruddhaKanhere: Sorry, this is removing the suppressing of warnings.

source/portable/BufferManagement/BufferAllocation_2.c
```diff
@@ -73,11 +73,10 @@
     #define ASSERT_CONCAT_( a, b )    a ## b
     #define ASSERT_CONCAT( a, b )     ASSERT_CONCAT_( a, b )
     #define STATIC_ASSERT( e ) \
-    ; enum { ASSERT_CONCAT( assert_line_, __LINE__ ) = 1 / ( !!( e ) ) }
+    enum { ASSERT_CONCAT( assert_line_, __LINE__ ) = 1 / ( !!( e ) ) }
```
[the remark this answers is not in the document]
htibosch: @AniruddhaKanhere wrote:

> I assume this change did not break anything.
> I wonder if this was done to please a specific compiler...
> Do you know the reason @htibosch?

The semicolon isn't necessary. It doesn't change the behaviour.

As you see this is a static assert of the expression `ipconfigETHERNET_MINIMUM_PACKET_BYTES <= baMINIMAL_BUFFER_SIZE`, where `baMINIMAL_BUFFER_SIZE` uses the sizeof operator.

I just tested these static asserts:

~~~
    STATIC_ASSERT( 5 > 4 );
    STATIC_ASSERT( 4 > 4 );  /* error */
    STATIC_ASSERT( 2 > 1 );
    STATIC_ASSERT( 4 > 5 );  /* error */
~~~

The compiler listed 2 failures:

~~~
    enumerator value for 'assert_line_87' is not an integer constant
	enumerator value for 'assert_line_89' is not an integer constant
~~~

`ipconfigETHERNET_MINIMUM_PACKET_BYTES` is the minimum number of bytes in a packet on the wire. When necessary, the packet is extended with zero's in order to get the minimum length.

`baMINIMAL_BUFFER_SIZE` is the minimum number of bytes that are allocated in a network buffer.

In the code, it is assumed that a network buffer always has enough space to store `ipconfigETHERNET_MINIMUM_PACKET_BYTES` bytes.

htibosch (approved):
Thank you very much [Kody](https://github.com/kstribrnAmzn) for these changes.

This function cast indeed triggers warnings and/or errors:

~~~
    typedef void (* SocketWakeupCallback_t)( struct xSOCKET * pxSocket );
    const void * pvOptionValue;
    SocketWakeupCallback_t pxUserWakeCallback;

    pxUserWakeCallback = ( SocketWakeupCallback_t ) pvOptionValue;
~~~

The application hook is set as a socket option. That takes a `const void *` as a parameter. This feature has been there for many years, so I wouldn't dare removing it.

A better option would be to create a separate access function:

~~~
BaseType_t xSetWakeupCallback( Socket_t xSocket, SocketWakeupCallback_t pfCallBack )
{
	BaseType_t xResult = pdFAIL;
	if( xSocketValid( pxSocket ) == pdTRUE )
	{
		xSocket->pxUserWakeCallback = pfCallBack;
		xResult = pdPASS;
	}
	return xResult;
}
~~~
I approve your PR.

AniruddhaKanhere (approved):

archigup (approved):

source/FreeRTOS_Sockets.c
swaldhoer: TI ARM CGT also complains about this. Can we provide a generic macro for this in ``source/portable/Compiler``? Otherwise one can not consume the common files without changing such code pieces.

I will add a separate issue to discuss this.

source/portable/BufferManagement/BufferAllocation_2.c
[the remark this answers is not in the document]
swaldhoer: Just for documentation purposes: TI ARM CGT also complained about the semicolon.

PR #799, opened 2023-03-16
[IPv6] Check dest address in UDP packets with endpoint address.

ActoryOu:
[IPv6] Check dest address in UDP packets with endpoint address.

Description
-----------

Check the destination IP adress in UDP packets with all endpoints' address. Drop the packets with no matched endpoint.

Test Steps
-----------

Run protocol test UDP/Others/003 and UDP/Others/004.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/docs/doxygen/config.doxyfile
+++ b/docs/doxygen/config.doxyfile
@@ -48,7 +48,7 @@ PROJECT_NAME           = FreeRTOS-Plus-TCP
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = V3.0.0
+PROJECT_NUMBER         = V4.0.0-rc1
 
 # Using the PROJECT_BRIEF tag one can provide an optional one line description
 # for a project that appears at the top of each page and should give viewer a
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,5 +1,5 @@
 name : "FreeRTOS-Plus-TCP"
-version: "V3.0.0"
+version: "V4.0.0-rc1"
 description: "Thread safe FreeRTOS TCP/IP stack working on top of the FreeRTOS-Kernel to implement the TCP/IP protocol. Suitable for microcontrollers."
 license: "MIT"
 dependencies:
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -2891,15 +2891,18 @@ static const ListItem_t * pxListFindListItemWithValue( const List_t * pxList,
 /**
  * @brief Find the UDP socket corresponding to the port number.
  *
+ * @param[in] pxRemoteAddress: The remote address with sin_family set.
  * @param[in] uxLocalPort: The port whose corresponding bound UDP socket
  *                         is to be found.
  *
  * @return The socket owning the port if found or else NULL.
  */
-FreeRTOS_Socket_t * pxUDPSocketLookup( UBaseType_t uxLocalPort )
+FreeRTOS_Socket_t * pxUDPSocketLookup( const struct freertos_sockaddr * pxRemoteAddress,
+                                       UBaseType_t uxLocalPort )
 {
     const ListItem_t * pxListItem;
     FreeRTOS_Socket_t * pxSocket = NULL;
+    NetworkEndPoint_t * pxRemoteEndpoint = NULL;
 
     /* Looking up a socket is quite simple, find a match with the local port.
      *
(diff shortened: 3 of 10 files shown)


htibosch (commented):
Thank you @ActoryOu for this PR.

I'll first try to summarise it:

The function `pxUDPSocketLookup()` is called for every **incoming** packet. It looks up the socket that is bound to the destination port number. When found, the UDP (or TCP) packet can be passed to the socket, which will normally result in a successful `recvfrom()` or `recv()`.

Here is, symbolically, the propose change:

~~~diff
FreeRTOS_Socket_t * pxUDPSocketLookup( const struct freertos_sockaddr * pxRemoteAddress,
                                       UBaseType_t uxLocalPort )
{
+    /* Check if we can match remote IP address in any endpoint. */
+    if( pxRemoteAddress->sin_family == ( uint8_t ) FREERTOS_AF_INET6 )
+    {
+        pxRemoteEndpoint = FreeRTOS_FindEndPointOnIP_IPv6( &( pxRemoteAddress->sin_address.xIP_IPv6 ) );
+    }
+    else
+    {
+        pxRemoteEndpoint = FreeRTOS_FindEndPointOnIP_IPv4( pxRemoteAddress->sin_address.ulIP_IPv4, 8 );
+    }
+    if( pxRemoteEndpoint != NULL )
+    {
+        return xSocket;
+    }
}
~~~

In other words, when a packet is handled, there must be an endpoint through which we can reach `pxRemoteAddress`, otherwise we call it a "destination port unreachable".

But, a couple of weeks ago, we have stated that the network interface **must** provide an endpoint that can handle the incoming packet. The network interface finds this endpoint by calling:

~~~
    pxNetworkBuffer->pxInterface = pxMyInterface;
    pxNetworkBuffer->pxEndPoint = FreeRTOS_MatchingEndpoint( pxMyInterface, pucEthernetBuffer );
~~~

We could even do this:

~~~diff
 static void prvHandleEthernetPacket( NetworkBufferDescriptor_t * pxBuffer )
 {
+    configASSERT( pxBuffer->pxEndPoint != NULL );
     prvProcessEthernetPacket( pxBuffer );
 }
~~~

In other words, when `xProcessReceivedUDPPacket()` is called, there is already a proper endpoint in the `pxNetworkBuffer`, and I don't see a reason to look it up again.

But maybe I'm missing something, and if so, please explain.

ActoryOu:
> Thank you @ActoryOu for this PR.
> 
> I'll first try to summarise it:
> 
> The function `pxUDPSocketLookup()` is called for every **incoming** packet. It looks up the socket that is bound to the destination port number. When found, the UDP (or TCP) packet can be passed to the socket, which will normally result in a successful `recvfrom()` or `recv()`.
> 
> Here is, symbolically, the propose change:
> 
> ```diff
> FreeRTOS_Socket_t * pxUDPSocketLookup( const struct freertos_sockaddr * pxRemoteAddress,
>                                        UBaseType_t uxLocalPort )
> {
> +    /* Check if we can match remote IP address in any endpoint. */
> +    if( pxRemoteAddress->sin_family == ( uint8_t ) FREERTOS_AF_INET6 )
> +    {
> +        pxRemoteEndpoint = FreeRTOS_FindEndPointOnIP_IPv6( &( pxRemoteAddress->sin_address.xIP_IPv6 ) );
> +    }
> +    else
> +    {
> +        pxRemoteEndpoint = FreeRTOS_FindEndPointOnIP_IPv4( pxRemoteAddress->sin_address.ulIP_IPv4, 8 );
> +    }
> +    if( pxRemoteEndpoint != NULL )
> +    {
> +        return xSocket;
> +    }
> }
> ```
> 
> In other words, when a packet is handled, there must be an endpoint through which we can reach `pxRemoteAddress`, otherwise we call it a "destination port unreachable".
> 
> But, a couple of weeks ago, we have stated that the network interface **must** provide an endpoint that can handle the incoming packet. The network interface finds this endpoint by calling:
> 
> ```
>     pxNetworkBuffer->pxInterface = pxMyInterface;
>     pxNetworkBuffer->pxEndPoint = FreeRTOS_MatchingEndpoint( pxMyInterface, pucEthernetBuffer );
> ```
> 
> We could even do this:
> 
> ```diff
>  static void prvHandleEthernetPacket( NetworkBufferDescriptor_t * pxBuffer )
>  {
> +    configASSERT( pxBuffer->pxEndPoint != NULL );
>      prvProcessEthernetPacket( pxBuffer );
>  }
> ```
> 
> In other words, when `xProcessReceivedUDPPacket()` is called, there is already a proper endpoint in the `pxNetworkBuffer`, and I don't see a reason to look it up again.
> 
> But maybe I'm missing something, and if so, please explain.

Hi @htibosch,
Thanks for review! From [pxEasyFit](https://github.com/ActoryOu/FreeRTOS-Plus-TCP/blob/dev/IPv6_integration/source/FreeRTOS_Routing.c#L790-L794), which is called by FreeRTOS_MatchingEndpoint(), it matches the endpoint by same MAC address. So here is the condition I met -- the router sends a packet with wrong IP address (or multicast address) and correct MAC address. For this kind of situation, I decided to check the destination again in protocol layer to avoid it.

Thanks.

ActoryOu:
This case is covered by #805. Closing.

PR #805, opened 2023-03-21
[IPv6] Check unspecified/loopback/multicast reserved/matching addresses in prvAllowIPPacketIPv6.

ActoryOu:
[IPv6] Check unspecified/loopback/multicast reserved/matching addresses in prvAllowIPPacketIPv6.

Description
-----------

From [RFC 4291](https://www.rfc-editor.org/rfc/rfc4291.html), some IPv6 addresses are reserved and shouldn't be used. Drop packets when the address met below conditions:
1. Source or destination address have unspecified address (All 0s).
2. Check endpoint's IP address with packet's destination.
3. Allow loopback packets from this node itself only.
4. Check if multicase address isn't valid. (Refer to [sec 2.7](https://www.rfc-editor.org/rfc/rfc4291.html#section-2.7) in RFC4291)
5. Fix coverity warnings.

Test Steps
-----------

Run protocol tests UDP/Others/003, UDP/Others/004, IPv6/datagram/010~017, 019.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IPv6.c
+++ b/source/FreeRTOS_IPv6.c
@@ -69,10 +69,13 @@ const struct xIPv6_Address FreeRTOS_in6addr_loopback = { { 0, 0, 0, 0, 0, 0, 0, 
  */
 static const struct xIPv6_Address xIPv6UnspecifiedAddress = { { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 } };
 
+#if ( ipconfigETHERNET_DRIVER_FILTERS_PACKETS == 0 )
+
 /*
  * Check if the packet is a loopback packet.
  */
-static BaseType_t xIsIPv6Loopback( const IPHeader_IPv6_t * const pxIPv6Header );
+    static BaseType_t xIsIPv6Loopback( const IPHeader_IPv6_t * const pxIPv6Header );
+#endif /* ipconfigETHERNET_DRIVER_FILTERS_PACKETS == 0 */
 
 /**
  * @brief Get the group ID and stored into IPv6_Address_t.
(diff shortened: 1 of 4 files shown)


source/FreeRTOS_IPv6.c
```diff
@@ -64,15 +130,34 @@ const struct xIPv6_Address FreeRTOS_in6addr_loopback = { { 0, 0, 0, 0, 0, 0, 0,
  */
 BaseType_t xIsIPv6Multicast( const IPv6_Address_t * pxIPAddress )
 {
-    BaseType_t xReturn;
+    BaseType_t xReturn = pdFALSE;
```
shubnil, line 133: Instead of xIsIPv6Multicast, xIsIPv6AllowedMulticast seems more apt

source/FreeRTOS_IPv6.c
```diff
@@ -55,6 +64,63 @@ const struct xIPv6_Address FreeRTOS_in6addr_any = { 0 };
...
+    }
+
+    return xReturn;
+}
+
+
+/*-----------------------------------------------------------*/
+
 /**
  * @brief Check whether this IPv6 address is a multicast address or not.
```
shubnil, line 125: Please change the description as well to something like "Check whether this IPv6 address is an allowed multicast address or not." And add the description for return value as well.

source/FreeRTOS_IPv6.c
```diff
@@ -55,6 +64,63 @@ const struct xIPv6_Address FreeRTOS_in6addr_any = { 0 };
...
+
+    configASSERT( pxIPv6Address != NULL );
+
+    ( void ) memcpy( &( xReturnGroupID.ucBytes[ 2 ] ), &( pxIPv6Address->ucBytes[ 2 ] ), 14 );
+
+    return xReturnGroupID;
+}
+
+/**
+ * @brief Check if the packet is a legal loopback packet.
```
shubnil, line 96: I think, we do not need legal , we can just say "Check if the packet is a loopback packet." Same in line 100.

shubnil (changes requested):

source/FreeRTOS_IPv6.c
ActoryOu: Agree. Let me rename it.

shubnil (approved):

moninom1 (approved):

PR #811, opened 2023-03-24
Check packet length for IPv6.

ActoryOu:
Check packet length for IPv6.

Description
-----------

We check packet size for IPv4, but miss to check it for IPv6.
Separate `xCheckSizeFields` into `xCheckIPv4SizeFields` and `xCheckIPv6SizeFields`. Then check the size in `prvAllowIPPacketIPv6`.

Test Steps
-----------

Run protocol test case IPv6/Datagram/009.

Checklist:
----------

Related Issue
-----------


shubnil (approved):
I was also thinking on the same lines as Hein if we can make this a common function, However, saw that do/while loop part of the code needs to know if the context is v4/v6 , hence a common function might be tricky.

source/FreeRTOS_IPv6.c
```diff
@@ -55,6 +55,86 @@ const struct xIPv6_Address FreeRTOS_in6addr_any = { 0 };
...
+            }
+
+            /* Check if the complete IPv6-header plus protocol data have been transferred: */
+            usPayloadLength = FreeRTOS_ntohs( pxIPv6Packet->xIPHeader.usPayloadLength );
+
+            if( uxBufferLength != ( size_t ) ( ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + ( size_t ) usPayloadLength ) )
+            {
+                DEBUG_SET_TRACE_VARIABLE( xLocation, 4 );
+                break;
+            }
```
shubnil, line 120: Do we not need the UDP and TCP part of the code for IPv6.

ActoryOu:
> Thank you for this change, @ActoryOu. I haven't checked all details but will do so. I only wonder, how much code is common in the two functions, and couldn't they be merged to save flash space?

Because the basic structure of IPv4/IPv6 header format is different, I'm afraid that there is no common code.
That's why I just separate them into two static functions.

ActoryOu:
> Thank you for this change, @ActoryOu. I haven't checked all details but will do so. I only wonder, how much code is common in the two functions, and couldn't they be merged to save flash space?

Because the header format is too different between IPv4 and IPv6, I don't think it can be commonized.
Or we'll get lots of comparation like:
```
if ( isIPv4 )
    return sizeof( IPv4Header )
else
    return sizeof( IPv6Header )
```

And I don't think this could save any code size.

moninom1 (approved):

PR #827, opened 2023-04-10
[IPv6] Reduce stack memory usage.

ActoryOu:
[IPv6] Reduce stack memory usage.

Description
-----------

- Output to address pointer in IPv6MC_GET_GROUP_ID instead of structure.
- Add FreeRTOS_in6addr_allnodes in static memory instread of stack.

Test Steps
-----------

Run protocol test cases IPv6/Datagram.

Checklist:
----------

Related Issue
-----------


htibosch (approved):
@ActoryOu, thank you for this PR, I think it is all good!
I only wonder is `FreeRTOS_in6addr_allnodes` will be useful for other modules, some day? But we can decide that later.
Hein

ActoryOu:
> @ActoryOu, thank you for this PR, I think it is all good! I only wonder is `FreeRTOS_in6addr_allnodes` will be useful for other modules, some day? But we can decide that later. Hein

Agree. Currently I have no idea when it'll be used in other modules. So let's keep it static for now.

moninom1 (approved):

shubnil (approved):

PR #843, opened 2023-04-24
Supress MISRA single use warning for FreeRTOS_in6addr_loopback

tony-josi-aws:
Description
-----------
This PR suppresses the MISRA Rule 8.9 [An object should be defined at block scope if its identifier only appears in a single function] during the Coverity scan. 

Test Steps
-----------
Coverity run

Checklist:
----------

Related Issue
-----------


moninom1 (approved):

archigup:
Which variable is triggering the warning?

Is this a false positive? Why do we need to suppress it instead of fixing  it?

tony-josi-aws:
> Which variable is triggering the warning?
> 
> Is this a false positive? Why do we need to suppress it instead of fixing it?

Hi @archigup,

The variable is `FreeRTOS_in6addr_loopback`, it can be used analogous to the `in6addr_loopback` of `<netinet/in.h>` (linux or similar) by applications\tools as well other than just by the stack, hence its defined with external linkage.

shubnil (approved):

PR #845, opened 2023-04-25
[IPv6] Add Unit Test for FreeRTOS_Routing.

ActoryOu:
[IPv6] Add Unit Test for FreeRTOS_Routing.

Description
-----------

As the FreeRTOS_Routing was added for merging IPv4 and IPv6, this change adds the Unit test case for this new file along with 100% coverage.

Test Steps
-----------

Unit Test
```
```
Coverage Test
```
make -C ${BUILD_DIR} coverage
```

Checklist:
----------

Related Issue
-----------


source/FreeRTOS_Routing.c
```diff
@@ -615,84 +627,74 @@ struct xIPv6_Couple
...
+                }
 
-            ( void ) memcpy( pxEndPoint->xMACAddress.ucBytes, ucMACAddress, ipMAC_ADDRESS_LENGTH_BYTES );
-            ( void ) FreeRTOS_AddEndPoint( pxNetworkInterface, pxEndPoint );
+                /* Copy the current values to the default values. */
+                ( void ) memcpy( &( pxEndPoint->ipv6_defaults ), &( pxEndPoint->ipv6_settings ), sizeof( pxEndPoint->ipv6_defaults ) );
+
+                ( void ) memcpy( pxEndPoint->ipv6_defaults.xIPAddress.ucBytes, pxIPAddress->ucBytes, ipSIZE_OF_IPv6_ADDRESS );
+
+                ( void ) memcpy( pxEndPoint->xMACAddress.ucBytes, ucMACAddress, ipMAC_ADDRESS_LENGTH_BYTES );
```
tony-josi-aws, line 670: Why are we not setting the static IP address here, shouldn't we set the endpoint IP address as the static IP address provided by the user if neither DHCPv6 nor RA is enabled? If we expect DHCP or RA to giveup and then assign the default IP we might have to wait till the timeout for the endpoint to be up. Also DHCP or RA can be disabled during build.

source/FreeRTOS_Routing.c
```diff
@@ -97,35 +97,38 @@ struct xIPv6_Couple
...
+            ( void ) memcpy( &( pxEndPoint->ipv4_defaults ), &( pxEndPoint->ipv4_settings ), sizeof( pxEndPoint->ipv4_defaults ) );
 
-        /* The field 'ipv4_settings.ulIPAddress' will be set later on. */
+            /* The default IP-address will be used in case DHCP is not used, or also if DHCP has failed, or
+             * when the user chooses to use the default IP-address. */
+            pxEndPoint->ipv4_defaults.ulIPAddress = ulIPAddress;
 
-        ( void ) memcpy( pxEndPoint->xMACAddress.ucBytes, ucMACAddress, sizeof( pxEndPoint->xMACAddress ) );
-        ( void ) FreeRTOS_AddEndPoint( pxNetworkInterface, pxEndPoint );
+            /* The field 'ipv4_settings.ulIPAddress' will be set later on. */
```
tony-josi-aws, line 127: [Similar case](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/845#discussion_r1189617806) for IPv4 endpoints as well

source/FreeRTOS_Routing.c
```diff
@@ -615,84 +627,74 @@ struct xIPv6_Couple
...
-
-            pxEndPoint->bits.bIPv6 = pdTRUE_UNSIGNED;
-
-            pxEndPoint->ipv6_settings.uxPrefixLength = uxPrefixLength;
-
-            if( pxGatewayAddress != NULL )
+            if( ( pxNetworkInterface == NULL ) ||
+                ( pxEndPoint == NULL ) ||
+                ( pxIPAddress == NULL ) ||
+                ( ucMACAddress == NULL ) )
```
tony-josi-aws, line 633: In my opinion asserts are more easier to set up and helpful while debugging than `printf`s.

ActoryOu: Because this is a public API, I believe less assertion the better.

source/FreeRTOS_Routing.c
ActoryOu: That's a good point. Please allow me to list down my answer below:
1. We have a [reset on endpoint](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/845/files/c80762bb530940fda9c9fc58ed1c0647059ae601#diff-cabc7b626b59843c03a1b0a4018200289c01a46d67b68e3445249b18db94a853R644) in this function, so the DHCP/RA bits are always zero. It's useless to check DHCP/RA before setting IP.
2. We use a general way to set IP addresses for endpoints, it's in [prvProcessNetworkDownEvent](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/c80762bb530940fda9c9fc58ed1c0647059ae601/source/FreeRTOS_IP_Utils.c#L896-L906) for now. It may make flow more complicated if we can't tell if an endpoint is for DHCP/RA or not.

source/FreeRTOS_Routing.c
ActoryOu: Same as comment [below](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/845/files/c80762bb530940fda9c9fc58ed1c0647059ae601#r1193305558).

htibosch:
Hello @ActoryOu, thank you for this UT.
As you are proposing quite a few changes to /source, I need some time to review and test it. Can you either put it on hold or put it in draft mode for a while?
Thank you, Hein

shubnil (approved):

source/FreeRTOS_Routing.c
htibosch: Yes you're right, @ActoryOu, `memset()` is called to clear the entire endpoint.

I added this code for a reason:
Endpoint 1 (global IPv6) needs DHCPv6
Endpoint 2 (static local-link) is used to communicate with the DHCP server

I would like to set 'ipv6_settings.xIPAddress' for all static endpoints, before `prvProcessNetworkDownEvent()` is called for the first time. Otherwise DHCP can not find a local-link IPv6 address.

source/FreeRTOS_Routing.c
htibosch: Here it is not so important because an IPv4 endpoint doesn't normally need another IPv4 endpoint.

source/FreeRTOS_Routing.c
ActoryOu: Hi @htibosch,
Thanks for comment.
But one query, how do we know if the endpoint is static in this function?
Or is it ok to always set ipv6_settings/ipv4_settings IP address for endpoints?

ActoryOu:
Hi @htibosch,
There is one more point need your review.
I changed [priority order](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/845/files#diff-cabc7b626b59843c03a1b0a4018200289c01a46d67b68e3445249b18db94a853R711-R715) when searching endpoint for coming packets. Please help to take a look on it and make sure my change makes sense.

Thanks!!

htibosch (approved):
I have only looked at the changes in `FreeRTOS_Routing.c`. The looks good to me and I approve it.

This is where you change the status of your PR back to normal:

image

I will test the changes today.

source/FreeRTOS_Routing.c
```diff
@@ -371,6 +373,10 @@ struct xIPv6_Couple
                 pxResult = pxResult->pxNext;
             }
         }
+        else
+        {
+            pxResult = FreeRTOS_FirstEndPoint( pxInterface );
+        }
```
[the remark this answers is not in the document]
ActoryOu: I make `FreeRTOS_NextEndPoint` to provide first endpoint when input is NULL pointer.
That might be easier for users to use. For example:
```
pxEndpoint = NULL;

while( FreeRTOS_NextEndPoint( pxEndpoint ) != NULL )
{
    ...
}
```

source/FreeRTOS_Routing.c
```diff
@@ -407,7 +413,7 @@ struct xIPv6_Couple
         {
             #if ( ipconfigUSE_IPv4 != 0 )
                 #if ( ipconfigUSE_IPv6 != 0 )
-                    if( ENDPOINT_IS_IPv4( pxEndPoint ) )
+                    if( pxEndPoint->bits.bIPv6 == 0U )
```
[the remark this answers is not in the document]
ActoryOu: The pxEndPoint is always valid (not NULL). That means the pointer check is useless here.
I can either change the macro to skip pointer check or just check bIPv6 here in this change.

kar-rahul-aws (approved):

PR #870, opened 2023-05-22
[IPv6] Add Unit Test for FreeRTOS_Sockets & FreeRTOS_IPv4_Sockets.

ActoryOu:
[IPv6] Add Unit Test for FreeRTOS_Sockets & FreeRTOS_IPv4_Sockets.

Description
-----------

Separate FreeRTOS_Sockets.c unit test cases into FreeRTOS_Sockets/FreeRTOS_IPv4_Sockets and add test cases to reach 100% coverage.

Test Steps
-----------

Unit Test
```
```
Coverage Test
```
make -C ${BUILD_DIR} coverage
```

Checklist:
----------

Related Issue
-----------


source/include/FreeRTOS_IP_Common.h
```diff
@@ -47,6 +47,14 @@ typedef union IP_Address
     IPv6_Address_t xIP_IPv6; /**< IPv6 address */
 } IP_Address_t;
 
+/** @brief A struct that can hold either an IPv4 or an IPv6 address. */
+typedef struct xxIPv46_Address
+{
+    /* A struct that can hold either an IPv4 or an IPv6 address. */
+    IP_Address_t xIPAddress; /**< The IPv6-address. */
```
tony-josi-aws, line 54: `xIPAddress` can be used to hold either v4 or v6 address here, comment only says about v6.

source/FreeRTOS_Sockets.c
```diff
@@ -100,6 +100,15 @@
 #define sock80_PERCENT     80U         /**< 80% of the defined limit. */
 #define sock100_PERCENT    100U        /**< 100% of the defined limit. */
 
+
+#if ( ( ipconfigHAS_DEBUG_PRINTF != 0 ) || ( ipconfigHAS_PRINTF != 0 ) )
+
+/**
+ * @brief A buffer for prvSocketProps to return socket property in string.
+ */
+    static char pucSocketProps[ 92 ];
```
tony-josi-aws, line 109: Why are we moving this variable out of the `prvSocketProps()` if its only used inside that function?

source/include/FreeRTOS_IP_Common.h
moninom1: We already have a structure which holds both v4 and v6 address, why is this needed?

moninom1: Not sure why we are introducing this structure.

source/include/FreeRTOS_IPv6_Sockets.h
```diff
@@ -102,8 +102,7 @@
  *         is connected to a remote IP-address. It will be called from a loop
  *         iterating through all sockets. */
     FreeRTOS_Socket_t * pxTCPSocketLookup_IPv6( FreeRTOS_Socket_t * pxSocket,
-                                                const IPv6_Address_t * pxAddress_IPv6,
-                                                uint32_t ulRemoteIP );
+                                                const IPv46_Address_t * pxAddress );
```
moninom1, line 105: We already have correct IP types passed in both functions, If we are not making the function as one function then what is the point of updating the arguments. The end goal of updating the arguments and making it common should be to get rid of one of the function.

source/FreeRTOS_Sockets.c
ActoryOu: This is used to validate in UT.
If the variable is declared inside the function, we can't extern it in the UT to verify the result.

source/include/FreeRTOS_IPv6_Sockets.h
ActoryOu: We need a method to differentiate if the address is belong to IPv4 or IPv6. An dthe new added structure `IPv46_Address_t` has a bit for it.

source/FreeRTOS_Sockets.c
ActoryOu: Because we removed keyword `static` as preprocess for unit test. This variable became local variable that we can't access outside of function. Move it to global section to keep the buffer accessible for unit test.

source/include/FreeRTOS_IP_Common.h
```diff
@@ -47,6 +47,14 @@ typedef union IP_Address
     IPv6_Address_t xIP_IPv6; /**< IPv6 address */
 } IP_Address_t;
 
+/** @brief A struct that can hold either an IPv4 or an IPv6 address. */
+typedef struct xxIPv46_Address
+{
+    /* A struct that can hold either an IPv4 or an IPv6 address. */
+    IP_Address_t xIPAddress; /**< IP address contains either IPv4 or IPv6. */
+    BaseType_t xIs_IPv6;     /**< pdTRUE if IPv6 address. */
+} IPv46_Address_t;
```
moninom1, line 56: Hein's structure is used in DNS files extensively and we are making it a union now by using IP_Address_t in the same strcture, Please check it is not breaking anything in DNS files.

source/FreeRTOS_Sockets.c
moninom1: This might give coverity error, please check

source/include/FreeRTOS_IP_Common.h
ActoryOu: Already updated in this PR. Please check the change in DNS_Cache.c & DNS_Parser.c.
And I can pass unit test after changing.

source/FreeRTOS_Sockets.c
tony-josi-aws: I believe the pointer returned by `prvSocketProps()` is the address of `pucSocketProps` and we could use that to access the buffer during the unit tests, this will involve maybe calling the `prvSocketProps()` twice in a test or we could store the address in a global variable during the initial test and then use it in further tests.

ActoryOu: Debug print is disabled in coverity scan, so this change doesn't give any coverity error.
But I did find some coverity issues. Just updated for them.
Thanks for notifying.

moninom1 (approved):

tony-josi-aws (approved):

shubnil (approved):

go2sh:
Hey,
this breaks compilation with a shadow warning if not disable due to the same name used globally and in a function pucSocketProps if printf is enabled.
BR

PR #874, opened 2023-05-24
Add Unit Test Coverage for FreeRTOS_TCP_Transmission and FreeRTOS_TCP_Transmission_IPv4

moninom1:
Description
-----------
Add Unit Test Coverage for FreeRTOS_TCP_Transmission and FreeRTOS_TCP_Transmission_IPv4

Test Steps
-----------
Unit Test
```
```

Coverage Test
```
make -C ${BUILD_DIR} coverage
```

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_TCP_Transmission_IPV4.c
+++ b/source/FreeRTOS_TCP_Transmission_IPV4.c
@@ -102,13 +102,6 @@ void prvTCPReturnPacket_IPV4( FreeRTOS_Socket_t * pxSocket,
 
     do
     {
-        /* Use do/while to be able to break out of the flow */
-        if( ( pxNetworkBuffer == NULL ) && ( pxSocket == NULL ) )
-        {
-            /* Either 'pxNetworkBuffer' or 'pxSocket' should be defined. */
-            break;
-        }
-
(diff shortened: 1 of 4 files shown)


source/FreeRTOS_TCP_Transmission_IPV4.c
```diff
@@ -102,13 +102,6 @@ void prvTCPReturnPacket_IPV4( FreeRTOS_Socket_t * pxSocket,
 
     do
     {
-        /* Use do/while to be able to break out of the flow */
-        if( ( pxNetworkBuffer == NULL ) && ( pxSocket == NULL ) )
```
ActoryOu, line 106: I understand that it's not necessary to check network buffer. But why don't we check socket handler?

ActoryOu (changes requested):

source/FreeRTOS_TCP_Transmission_IPV4.c
moninom1: This function is called only when atleast one of them is not NULL. And both socket null or buffer null cases are handled in this function. Hence it is removed.

ActoryOu (approved):

shubnil (approved):

Issue #894, opened 2023-06-01
[Feature Request] Add priority event queues


go2sh:
**Is your feature request related to a problem? Please describe.**
For certain application it is desirable to have multiple priorities during sending of packets. For example an audio stream might have strict jitter requirements and thus should always be send on time. In conjunction with a network hardware supporting mutiple queues, this provides a powerful tool for time sensitive networking.

**Describe the solution you'd like**
Have multiple worker queues for tx and rx events to allow packets to overtake each other. All other events would be processed in the highst priority queue.

**Describe alternatives you've considered**
None

This feature could be implemented as an opt in. Only two functions access the network queue currently. These could be expanded to support priorities. The priority could set through setsockopt (SO_PRIORITY and/or IP_TOS) like in linux/posix. The user could create a map between the priority/TOS and queues and also between queues/priority and hw queues.

I have a working prototype.

shubnil:
Hi Christoph,

Thanks for adding this feature request.
I agree that this will be a useful addition for network hardware supporting mutiple FIFOs.

You also mentioned that you have a working prototype. It will be good if you can share the changes as a PR and we can work together for review and fine-tuning and merging the changes.

shubnil:
Hi @go2sh,
Gentle Reminder.

go2sh:
Here you go @shubnil . Happy to see your comments coming in :)

shubnil:
Thanks a lot, Christoph. This is really helpful.
We have started reviewing the change and will discuss internally more on the design and synch up with you as well more on the design part.
Please note that the actual merge might happen only in August. We have an important GA release coming up and we will be freezing the branch in a week. However, we will make sure that we can merge the change as soon as the release tagging is done.
Thanks for being patient.

PR #925, opened 2023-06-15
[RFC] Add priority queues

go2sh:
Add priority queues.

Description
-----------
This PR adds priorities to socket and packets and multiple event queues to handle packets. This change enables multiple things with in the time sensitive networking space:
 - The stack can handle packets based on their priority to allow overtakes or idle background traffic using the remaining bandwidth. An example would be audio/video data or PTP message as high priority or a data dump in the background.
 - The packet priority can be used by network drivers to assign the packet to a different hardware queue.
 - The HW can add VLAN tags with priority fields. (And maybe later FreeRTOS-Plus-TCP :) )

Test Steps
-----------
Enable the feature via ipconfigPRIORITIES && ipconfigEVENT_QUEUES and add two udp sockets. The first socket gets a higher priority  (e.g. 5). The second socket sends a lot of packets and the first socket afterwards one. This packet should overtake some of the packets of the second socket.

Checklist:
----------

- [X] I have tested my changes. No regression in existing tests.

Related Issue
-----------
#894 

TODOs
-----------
- [ ] Make TCP sockets adhere the priority. This would basicly mean, that a TCP packet before sending must be placed back into the Queues and cannot be send directly.
- [ ] Add unit tests for the new features.

### DIFF
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -178,6 +178,11 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
 /** @brief The queue used to pass events into the IP-task for processing. */
 QueueHandle_t xNetworkEventQueue = NULL;
 
+#if ipconfigEVENT_QUEUES > 1
+    QueueHandle_t xNetworkEventQueues[ ipconfigEVENT_QUEUES ] = { 0 };
+    uint8_t xQueueMapping[ ipconfigPACKET_PRIORITIES ] = ipconfigPACKET_PRIORITY_MAPPING;
+#endif
+
 /** @brief The IP packet ID. */
 uint16_t usPacketIdentifier = 0U;
 
@@ -271,10 +276,31 @@ static void prvProcessIPEventsAndTimers( void )
     /* Wait until there is something to do. If the following call exits
      * due to a time out rather than a message being received, set a
      * 'NoEvent' value. */
-    if( xQueueReceive( xNetworkEventQueue, ( void * ) &xReceivedEvent, xNextIPSleep ) == pdFALSE )
-    {
-        xReceivedEvent.eEventType = eNoEvent;
-    }
+    #if ipconfigEVENT_QUEUES > 1
+        if( ulTaskNotifyTake( pdFALSE, xNextIPSleep ) == pdFALSE )
+        {
+            xReceivedEvent.eEventType = eNoEvent;
+        }
+        else
+        {
+            BaseType_t xIndex;
+            BaseType_t xReturn;
+
+            for( xIndex = ipconfigEVENT_QUEUES - 1; xIndex >= 0; xIndex-- )
+            {
+                if( uxQueueMessagesWaiting( xNetworkEventQueues[ xIndex ] ) > 0 )
+                {
+                    xQueueReceive( xNetworkEventQueues[ xIndex ], &xReceivedEvent, 0 );
+                    break;
+                }
+            }
+        }
+    #else /* if ipconfigEVENT_QUEUES > 1 */
+        if( xQueueReceive( xNetworkEventQueue, ( void * ) &xReceivedEvent, xNextIPSleep ) == pdFALSE )
+        {
+            xReceivedEvent.eEventType = eNoEvent;
+        }
+    #endif /* if ipconfigEVENT_QUEUES > 1 */
 
     #if ( ipconfigCHECK_IP_QUEUE_SPACE != 0 )
         {
@@ -942,6 +968,11 @@ void * FreeRTOS_GetUDPPayloadBuffer_Multi( size_t uxRequestedSizeBytes,
 BaseType_t FreeRTOS_IPInit_Multi( void )
 {
     BaseType_t xReturn = pdFALSE;
+    BaseType_t xAllocSuccessful = pdFALSE;
+
+    #if ipconfigEVENT_QUEUES > 1
+        BaseType_t xIndex;
+    #endif
 
     /* There must be at least one interface and one end-point. */
     configASSERT( FreeRTOS_FirstNetworkInterface() != NULL );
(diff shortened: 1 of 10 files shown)


source/FreeRTOS_IP.c
```diff
@@ -271,10 +276,31 @@ static void prvProcessIPEventsAndTimers( void )
     /* Wait until there is something to do. If the following call exits
      * due to a time out rather than a message being received, set a
      * 'NoEvent' value. */
-    if( xQueueReceive( xNetworkEventQueue, ( void * ) &xReceivedEvent, xNextIPSleep ) == pdFALSE )
-    {
-        xReceivedEvent.eEventType = eNoEvent;
-    }
+    #if ipconfigEVENT_QUEUES > 1
+        if( ulTaskNotifyTake( pdFALSE, xNextIPSleep ) == pdFALSE )
```
AniruddhaKanhere, line 280: Makes sense to decrement it by making the first parameter 'pdFALSE'. Behaves like a counting semaphore.

source/FreeRTOS_IP.c
```diff
@@ -1411,7 +1489,24 @@ BaseType_t xSendEventStructToIPTask( const IPStackEvent_t * pxEvent,
...
+                {
+                    NetworkBufferDescriptor_t * pxBuffer = ( NetworkBufferDescriptor_t * ) pxEvent->pvData;
+                    xQueue = xQueueMapping[ pxBuffer->ucPriority ];
+                }
+
+                xReturn = xQueueSendToBack( xNetworkEventQueues[ xQueue ], pxEvent, uxUseTimeout );
+
+                if( xReturn != pdFAIL )
+                {
+                    xTaskNotifyGive( xIPTaskHandle );
```
AniruddhaKanhere, line 1505: Here too, makes sense. Behaves like a counting semaphore.

source/FreeRTOS_IP.c
```diff
@@ -1392,7 +1466,11 @@ BaseType_t xSendEventStructToIPTask( const IPStackEvent_t * pxEvent,
                      * IP task is already awake processing other message. */
                     vIPSetTCPTimerExpiredState( pdTRUE );
 
-                    if( uxQueueMessagesWaiting( xNetworkEventQueue ) != 0U )
+                    #if ipconfigEVENT_QUEUES > 1
+                        if( ulTaskGenericNotifyValueClear( xIPTaskHandle, 0, 0 ) != 0U )
```
AniruddhaKanhere, line 1470: I believe `ulTaskNotifyValueClear` is the better version to use here to be consistent with all your other task notification API calls.

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -1161,4 +1161,32 @@
...
+    #if ipconfigEVENT_QUEUES > 1
+        #define ipconfigPACKET_PRIORITIES    8
+    #else
+        #define ipconfigPACKET_PRIORITIES    1
+    #endif
+#endif
+
+#ifndef ipconfigPACKET_PRIORITY_MAPPING
+    #if ipconfigPACKET_PRIORITIES == 8 && ipconfigEVENT_QUEUES == 3
+        #define ipconfigPACKET_PRIORITY_MAPPING    { 0, 1, 1, 1, 2, 2, 2, 2, }
```
AniruddhaKanhere, line 1178: Just trying to understand this scheme here. Please do correct me if I am wrong.

All messages with priority 0 will be added to Queue 0.
And all messages with priority 1-3 will be added to Queue 1.
And all messages with priority 4-7 will be added to Queue 2.

Is that right?
If so, then what is the point of having 8 different priorites if you just have 3 queues?

AniruddhaKanhere (commented):
Hello @go2sh,

This is a very interesting prospect/suggestion.
While we discuss this PR with our team, I have an initial review with a couple of questions/concerns.

Thanks for taking the time to contribute to FreeRTOS+TCP.

-Aniruddha

source/FreeRTOS_IP.c
```diff
@@ -1411,7 +1489,24 @@ BaseType_t xSendEventStructToIPTask( const IPStackEvent_t * pxEvent,
...
+            #if ipconfigEVENT_QUEUES > 1
+                BaseType_t xQueue = ipconfigEVENT_QUEUES - 1;
+
+                if( ( pxEvent->eEventType == eNetworkRxEvent ) || ( pxEvent->eEventType == eNetworkTxEvent ) || ( pxEvent->eEventType == eStackTxEvent ) )
+                {
+                    NetworkBufferDescriptor_t * pxBuffer = ( NetworkBufferDescriptor_t * ) pxEvent->pvData;
+                    xQueue = xQueueMapping[ pxBuffer->ucPriority ];
+                }
+
+                xReturn = xQueueSendToBack( xNetworkEventQueues[ xQueue ], pxEvent, uxUseTimeout );
```
AniruddhaKanhere, line 1501: Does this mean that all non-data-related events (like a `eNetworkDown` event) are highest priority and will be processed first?
I am not sure whether that will cause any unforeseen race conditions... It should not... But I am still thinking about it :)

go2sh: Exactly. Initially I thought to flush the other queues but that is not possible. , since there might be other interfaces.

source/include/FreeRTOSIPConfigDefaults.h
go2sh: The idea is based on the current linux implementation, where tc-prio has 3 fifos. The 8 comes from the size of the vlan pcp which has 3 bits so 8 priorities. I do hw mappings inside of the driver and add vlan tags with the priority.

source/FreeRTOS_IP.c
go2sh: I don't understand your comment. From https://www.freertos.org/ulTaskNotifyTake.html . The first parameter is the clear on exit and the second parameter the time to wait. Isn't the call exactly like it should be? Wait for the notification value and decrement it on exit. Only look at the event queue if the notification value is not 0.

source/FreeRTOS_IP.c
go2sh: Same here. xTaskNotifyGive just increments the notification value by 1.

source/FreeRTOS_Sockets.c
```diff
@@ -2928,6 +2934,24 @@ BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                         xReturn = prvSetOptionStopRX( pxSocket, pvOptionValue );
                         break;
                 #endif /* ipconfigUSE_TCP == 1 */
+                #if ( ipconfigPACKET_PRIORITIES > 1 )
+                    case FREERTOS_SO_PRIORITY:
+                       {
+                           uint8_t * pucValue = ( uint8_t * ) pvOptionValue;
```
htibosch, line 2940: In Linux, the priority is passed as an `int`, in FreeRTOS+tCP we have the habit of using `BaseType_t`:
~~~diff
-    uint8_t * pucValue = ( uint8_t * ) pvOptionValue;
+    BaseType_t * pxValue = ( BaseType_t * ) pvOptionValue;
~~~

source/include/FreeRTOSIPConfigDefaults.h
htibosch: Could you please add a bit of comment about the solution here? It is confusing to see 3 queues, and 8 different priorities.

htibosch (commented):
Hi @go2sh, this looks like a useful addition: priorities. It is a good idea to use TaskNotify in combination with `uxQueueMessagesWaiting()`.

Do you have an application that needs priorities? I am curious to hear how it works for you, with and without multiple queues.

As I write here below, `eNetworkRxEvent` packets have just arrived in the interface, and they do not have a specific priority. So you don't have to test it, just let them travel with the highest priority in order to avoid congestion.

Thank you very much for this contribution.
Hein

go2sh:
A general comment about priorities and queues:
* The number of priorities is chosen based on the VLAN Tag PCP field with 3 bit. (See https://en.wikipedia.org/wiki/IEEE_P802.1p) Other possibilities are TOS field from IP header (4 bit / 16 numbers) or simply the number queues. Its really application dependent. I think the most useful is the VLAN PCP variant with 8. In my case: The NetworkInterface (a Infineon TC3 and TC4 port) inserts and removes a vlan tag via the descriptors and the PCP value is used to assign dedicated tx and rx hardware queues. Also the PCP field for RX packets is written as priority into the NetworkBufferDescriptor_t field of a received packet overwriting the default value and thus the priority handling of rx events.
* The number of queues is chosen based on the default number 3 from the tc-prio scheduler of the linux kernel. (https://www.man7.org/linux/man-pages/man8/tc-prio.8.html) The idea behind 3 is: One queue for best effort. No garuantees at all. Gets what is left. One queue for normal app traffic like HTTP/FTP/MQTT etc. One queue for high priority traffic like low latency traffic or real time traffic. (e.g. Audio/Video, Real-Time Sense and Control or Time Sync). The number of queues might be equeal to the number of priorities, but for most applications the number of packets with a high priority are rather small compared to the rest of the traffic. So more queues might waste resources, since only a few messages are stored in there.

I don't mind the numbers at all. A just wanted to choose a reasonable default, since anybody can overwrite it relatively easy.

A comment on the application: I work at Infineon and I port FreeRTOS+TCP on our TC3 and TC4 microcontroller. These are rather powerful devices with a lot of cores (up to 6), a lot of SRAM (a few MB) and up to 5G Ethernet MACs with 8 DMA and 8 Hardware Queues each for RX and TX). We run relativly complex applications with multiple tasks working on network traffic with diffrent application domains. From SOTA, CAN Tunneling, Service Handling (like SOMEIP or MQTT) and real time data like Audio and Video (or Radar) streams. These patch is a first step off removing some bottlenecks with the stack and to fulfill the real-time requirements.

As an concrete example: I have an app which runs some low priority Debug Instrumentation, an MQTT Client and some Audio Processing. Since the system has a lot of ram, the number of NetworkBuffers is high (eg. 128). This lead to the instrumentaion filling up the single queue and the audio task had no chance of meeting the real time latency requirement. With this multiple queue approach: The instrumentation runs in the lowest priority and gets rest of the bandwith. Its no problem if a packets waits a bit. The overall bandwidth of the system is sufficient. The MQTT traffic runs in the middle priority and the audio and timesync traffic in the highest. (Only a few 100 kBit/s but very latency sensitive). With this multiple queues approach, I don't need take any measures within the software to meet my real-time goals.

I hope this makes this a bit more clear. :)

source/include/FreeRTOSIPConfigDefaults.h
go2sh: Add added a big comment below. I will add comment in the code as well.

source/FreeRTOS_IP.c
go2sh: I was confused, sorry. :D

go2sh:
One more comment regarding: RX Events. The problem why I also added the priorities for RX events is the following: Think about a lot of NetworkBuffers and your IP Tasks start blocking until a descriptor entry is free inside the descriptor table. During that time no RX packet handling will happen as the IP task is blocking. The high priority queue makes it possible, that an high priority packet is processed a bit earlier than otherwise. The ideal solution would be to split RX and TX Packet processing. Then no extra queues for RX would be needed, as the traffic comes in prioritized already.

shubnil:
Thanks Christoph for creating the PR.
We will discuss internally more on the design and , if needed, I will synch up with you as well more on the design part and take the change forward.

source/include/FreeRTOSIPConfigDefaults.h
tony-josi-aws: Wouldn't it be better to have the number of priorities and the number queues to be equal in the default configuration (if we are keeping a default mapping), to avoid users getting confused about packets with different priorites getting into the same queue (depending on the mapping) and processed together regardless of their actual priorities.

go2sh:
> @go2sh thoughts on [mq-prio](https://man7.org/linux/man-pages/man8/tc-mqprio.8.html)?
That is my final step of my TSN chain.

I would see this functionality as part of the network interface. In my case, the driver checks the NetworkBuffer ucPriority and has a custom mapping array for mapping the packets to the different DMA queues. But each devices does it differently and I know that alot of other devices support multiple hw queues, which are already part of FreeRTOS+TCP.

htibosch:
Hi @go2sh, I am sorry to see that your PR is still here, already for two years.
What do you want to do with it? Withdraw it? Or give it a second chance?
My personal opinion is: it is nice-to-have, but mostly for the audio/video people.
But it will change the code all-over, and therefore it will need a lot of testing and checking.

Issue #930, opened 2023-06-18
Unused function and unused variable warnings when ipconfigUSE_TCP_WIN is disabled


timaydin:
Testing with version 3.1.0

When ipconfigUSE_TCP_WIN is disabled, the following warnings occur:

../src/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_Transmission.c:1317:20: warning: unused variable 'xSizeWithoutData' [-Wunused-variable]

../src/FreeRTOS-Plus-TCP/source/FreeRTOS_TCP_WIN.c:386:17: warning: 'vListInsertGeneric' defined but not used [-Wunused-function]

kar-rahul-aws:
Hi @timaydin 
Thank you for reporting the issue. We are looking into it and will update after an internal discussion.

kar-rahul-aws:
Hi @timaydin 
We have added this [patch](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/932) to remove the warnings , on disabling the macro **ipconfigUSE_TCP_WIN**
We request you to check running your project with the latest change, and see if that fixes the issue reported.
Thanks.

moninom1:
Closing the issue with the PR being merged : https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/932
Feel free to let us know if you face any issue.

PR #932, opened 2023-06-20
Fix warnings on disable USETCP_WIN macro

kar-rahul-aws:
Description
-----------
This change fixes warnings which appear on disabling the macro ipconfigUSE_TCP_WIN
This PR is a fix for the [customer reported issue](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/930)

Test Steps
-----------
Run the [Echo Posix Demo](https://github.com/FreeRTOS/FreeRTOS/tree/main/FreeRTOS-Plus/Demo/FreeRTOS_Plus_TCP_Echo_Posix) with warnings enabled , the demo runs successfully without these warnings

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_ND.c
+++ b/source/FreeRTOS_ND.c
@@ -479,6 +479,7 @@
         void FreeRTOS_PrintNDCache( void )
         {
             BaseType_t x, xCount = 0;
+            char pcBuffer[ 40 ];
 
             /* Loop through each entry in the ND cache. */
             for( x = 0; x < ipconfigARP_CACHE_ENTRIES; x++ )
@@ -486,7 +487,6 @@
                 if( xNDCache[ x ].ucValid != ( uint8_t ) 0U )
                 {
                     /* See if the MAC-address also matches, and we're all happy */
-                    char pcBuffer[ 40 ];
 
                     FreeRTOS_printf( ( "ND %2d: age %3u - %pip MAC %02x-%02x-%02x-%02x-%02x-%02x endPoint %s\n",
                                        ( int ) x,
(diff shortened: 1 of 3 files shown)


source/FreeRTOS_TCP_WIN.c
```diff
@@ -383,28 +385,30 @@
...
+            listLIST_ITEM_CONTAINER( pxNewListItem ) = ( struct xLIST * configLIST_VOLATILE ) pxList;
 
-        ( pxList->uxNumberOfItems )++;
-    }
+            ( pxList->uxNumberOfItems )++;
+        }
+    #endif /* if ( ipconfigUSE_TCP_WIN == 1 ) */
 /*-----------------------------------------------------------*/
 
     #if ( ipconfigUSE_TCP_WIN == 1 )
```
moninom1, line 414: There is already a flag, can we add above function to the same flag. Also is the complete file not expected to be within the ipconfigUSE_TCP_WIN flag.

source/FreeRTOS_ND.c
```diff
@@ -479,14 +479,14 @@
         void FreeRTOS_PrintNDCache( void )
         {
             BaseType_t x, xCount = 0;
+            char pcBuffer[ 40 ];
```
moninom1, line 482: What does this change, i do not see any flag difference between the 2 locations.

moninom1 (approved):

source/FreeRTOS_TCP_WIN.c
moninom1: This will require separate effort, may be we can add it to backlog. Approving for now.

shubnil (approved):

PR #935, opened 2023-06-20
resolve Wconversion related warnings

timaydin:
Description
-----------
Avoid warnings produced as a result of -Wconversion option to gcc

### DIFF
--- a/source/FreeRTOS_TCP_Reception.c
+++ b/source/FreeRTOS_TCP_Reception.c
@@ -331,14 +331,14 @@
                          */
                         if( pucPtr[ 0U ] == tcpTCP_OPT_SACK_A )
                         {
-                            ucLen -= 2U;
+                            ucLen -= 2;
                             lIndex += 2;
 
                             while( ucLen >= ( uint8_t ) 8U )
                             {
                                 prvReadSackOption( pucPtr, ( size_t ) lIndex, pxSocket );
                                 lIndex += 8;
-                                ucLen -= 8U;
+                                ucLen -= 8;
                             }
 
                             /* ucLen should be 0 by now. */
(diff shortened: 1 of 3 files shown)


moninom1:
Hi @timaydin 

There are some CI checks failure, can you please help in checking them.

Thank You

source/FreeRTOS_TCP_Reception.c
```diff
@@ -331,14 +331,14 @@
                          */
                         if( pucPtr[ 0U ] == tcpTCP_OPT_SACK_A )
                         {
-                            ucLen -= 2U;
+                            ucLen -= 2;
```
AniruddhaKanhere, line 334: I am unsure about this change. `ucLen` is unsigned. Then while assigning 2 to it, why not make 2 explicitly signed by using `U`?

moninom1: This will also give MISRA violation.

source/FreeRTOS_TCP_Reception.c
```diff
@@ -331,14 +331,14 @@
...
                             lIndex += 2;
 
                             while( ucLen >= ( uint8_t ) 8U )
                             {
                                 prvReadSackOption( pucPtr, ( size_t ) lIndex, pxSocket );
                                 lIndex += 8;
-                                ucLen -= 8U;
+                                ucLen -= 8;
```
moninom1, line 341: This is not solving the issue . Earlier warning was "**conversion from ‘unsigned int’ to ‘uint8_t**’" with your fix the warning has now changed to **conversion from ‘int’ to ‘uint8_t’**

moninom1:
Hi @timaydin  I was reproducing the issue and i can see that there are 40+ warning with enabling Wconverstion. Were you seeing the same number of issue?  If not can you let us know how you reproduced the issue?

PR #927, opened 2023-06-16
[IPv6] Allow buffer length to be greater than IP packet in IPv6.

ActoryOu:
[IPv6] Allow buffer length to be greater than IP packet in IPv6.

Description
-----------

The change is made by @htibosch.

Thanks @htibosch to notify that SAM driver returns more buffer length than size of IP packet.
SAM driver puts some data in the buffer and it increases the buffer length than size of packet. So we made this change to allow data length to be greater than IPv6 packets.

Test Steps
-----------

Below scenarios are tested (by @htibosch) on both IPv4/IPv6:
- DNS
- UDP
- TCP
- Ping
- IPERF3

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [x] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #941, opened 2023-06-28
[IPv6] Revert "Allow buffer length to be greater than IP packet in IPv6"

ActoryOu:
[IPv6] Revert PR#927 "Allow buffer length to be greater than IP packet in IPv6"

Description
-----------

Revert #927 because we're failing one critical test case in protocol tester.

Test Steps
-----------

Run test case IPv6/Datagram/009

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IPv6.c
+++ b/source/FreeRTOS_IPv6.c
@@ -132,9 +132,7 @@ const struct xIPv6_Address FreeRTOS_in6addr_loopback = { { 0, 0, 0, 0, 0, 0, 0, 
             /* Check if the complete IPv6-header plus protocol data have been transferred: */
             usPayloadLength = FreeRTOS_ntohs( pxIPv6Packet->xIPHeader.usPayloadLength );
 
-            /* Since network interface might put some data in the network buffer,
-             * we allow buffer length to be greater than necessary. */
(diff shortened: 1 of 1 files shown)


moninom1 (approved):

htibosch (approved):
The packet length in IPv6 seems to be more critical than in IPv4, I didn't know that.
Note that we'll have to change the reception in the SAM driver: subtract 2 bytes from the received length.

ActoryOu:
> The packet length in IPv6 seems to be more critical than in IPv4, I didn't know that. Note that we'll have to change the reception in the SAM driver: subtract 2 bytes from the received length.

It's good to see your comment here.
Yeah it's better to handle this kind of "extra length" in network interface itself.

shubnil:
The check for packet payload greater than the header length is specific to a particular SKU of SAM board. Hence, the decision to revert this change, so that the particular Network Interface can take the necessary active and pass the buffer to the FreeRTOS-plus-TCP stack in the standard form where the payload size matches with the header payload length.

shubnil (approved):

shubnil:
> The packet length in IPv6 seems to be more critical than in IPv4, I didn't know that. Note that we'll have to change the reception in the SAM driver: subtract 2 bytes from the received length.

Do we know why we have those 2 extra bytes and if we will ever need them in the stack for any future processing.

htibosch:
The patch #950 should correct the number of bytes received.

PR #949, opened 2023-06-29
Fixes mDNS over IPv6.

evpopov:
I just ran into what appears to be a simple oversight.. maybe someone forgot to copy/paste the mDNS code from v4 to v6
Until this code was added in, my xApplicationDNSQueryHook_Multi() was not getting called at all for IPv6 packets.

Please verify as this was a quick and dirty fix on my end and I may be missing something.

### DIFF
--- a/source/FreeRTOS_UDP_IPv6.c
+++ b/source/FreeRTOS_UDP_IPv6.c
@@ -608,6 +608,16 @@ BaseType_t xProcessReceivedUDPPacket_IPv6( NetworkBufferDescriptor_t * pxNetwork
                 else
             #endif /* ipconfigUSE_LLMNR */
 
+            #if ( ipconfigUSE_DNS == 1 ) && ( ipconfigUSE_MDNS == 1 )
+                /* A MDNS request, check for the destination port. */
+                if( ( usPort == FreeRTOS_ntohs( ipMDNS_PORT ) ) ||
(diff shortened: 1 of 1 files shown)


htibosch (approved):
It is very wel possible that the code was missing. It has happened in the period when we were all very busy remodeling the IPv6/multi library.
As I'm not in my office these weeks, I can not test it. But I trust that you have tested it well.
Thanks, Hein

kstribrnAmzn:
There looks to be some slight differences between this UDP_IPv6 and [UDP_IPv4 segment](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/dev/IPv6_integration/source/FreeRTOS_UDP_IPv4.c#L529-L548) but I'm no expert here. I'll forward this along to the IPv6 +TCP folks on our team to take a look.

kstribrnAmzn:
@shubnil and @ActoryOu can you take a look?

ActoryOu (approved):
It looks good to me.

moninom1 (approved):

moninom1:
Hi @evpopov 

Thank you for the change, we are presently in the process of tagging RC3, we will merge this change after that.
Thank you for the patience.

moninom1 (approved):

shubnil (approved):

PR #950, opened 2023-06-30
Atmel SAM driver: let gmac_dev_read() return the proper length

htibosch:
This PR is now closed and replaced with PR #1000 
---
Description
-----------
When receiving data, the driver for SAM will increase the number of bytes:
~~~
    /* Read +2 bytes because buffers are aligned at -2 bytes */
    bytesLeft = min( bytesLeft + 2, ( int32_t ) ul_frame_size );
~~~

This worked OK until PR #941, which is stricter on checking the received length:
~~~diff
-    if( uxBufferLength < ( size_t ) ( ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + ( size_t ) usPayloadLength ) )
+    if( uxBufferLength != ( size_t ) ( ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + ( size_t ) usPayloadLength ) )
~~~

Test Steps
-----------
Compile an application for SAM with the latest release ( after PR 941 ). Compare the behaviour before and after this PR.

Checklist:
----------

As I am not in my office, I am not able to test the change in real hardware.

### DIFF
--- a/source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
+++ b/source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
@@ -1011,7 +1011,7 @@ void gmac_handler( gmac_device_t * p_gmac_dev )
 /*/ @cond 0 */
 /**INDENT-OFF**/
 #ifdef __cplusplus
-}
+    }
 #endif
 /**INDENT-ON**/
 /*/ @endcond */
--- a/source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
+++ b/source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
@@ -621,6 +621,9 @@ uint32_t gmac_dev_read( gmac_device_t * p_gmac_dev,
         return GMAC_RX_NO_DATA;
     }
 
+    /* Return the number of bytes received. */
+    *p_rcv_size = bytesLeft;
+
     /* gmac_dev_poll has confirmed that there is a complete frame at
      * the current position 'ul_rx_idx'
      */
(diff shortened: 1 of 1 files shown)


kstribrnAmzn (approved):

source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
```diff
@@ -621,6 +621,9 @@ uint32_t gmac_dev_read( gmac_device_t * p_gmac_dev,
         return GMAC_RX_NO_DATA;
     }
 
+    /* Return the number of bytes received. */
+    *p_rcv_size = bytesLeft;
```
ActoryOu, line 625: Is it necessary to compare with frame size?
The `*p_rcv_size` might be different when `ul_frame_size` is smaller than `bytesLeft + 2`.
```
if( bytesLeft + 2 <= ul_frame_size )
{
    *p_rcv_size = bytesLeft;
}
else
{
    *p_rcv_size = ul_frame_size;
}
```

AniruddhaKanhere, line 625:
```suggestion
    *p_rcv_size = min( bytesLeft + 2, ( int32_t ) ul_frame_size );
```

AniruddhaKanhere: @ActoryOu and @htibosch this should be the fix. Am I correct?

AniruddhaKanhere (changes requested):

source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
ActoryOu: Hi @AniruddhaKanhere,
It can't be `bytesLeft + 2` because the stack compares packet length with buffer length.
It's only allowed when buffer length is exactly equal to packet length, which is bytesLeft.
Take #941 as reference.

PR #952, opened 2023-07-04
[IPv6] Fix checksum calculation of IPv6 with extension headers.

ActoryOu:
[IPv6] Fix checksum calculation of IPv6 with extension headers- RFC 2460 section 4.1

Description
-----------

With this change, FreeRTOS-Plus-TCP supports to calculate checksum for IPv6 packets with IPv6 extension headers.

Test Steps
-----------

Pass IPv6 TCP/UDP/ICMP checksum test cases with this change.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IPv6_Utils.c
+++ b/source/FreeRTOS_IPv6_Utils.c
@@ -76,6 +76,7 @@ BaseType_t prvChecksumIPv6Checks( uint8_t * pucEthernetBuffer,
                                   struct xPacketSummary * pxSet )
 {
     BaseType_t xReturn = 0;
+    size_t uxExtensionHeaderLength = 0;
 
     pxSet->xIsIPv6 = pdTRUE;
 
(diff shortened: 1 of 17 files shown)


source/FreeRTOS_IPv6_Utils.c
```diff
@@ -161,6 +172,117 @@ BaseType_t prvChecksumICMPv6Checks( size_t uxBufferLength,
...
+        if( xGetExtensionOrder( ucCurrentHeader, 0U ) > 0 )
+        {
+            while( ( uxIndex + 8U ) < uxBufferLength )
+            {
+                ucNextHeader = pucEthernetBuffer[ uxIndex ];
+
+                xCurrentOrder = xGetExtensionOrder( ucCurrentHeader, ucNextHeader );
+
+                /* To avoid compile warning if debug print is disabled. */
+                ( void ) xCurrentOrder;
```
tony-josi-aws, line 215: This line may not be needed as `xCurrentOrder` is getting used in line 277.

tony-josi-aws:
In the case of NA, NS and RS packets shouldn't we offload the checksum calculation to hardware based on `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM` flag?

source/FreeRTOS_IPv6_Utils.c
ActoryOu: No, line 277 is using `ucCurrentHeader`, not `xCurrentOrder`.

ActoryOu:
> In the case of NA, NS and RS packets shouldn't we offload the checksum calculation to hardware based on `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM` flag?

Good catch. Updated to check `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM` before calling `usGenerateProtocolChecksum` in NA/NS/RS outgoing packets.

tony-josi-aws (approved):

tony-josi-aws (approved):

source/FreeRTOS_ND.c
```diff
@@ -639,9 +639,18 @@
                 ( void ) memcpy( pxICMPHeader_IPv6->ucOptionBytes, pxEndPoint->xMACAddress.ucBytes, ipMAC_ADDRESS_LENGTH_BYTES );
 
                 /* Checksums. */
-                pxICMPHeader_IPv6->usChecksum = 0U;
-                /* calculate the ICMP checksum for the outgoing package. */
-                ( void ) usGenerateProtocolChecksum( pxDescriptor->pucEthernetBuffer, pxDescriptor->xDataLength, pdTRUE );
+                #if ( ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM == 0 )
+                    {
+                        /* calculate the ICMPv6 checksum for outgoing package */
```
moninom1, line 644: Is the pxICMPHeader_IPv6->usChecksum = 0U; not needed here? Earlier we were calling generate checksum after initializing to zero.

ActoryOu: It's not necessary because we use 0 as initial value when calculating checksum.

moninom1:
Can you please mentioned RFC in the commit message as well.

AniruddhaKanhere (approved):

PR #969, opened 2023-07-14
Uncrustify PR Check Update

Skptak:
Uncrustify PR Check Update

Description
-----------
Change to uncrustify to show files that need to be updated and create a cleaner log on runs

Test Steps
-----------
Ran the workflow after the changes to ensure it worked, [failed run one](https://github.com/Skptak/FreeRTOS-Plus-TCP/actions/runs/5549996475/jobs/10134669506#step:4:577), [failed run two](https://github.com/Skptak/FreeRTOS-Plus-TCP/actions/runs/5549885222/jobs/10134447280#step:4:577), and a [passing run](https://github.com/Skptak/FreeRTOS-Plus-TCP/actions/runs/5550053863/jobs/10134787257#step:4:577)

Checklist:
----------

Related Issue
-----------


.github/workflows/ci.yml
```diff
@@ -8,81 +8,23 @@ on:
   workflow_dispatch:
 
 jobs:
-  unittest:
```
AniruddhaKanhere, line 11: Why are we removing unit test?
Same goes for doxygen check.

Is this intentional?

Skptak: I had removed the jobs so that I didn't run everything on every commit to the PR. I've added them back in at this point.

.github/workflows/ci.yml
```diff
@@ -42,62 +42,20 @@ jobs:
...
-        run: apt-get update && apt-get install uncrustify
-      - name: Run Uncrustify
-        run: |
-          uncrustify --version
-          find . -iname "*.[hc]" -exec uncrustify --check -c tools/uncrustify.cfg {} +
-          if [ "$?" = "0" ]; then
-            exit 0
-          else
-            echo -e "\033[31;1;43mFormatting check (using Uncrustify) failed...\033[0m"
-            echo -e "\033[32;3mTo have the code uncrustified for you, please comment '/bot run uncrustify' (without the quotes) on the Pull Request.\033[0m"
```
AniruddhaKanhere, line 88: I would still like for this to be printed out in case of failure. This repo has a workflow which uncurstifies the code when you comment that.

.github/workflows/ci.yml
```diff
@@ -42,62 +42,20 @@ jobs:
...
-          path: ./tcp
-      - name: Install spell
-        run: |
-          sudo apt-get install spell
-          sudo apt-get install util-linux
-      - name: Check spelling
-        run: |
-          PATH=$PATH:main/tools/spell
-          # Make sure that the portable directory is not included in the spellcheck.
-          sed -i 's/find $DIRNAME/find $DIRNAME -not -path '*portable*'/g' main/tools/spell/find-unknown-comment-words
```
AniruddhaKanhere, line 63: How is the spellcheck now ignoring the 'portable' directory with this change?

.github/workflows/ci.yml
```diff
@@ -42,62 +42,20 @@ jobs:
...
-            echo "Files have trailing whitespace."
-            exit 1
-          else
-            exit 0
-          fi
+      - uses: actions/checkout@v2
+      - name: Check formatting
+        uses: FreeRTOS/CI-CD-Github-Actions/formatting@main
+        with:
+          path: ./
```
AniruddhaKanhere, line 58: If we are using the uncrustify.cfg from the CI-CD repo, then we should remove the uncrustify.cfg from this repo.

Skptak: Thanks for the catch, removed the file

.github/workflows/ci.yml
Skptak: That bot would need to be updated to use the new version of uncrustify. I also dislike that this is the only repo out of all of ours that offers that as an option. I'm going to talk to @A-Zaba about getting this added into the repo and maybe tying it into this one.

.github/workflows/ci.yml
Skptak: The spell check passes while not ignoring this directory - I am also confused as to why this directory was excluded to begin with.

moninom1: @AniruddhaKanhere  , @Skptak  I am not sure what was reasoning about skipping spell check for portable layer? Do you have any idea about it? With removing the portable layer skip i can see spell checker is failing with a lot of errors from portable layer

A-Zaba (dismissed):

.github/workflows/ci.yml
```diff
@@ -61,43 +61,21 @@ jobs:
...
-        run: |
-          set +e
-          grep --exclude="README.md" -rnI -e "[[:blank:]]$" .
-          if [ "$?" = "0" ]; then
-            echo "Files have trailing whitespace."
-            exit 1
-          else
-            exit 0
-          fi
+      - uses: actions/checkout@v2
```
ydhuang28, line 74: This is probably a nit pick but maybe we want to use `actions/checkout@v3`? It uses a more up-to-date version of Node.

Skptak: Fully agree with you, but all of our repos are using the checkout@v2 where I would prefer to do that change across all of them at once compared to individual repos.

AniruddhaKanhere (dismissed):
LGTM! Let's ship it :)

AniruddhaKanhere (approved):

A-Zaba (approved):

PR #975, opened 2023-07-20
Uncrustify Bot Check

Skptak:
Uncrustify Bot Update

Description
-----------
Fixing the uncrustify bot

Test Steps
-----------
Opened a [PR into my own fork](https://github.com/Skptak/FreeRTOS-Plus-TCP/pull/1) and ran the bot to ensure it would find a formatting issue and then fix it, after the fix made sure the formatting check passed.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/uncrustify.yml
+++ b/.github/workflows/uncrustify.yml
@@ -8,7 +8,7 @@ jobs:
   Uncrustify:
     name: Run_Uncrustify
     if: ${{ github.event.issue.pull_request && github.event.comment.body == '/bot run uncrustify' }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Dump GitHub context
         env:
@@ -16,9 +16,9 @@ jobs:
         run: echo "$GITHUB_CONTEXT"
       - name: Install Utils
         run: |
-          apt-get update && apt-get --assume-yes install software-properties-common curl jq sed
-          add-apt-repository ppa:git-core/ppa
-          apt-get update && apt-get --assume-yes install git
+          sudo apt-get update && sudo apt-get --assume-yes install software-properties-common curl jq sed
+          sudo add-apt-repository ppa:git-core/ppa
+          sudo apt-get update && sudo apt-get --assume-yes install git
           git --version
       - name: get pullrequest url
         run: |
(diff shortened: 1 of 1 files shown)


ActoryOu (approved):

.github/workflows/uncrustify.yml
```diff
@@ -8,17 +8,17 @@ jobs:
   Uncrustify:
     name: Run_Uncrustify
     if: ${{ github.event.issue.pull_request && github.event.comment.body == '/bot run uncrustify' }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
kstribrnAmzn, line 11: Why do we need to lock this to the older ubuntu version? Does the latest Ubuntu LTS not have a compatible version of uncrustify?

AniruddhaKanhere: @Skptak, please correct me if I am wrong.
@kstribrnAmzn, I believe that we need to stick with this version because uncrustify changes formatting based on which version are we using. So, if we use ubuntu-latest, then we will need to pull in a specific version of uncrustify 'manually'. If we just specify the OS version (20.04 is also LTS I believe), we will get the uncrustify we want by just doing `apt-get install`.

I understand that this is not maintainable, but so is the nature of the tool that we use. Thus, we are looking at other formatting tools which do not depend on the version of the tool we are using.

Skptak: At time of writing: [Per this run](https://github.com/Skptak/FreeRTOS-Plus-TCP/actions/runs/5616954338/job/15220209408): [Ubuntu-latest](https://github.com/Skptak/FreeRTOS-Plus-TCP/actions/runs/5616954338/workflow#L11) installs ubuntu  22.04.02, which installs [uncrustify version .72](https://github.com/Skptak/FreeRTOS-Plus-TCP/actions/runs/5616954338/job/15220209408#step:9:28)
The version used in current Formatting check, [ubuntu-20.04](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/.github/workflows/ci.yml#L72), installs [uncrustify version .69](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/5621976362/job/15233779576#step:3:21)
So this is to make sure the bot formats the code using the same version of uncrustify as the format checker uses.

kstribrnAmzn: Thanks! That's what I assumed.

kstribrnAmzn (approved):

PR #999, opened 2023-08-10
Ethernet driver for TI Hercules RM46

ChandranshuMalhotra:
Ethernet driver for TI Hercules RM46, to enable the board to communicate via TCP/IPv4 stack.

### DIFF
--- a/source/portable/NetworkInterface/RM46L852/NetworkInterface.c
+++ b/source/portable/NetworkInterface/RM46L852/NetworkInterface.c
@@ -1,0 +1,288 @@
+/*
+ * RM46L852.c
+ *
+ *  Created on: Jul 10, 2023
+ *      Author: chmalho
+ */
+
+#include "FreeRTOS.h"
+#include "list.h"
+#include "queue.h"
+#include "semphr.h"
+#include "task.h"
+
+/* FreeRTOS+TCP includes. */
+#include "FreeRTOS_IP.h"
+#include "FreeRTOS_Sockets.h"
+#include "FreeRTOS_IP_Private.h"
+#include "NetworkBufferManagement.h"
+#include "NetworkInterface.h"
+
+#include "emac.h"
+#include "mdio.h"
+#include "phy_dp83640.h"
+#include "sci.h"
+
+struct emac_tx_bdp {
+  volatile struct emac_tx_bdp *next;
+  volatile uint32 bufptr;
+  volatile uint32 bufoff_len;
+  volatile uint32 flags_pktlen;
+
+  /* helper to know which pbuf this tx bd corresponds to */
+  volatile struct NetworkBufferDescriptor_t *pbuf;
+}emac_tx_bdp;
+
+/* EMAC RX Buffer descriptor data structure */
+struct emac_rx_bdp {
+  volatile struct emac_rx_bdp *next;
+  volatile uint32 bufptr;
+  volatile uint32 bufoff_len;
+  volatile uint32 flags_pktlen;
+
+  /* helper to know which pbuf this rx bd corresponds to */
+  volatile struct NetworkBufferDescriptor_t *pbuf;
+}emac_rx_bdp;
+
+/**
+ * Helper struct to hold the data used to operate on a particular
+ * receive channel
+ */
+struct rxch {
+  volatile struct emac_rx_bdp *free_head;
+  volatile struct emac_rx_bdp *active_head;
+  volatile struct emac_rx_bdp *active_tail;
+  uint32 freed_pbuf_len;
+}rxch;
+
+/**
+ * Helper struct to hold the data used to operate on a particular
+ * transmit channel
+ */
+struct txch {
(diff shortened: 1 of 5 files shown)


source/portable/NetworkInterface/RM46L852/NetworkInterface.c
```diff
@@ -0,0 +1,369 @@
...
+ */
+
+BaseType_t xNetworkInterfaceInitialise( void )
+{
+    BaseType_t xFirstCall = pdTRUE;
+    BaseType_t xTaskCreated;
+
+    if (EMACHWInit(emacAddress) == EMAC_ERR_OK)
+    {
+        if( ( xFirstCall == pdTRUE ) || ( receiveTaskHandle == NULL ) )
```
Skptak, line 78: You create and set `xFirstCall` inside of this function. This check will always return true.

source/portable/NetworkInterface/RM46L852/NetworkInterface.c
```diff
@@ -0,0 +1,369 @@
...
+      xpbuf.len = pxNetworkBuffer->xDataLength;
+      xpbuf.tot_len = pxNetworkBuffer->xDataLength;
+      xpbuf.next = NULL;
+      if(xpbuf.tot_len < MIN_PKT_LEN)
+      {
+          xpbuf.tot_len = MIN_PKT_LEN;
+          xpbuf.len = MIN_PKT_LEN;
+      }
+
+      taskENTER_CRITICAL();
```
Skptak, line 126: At time of writing only 4 Network Interfaces in [FreeRTOS-Plus-TCP/source/portable/NetworkInterface](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/bb654636d2cccd1121aeef681f005dfe8e38ca1a/source/portable/NetworkInterface) have calls to enter or exit critical sections
The [STM32Hxx/NetworkInterface.c](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/bb654636d2cccd1121aeef681f005dfe8e38ca1a/source/portable/NetworkInterface/STM32Hxx/NetworkInterface.c#L995) interface calls it exclusively to clear option bits, which is a few assembly instructions.
The other two interfaces, the [Zynq/x_emacspif_hw.c:HandleTxErrors()](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/bb654636d2cccd1121aeef681f005dfe8e38ca1a/source/portable/NetworkInterface/Zynqx_emacpsif_hw.c#L235) and the [xilinx_ultrascale/x_emacpsif_hw.c:HandleTxErrors()]  **which actually both have the calls commented out.** Where it appears to be there for people to uncomment if attempting to de-bug an issue with a transmit error. 
Lastly the [ThirdParty/MSP432/NetworkInterface.c](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/bb654636d2cccd1121aeef681f005dfe8e38ca1a/source/portable/NetworkInterface/ThirdParty/MSP432/NetworkInterface.c#L788) which does enter and exit critical sections for its transfers. However, it does this in the task code that executes the packet transfers, not directly in the interface input and output calls. Additionally, this interface has a disclaimer about its use being untested with the current version of the FreeRTOS-Plus-TCP Stack.

**Why does this interface require the use of critical sections to transmit data when no other supported interfaces do?** This approach seems extremely heavy handed as it requires the end user of this interface to essentially set the networking actions as the single most import activity on the system.

source/portable/NetworkInterface/RM46L852/NetworkInterface.c
```diff
@@ -0,0 +1,369 @@
...
+            {
+                vReleaseNetworkBufferAndDescriptor( pxBufferDescriptor );
+                iptraceETHERNET_RX_EVENT_LOST();
+                FreeRTOS_printf( ( "RX Event Lost\n" ) );
+            }
+        }
+        else
+        {
+            #if ( ( ipconfigHAS_DEBUG_PRINTF == 1 ) && defined( FreeRTOS_debug_printf ) )
+                const EthernetHeader_t * pxEthernetHeader;
```
Skptak, line 186: It's against C coding conventions to declare variables in the middle of the function. These variables should be wrapped in an ifdef at the start of the function.

htibosch: Hello @Skptak, as the variable ` pxEthernetHeader` is only used for logging, I prefer to see a local declaration.

Normally we ( in FreeRTOS+TCP ) adhere to the rule that local declarations are OK, unless they're mixed with code, They should be placed at the beginning of a compound block:

~~~
    if ()
    {
        unsigned rc;  // good
        counter++;
    }
    if ()
    {
        counter++;
        unsigned rc;  // bad: code followed by a declaration
    }
~~~

The advantage is that the code will be easier to read. The reader doesn't have don't have to scroll back to see the declarations. Another advantage is that the scope of the variable is limited when declared in a local block.

PR #1000, opened 2023-08-10
DriverSAM: let gmac_dev_read() return the proper length (version 2)

htibosch:
Description
-----------
This PR replaces #950 which was compared to the earlier branch "dev/IPv6_integration".

Test Steps
-----------
I sent (IPv6) pings of increasing size from 0 up to to 1452 bytes, where my MTU was 1500.

    ping -c 1 -s 0 fe80::8008
    ping -c 1 -s 1 fe80::8008
    ...
    ping -c 1 -s 1452 fe80::8008  // Results in a packet of 1500 + 14 bytes.

The longest packet for ICMPv4 echo is 1472 bytes.

Checklist:
----------


source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
```diff
@@ -624,29 +624,30 @@ uint32_t gmac_dev_read( gmac_device_t * p_gmac_dev,
         return GMAC_RX_NO_DATA;
     }
 
+    /* Return the number of bytes received. */
+    *p_rcv_size = bytesLeft;
```
ActoryOu, line 628: Same question in previous PR.

Is it necessary to compare with frame size?
The `*p_rcv_size` might be different when `ul_frame_size` is smaller than `bytesLeft + 2`.
```
if( bytesLeft + 2 <= ul_frame_size )
{
    *p_rcv_size = bytesLeft;
}
else
{
    *p_rcv_size = ul_frame_size;
}
```

htibosch: About:

~~~
    left = min( bytesLeft + 2, ( int32_t ) ul_frame_size );
~~~

In fact this code is never reached. It is kept in case someone wants to avoid zero-copy reception.

The following code works equally well, I tested it by disabling `ipconfigZERO_COPY_RX_DRIVER`:

~~~
    left = bytesLeft + 2;
~~~

When `gmac_dev_read()` is called, `ul_frame_size` equals `GMAC_RX_UNITSIZE`, which is 1536 bytes. The maximum RX size is 1514 bytes: no jumbo frames are allowed.

Your example should be:
~~~diff
    if( bytesLeft + 2 <= ul_frame_size )
    {
-        *p_rcv_size = bytesLeft;
+        *p_rcv_size = bytesLeft + 2;
    }
    else
    {
        *p_rcv_size = ul_frame_size;
    }
~~~

ActoryOu: Then it's good to me.
It'd be great if this info can be added to the code comments.

ActoryOu (dismissed):

ActoryOu (approved):

source/portable/NetworkInterface/DriverSAM/gmac_SAM.c
```diff
@@ -624,29 +624,30 @@ uint32_t gmac_dev_read( gmac_device_t * p_gmac_dev,
...
+            if( p_frame != NULL )
             {
                 const uint8_t * source;
                 int32_t left;
                 int32_t toCopy;
 
                 source = gs_uc_rx_buffer + nextIdx * GMAC_RX_UNITSIZE;
-                left = bytesLeft;
+                /* Read +2 bytes because buffers are aligned at -2 bytes */
+                left = min( bytesLeft + 2, ( int32_t ) ul_frame_size );
```
htibosch, line 650: Note that this code will normally not be executed, only when `ipconfigZERO_COPY_RX_DRIVER = 1`.
We leave it here for those who want to play with the copy-method.
The driver receives frames op to 1514 bytes long. The actual value of `ul_frame_size` is 1536, so the following test is not really necessary:
~~~
left = min( bytesLeft + 2, ( int32_t ) ul_frame_size );
~~~

ActoryOu (approved):

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -687,8 +687,10 @@ static BaseType_t prvGMACInit( NetworkInterface_t * pxInterface )
     GMAC->GMAC_NCR |= GMAC_NCR_MPE;
 
     memset( &gmac_option, '\0', sizeof( gmac_option ) );
-    gmac_option.uc_copy_all_frame = 0;
-    gmac_option.uc_no_boardcast = 0;
+
+    /* Note that 'gmac_option.uc_copy_all_frame' is false, do not copy all frames.
+     * And 'gmac_option.uc_no_boardcast' is false, meaning that broadcast is received.
+     * 'boardcast' is a typo. */
```
moninom1, line 693: We can update it to uc_no_broadcast

htibosch: @moninom1 wrote:
> We can correct the name to be uc_no_broadcast

We can, but this is code provided by Microchip.

I had never noticed the typo `boardcast` until I noticed that broadcasts were not received. There is another misspelling `b_boardcast_detect`.

moninom1: Oh okay. Thanks for the info.

moninom1 (approved):

AniruddhaKanhere (approved):

Issue #1008, opened 2023-08-18
TCP zero-copy TX doesn't work.


evpopov:
Tested on: the "main" branch, aka V4.0.0 as of Aug 10 2023

From the scarce information about TCP zero copy transmit I was able to figure out that I need to get a pointer to the stream buffer, copy my data and call FreeRTOS_send() with a NULL pointer.
The issue is the code snippet below from FreeRTOS_send() :
```
        BaseType_t xByteCount = -pdFREERTOS_ERRNO_EINVAL;
        FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;

        if( pvBuffer != NULL )
        {
            /* Check if this is a valid TCP socket, affirm that it is not closed or closing,
             * affirm that there was not malloc-problem, test if uxDataLength is non-zero,
             * and if the connection is not in a confirmed FIN state. */
            xByteCount = ( BaseType_t ) prvTCPSendCheck( pxSocket, uxDataLength );
        }
        if( xByteCount > 0 )
        {
            /* prvTCPSendLoop() will try to send as many bytes as possible,
             * returning number of bytes that have been queued for transmission.. */
            xByteCount = prvTCPSendLoop( pxSocket, pvBuffer, uxDataLength, xFlags );

            if( xByteCount == 0 )
            {
                      .
                      .
                      . 
            }
        }

        return xByteCount;
```
This will never work with pvBuffer = NULL and I was always getting a return of -pdFREERTOS_ERRNO_EINVAL.
That leads to a question: Is the use of NULL pointer deprecated now? If it is still allowed, then the check above is not correct.

After commenting out the ` if( pvBuffer != NULL )` check I was able to send data in the "zero copy" way.

**Target**
- All

**Host**
- NA

**To Reproduce**
- call FreeRTOS_send() with a NULL buffer pointer

**Expected behavior**
FreeRTOS_send() should return the number of bytes sent.

htibosch:
Hi @evpopov, you are right, this doesn't work (anymore).
I will have a look at it and report back here.
Thanks for raising the issue.
Hein

htibosch:
The following seems to solve the problem:

~~~diff
-    BaseType_t xByteCount = -pdFREERTOS_ERRNO_EINVAL;
+    BaseType_t xByteCount;
     FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
 
[...]

PR #1018, opened 2023-09-05
Fixes the TCP zero-copy functionality...

evpopov:
As discussed in #1008 
It appears that this worked at some point but then got broken.
To reproduce: call FreeRTOS_send() with pvBuffer being NULL and the call always fails.

### DIFF
--- a/source/FreeRTOS_Sockets.c
+++ b/source/FreeRTOS_Sockets.c
@@ -4594,16 +4594,10 @@ void vSocketWakeUpUser( FreeRTOS_Socket_t * pxSocket )
                               size_t uxDataLength,
                               BaseType_t xFlags )
     {
-        BaseType_t xByteCount = -pdFREERTOS_ERRNO_EINVAL;
+        BaseType_t xByteCount;
         FreeRTOS_Socket_t * pxSocket = ( FreeRTOS_Socket_t * ) xSocket;
 
-        if( pvBuffer != NULL )
-        {
-            /* Check if this is a valid TCP socket, affirm that it is not closed or closing,
-             * affirm that there was not malloc-problem, test if uxDataLength is non-zero,
-             * and if the connection is not in a confirmed FIN state. */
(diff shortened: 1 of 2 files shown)


bjsowa:
Is the TCP zero-copy functionality documented anywhere? Are there any examples I can find?

evpopov:
@bjsowa 
Check these forum posts
[https://forums.freertos.org/t/releaseudppayloadbuffer-but-for-tcp/13890](https://forums.freertos.org/t/releaseudppayloadbuffer-but-for-tcp/13890) That post also has a link to the FTP code that you can also look at.
[https://forums.freertos.org/t/stm32h743-freertos-tcp-issue-with-zero-copy/15390](https://forums.freertos.org/t/stm32h743-freertos-tcp-issue-with-zero-copy/15390) This post has a bunch of gems but you may have to read through to the end.

shubnil:
Hi @evpopov , We have a new fix https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1023 in the zero copy flow. Please rebase the PR on top of the same. After that we would be good to merge.

evpopov:
@shubnil Thanks for letting me know.

evpopov:
@shubnil , that unit test that is failing because my PR allows FreeRTOS_send to be called with a NULL pointer and the unit test expects a call with a NULL pointer to fail, but it now succeeds.
The unit test should be modified, but I do not feel comfortable modifying it. Can someone with more experience help with this and disable that specific check in the unit test?

Thanks.

moninom1 (approved):

shubnil:
Hi @evpopov,

We understand the situation. We have fixed the unit test failure for this PR.
However, it will be good for long term if you can create a Linux machine setup or install a virtual Linux machine on your windows. We can help you in creating the local unit test setup.

Thanks,
Shub

shubnil (approved):

PR #1020, opened 2023-09-06
Preparing library to use loopback device

htibosch:
This PR prepares the IP-stack to use loopback addresses `127.0.0.0` and `::1`.

The issue was raised on the FreeRTOS forum [here](https://forums.freertos.org/t/freertos-tcp4-0-0-init-and-loopback-problems/18034) by [ozanagma](https://forums.freertos.org/u/ozanagma).

Description
-----------

● Routing.c :

It now recognises a new IP type for loopback: "eIPv6_Loopback"

Removed a minor bug in `FreeRTOS_AddNetworkInterface()`, and also `FreeRTOS_AddEndPoint()`.

When adding the following interfaces:
~~~
    FreeRTOS_AddNetworkInterface( &iface_a );
    FreeRTOS_AddNetworkInterface( &iface_b );
    FreeRTOS_AddNetworkInterface( &iface_a );
    FreeRTOS_AddNetworkInterface( &iface_c );
~~~
In the above example, "iface_b" will become an orphan because of this piece of code:
~~~
    if( pxInterface != NULL )
    {
        pxInterface->pxNext = NULL;
        pxInterface->pxEndPoint = NULL;
~~~

● FreeRTOS_IPv6.h :

Added `xBadIPv6Loopback()`
Added `xIsIPv6Loopback()`

● FreeRTOS_IPv4.h :

Added `xIsIPv4Loopback()`
Added `xBadIPv4Loopback()`

Communication with a loopback device always takes place within a system, and from a loopback IP address to another loopback IP.

Here is how I implemented the XOR operation:
~~~
    BaseType_t xReturn = pdFALSE;
    if( pxEndPoint != NULL )
    {
        BaseType_t x1 = ( xIsIPv4Loopback( pxIPHeader->ulDestinationIPAddress ) != 0 ) ? pdTRUE : pdFALSE;
        BaseType_t x2 = ( xIsIPv4Loopback( pxIPHeader->ulSourceIPAddress ) != 0 ) ? pdTRUE : pdFALSE;

        if( x1 != x2 )
        {
            /* Either the source or the destination address is a loopback address. */
            xReturn = pdTRUE;
        }
    }
~~~

● FreeRTOS_ND.c :

The function `FreeRTOS_CreateIPv6Address()` will allow prefix address length of 128 bits, which is necessary to allow IPv6 loopback addresses.

● FreeRTOS_IP.c :

and the most important change:

~~~diff
-    else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-             ( ipFIRST_LOOPBACK_IPv4 <= ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) ) &&
-             ( ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) < ipLAST_LOOPBACK_IPv4 ) )
+    else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
+             ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
+    {
+        /* The local loopback addresses must never appear outside a host. See RFC 1122
+         * section 3.2.1.3. */
+        eReturn = eReleaseBuffer;
+    }
~~~

● FreeRTOS_IPv6.c :

a similar changes as in FreeRTOS_IP.c here above.

Test Steps
-----------
In my next PR I will present the loopback network interface.

Also I developed a module that starts 4 tasks to do TCPv4, TCPv6, UDPv4, and UDPv6. Each task will create to sockets that will exchange data through the loopback device.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -1716,8 +1716,7 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
         eReturn = eReleaseBuffer;
     }
     else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-             ( ipFIRST_LOOPBACK_IPv4 <= ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) ) &&
-             ( ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) < ipLAST_LOOPBACK_IPv4 ) )
+             ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
     {
         /* The local loopback addresses must never appear outside a host. See RFC 1122
          * section 3.2.1.3. */
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -1707,16 +1707,6 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
     {
         eReturn = eReleaseBuffer;
     }
-
-    #if ( ipconfigUSE_IPv4 != 0 )
-        else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-                 ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
-        {
-            /* The local loopback addresses must never appear outside a host. See RFC 1122
-             * section 3.2.1.3. */
-            eReturn = eReleaseBuffer;
-        }
-    #endif
     else if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
              ( uxLength >= sizeof( UDPHeader_t ) ) )
     {
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -1702,12 +1702,15 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
     /* Note the header values required prior to the checksum
      * generation as the checksum pseudo header may clobber some of
      * these values. */
+	#if ( ipconfigUSE_IPv4 != 0 )
     if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
         ( usLength > ( FreeRTOS_ntohs( pxUDPPacket->xIPHeader.usLength ) - uxIPHeaderSizePacket( pxNetworkBuffer ) ) ) )
     {
         eReturn = eReleaseBuffer;
     }
-    else if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
+    else
+	#endif /* ( ipconfigUSE_IPv4 != 0 ) */
+	if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
              ( uxLength >= sizeof( UDPHeader_t ) ) )
     {
         size_t uxPayloadSize_1, uxPayloadSize_2;
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -1715,13 +1715,16 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
     {
         eReturn = eReleaseBuffer;
     }
-    else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-             ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
-    {
-        /* The local loopback addresses must never appear outside a host. See RFC 1122
-         * section 3.2.1.3. */
-        eReturn = eReleaseBuffer;
-    }
+
+    #if ( ipconfigUSE_IPv4 != 0 )
+        else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
+                 ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
+        {
+            /* The local loopback addresses must never appear outside a host. See RFC 1122
+             * section 3.2.1.3. */
+            eReturn = eReleaseBuffer;
+        }
+    #endif
     else if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
              ( uxLength >= sizeof( UDPHeader_t ) ) )
     {
--- a/source/FreeRTOS_IP.c
+++ b/source/FreeRTOS_IP.c
@@ -1702,16 +1702,17 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
     /* Note the header values required prior to the checksum
      * generation as the checksum pseudo header may clobber some of
      * these values. */
-	#if ( ipconfigUSE_IPv4 != 0 )
-    if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-        ( usLength > ( FreeRTOS_ntohs( pxUDPPacket->xIPHeader.usLength ) - uxIPHeaderSizePacket( pxNetworkBuffer ) ) ) )
-    {
-        eReturn = eReleaseBuffer;
-    }
-    else
-	#endif /* ( ipconfigUSE_IPv4 != 0 ) */
-	if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
-             ( uxLength >= sizeof( UDPHeader_t ) ) )
+    #if ( ipconfigUSE_IPv4 != 0 )
+        if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
+            ( usLength > ( FreeRTOS_ntohs( pxUDPPacket->xIPHeader.usLength ) - uxIPHeaderSizePacket( pxNetworkBuffer ) ) ) )
+        {
+            eReturn = eReleaseBuffer;
+        }
+        else
+    #endif /* ( ipconfigUSE_IPv4 != 0 ) */
+
+    if( ( pxNetworkBuffer->xDataLength >= uxMinSize ) &&
+        ( uxLength >= sizeof( UDPHeader_t ) ) )
     {
         size_t uxPayloadSize_1, uxPayloadSize_2;
 
--- a/source/FreeRTOS_IPv4.c
+++ b/source/FreeRTOS_IPv4.c
@@ -243,7 +243,7 @@ BaseType_t xBadIPv4Loopback( const IPHeader_t * const pxIPHeader )
 /**
  * @brief Is the IP address an IPv4 loopback address.
  *
- * @param[in] ulIPAddress The IP address being checked.
+ * @param[in] ulAddress The IP address being checked.
  *
  * @return pdTRUE if the IP address is a loopback address or else, pdFALSE.
  */
--- a/source/FreeRTOS_IPv4.c
+++ b/source/FreeRTOS_IPv4.c
@@ -313,7 +313,7 @@ enum eFrameProcessingResult prvAllowIPPacketIPv4( const struct xIP_PACKET * cons
             eReturn = eReleaseBuffer;
         }
         else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-                 ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
+                 ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) == pdTRUE ) )
         {
             /* The local loopback addresses must never appear outside a host. See RFC 1122
              * section 3.2.1.3. */
--- a/source/FreeRTOS_IPv4.c
+++ b/source/FreeRTOS_IPv4.c
@@ -312,8 +312,7 @@ enum eFrameProcessingResult prvAllowIPPacketIPv4( const struct xIP_PACKET * cons
             /* Can not handle, unknown or invalid header version. */
             eReturn = eReleaseBuffer;
         }
-        else if( ( pxIPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-                 ( xBadIPv4Loopback( &( pxIPPacket->xIPHeader ) ) == pdTRUE ) )
+        else if( xBadIPv4Loopback( &( pxIPPacket->xIPHeader ) ) == pdTRUE )
         {
             /* The local loopback addresses must never appear outside a host. See RFC 1122
              * section 3.2.1.3. */
--- a/source/FreeRTOS_IPv4.c
+++ b/source/FreeRTOS_IPv4.c
@@ -211,6 +211,58 @@ BaseType_t xIsIPv4Multicast( uint32_t ulIPAddress )
 /*-----------------------------------------------------------*/
 
 /**
+ * @brief Check if the packet is an illegal loopback packet.
+ *
+ * @param[in] pxIPHeader The IP-header being checked.
+ *
+ * @return Returns pdTRUE if the packet should be stopped, because either the source
+ *         or the target address is a loopback address.
+ */
+BaseType_t xBadIPv4Loopback( const IPHeader_t * const pxIPHeader )
+{
+    BaseType_t xReturn = pdFALSE;
+    const NetworkEndPoint_t * pxEndPoint = FreeRTOS_FindEndPointOnIP_IPv4( pxIPHeader->ulSourceIPAddress, 3 );
+
+    /* Allow loopback packets from this node itself only. */
+    if( pxEndPoint != NULL )
+    {
+        BaseType_t x1 = ( xIsIPv4Loopback( pxIPHeader->ulDestinationIPAddress ) != 0 ) ? pdTRUE : pdFALSE;
+        BaseType_t x2 = ( xIsIPv4Loopback( pxIPHeader->ulSourceIPAddress ) != 0 ) ? pdTRUE : pdFALSE;
+
+        if( x1 != x2 )
+        {
+            /* Either the source or the destination address is an IPv4 loopback address. */
+            xReturn = pdTRUE;
+        }
+    }
+
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Is the IP address an IPv4 loopback address.
+ *
+ * @param[in] ulIPAddress The IP address being checked.
+ *
+ * @return pdTRUE if the IP address is a loopback address or else, pdFALSE.
+ */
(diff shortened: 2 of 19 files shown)


source/FreeRTOS_Routing.c
```diff
@@ -241,10 +241,6 @@ struct xIPv6_Couple
     {
         NetworkEndPoint_t * pxIterator = NULL;
 
-        /* This end point will go to the end of the list, so there is no pxNext
```
moninom1, line 251: Yeah i think this needs to be removed in the case the endpoint is already present on the list.

source/FreeRTOS_Routing.c
```diff
@@ -241,10 +241,6 @@ struct xIPv6_Couple
...
-         * yet. */
-        pxEndPoint->pxNext = NULL;
-
```
moninom1, line 254: But i think we should also add pxEndPoint->pxNext = NULL; in line 262 in case there is no endpoint in the list

moninom1: Also can we move [commit](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1020/commits/91badc97b5a4073f7e43d65e35fa094acddff16c) as a separate PR?

moninom1 (approved):

source/FreeRTOS_IP.c
```diff
@@ -1707,14 +1707,16 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
-    else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-             ( ipFIRST_LOOPBACK_IPv4 <= ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) ) &&
-             ( ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) < ipLAST_LOOPBACK_IPv4 ) )
-    {
-        /* The local loopback addresses must never appear outside a host. See RFC 1122
-         * section 3.2.1.3. */
-        eReturn = eReleaseBuffer;
-    }
+
+    #if ( ipconfigUSE_IPv4 != 0 )
```
shubnil, line 1711: The ifdef (ipconfigUSE_IPv4 != 0 ) makes more sense to be at line 1705.

source/FreeRTOS_IP.c
```diff
@@ -1707,14 +1707,16 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
+        else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
+                 ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
```
shubnil, line 1713: I feel that a condition ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) == pdTRUE ) ) will probably be better for readability.

source/FreeRTOS_IPv4.c
```diff
@@ -210,6 +210,58 @@ BaseType_t xIsIPv4Multicast( uint32_t ulIPAddress )
...
+BaseType_t xBadIPv4Loopback( const IPHeader_t * const pxIPHeader )
+{
+    BaseType_t xReturn = pdFALSE;
+    const NetworkEndPoint_t * pxEndPoint = FreeRTOS_FindEndPointOnIP_IPv4( pxIPHeader->ulSourceIPAddress, 3 );
+
+    /* Allow loopback packets from this node itself only. */
+    if( pxEndPoint != NULL )
+    {
+        BaseType_t x1 = ( xIsIPv4Loopback( pxIPHeader->ulDestinationIPAddress ) != 0 ) ? pdTRUE : pdFALSE;
+        BaseType_t x2 = ( xIsIPv4Loopback( pxIPHeader->ulSourceIPAddress ) != 0 ) ? pdTRUE : pdFALSE;
```
shubnil, line 230: This logic is not very clear.Isn't it enough to check only for destination address.
 In my understanding, We should only process packets for which the destination address is loopback. Source address can only be loopback if there is a bridging/routing supported in the device and as of now our stack does not support bridging/routing. And from an external device only the destination address can be loopback. Please suggest if I am missing anything here.

source/FreeRTOS_IPv6.c
```diff
@@ -62,7 +62,7 @@ const struct xIPv6_Address FreeRTOS_in6addr_any = { 0 };
 /**
  * This variable is initialized by the system to contain the loopback IPv6 address.
  */
-const struct xIPv6_Address FreeRTOS_in6addr_loopback = { { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1 } };
+const struct xIPv6_Address FreeRTOS_in6addr_loopback = { { 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 1U } };
```
shubnil, line 65: Line no 65 and 243 both have a constant IPv6 address, however one is just const and other is static const. We should ideally have same for both.

source/FreeRTOS_IPv6.c
```diff
@@ -270,34 +262,60 @@ static void xGetIPv6MulticastGroupID( const IPv6_Address_t * pxIPv6Address,
...
-        if( ( pxEndPoint != NULL ) &&
-            ( memcmp( pxIPv6Header->xDestinationAddress.ucBytes, FreeRTOS_in6addr_loopback.ucBytes, sizeof( IPv6_Address_t ) ) == 0 ) &&
-            ( memcmp( pxIPv6Header->xSourceAddress.ucBytes, pxEndPoint->ipv6_settings.xIPAddress.ucBytes, sizeof( IPv6_Address_t ) ) == 0 ) )
+        if( pxEndPoint != NULL )
         {
-            xReturn = pdTRUE;
+            BaseType_t x1 = ( xIsIPv6Loopback( &( pxIPv6Header->xDestinationAddress ) ) != 0 ) ? pdTRUE : pdFALSE;
+            BaseType_t x2 = ( xIsIPv6Loopback( &( pxIPv6Header->xSourceAddress ) ) != 0 ) ? pdTRUE : pdFALSE;
+
+            if( x1 != x2 )
```
shubnil, line 309: This logic is not very clear.Isn't it enough to check only for destination address.
 In my understanding, We should only process packets for which the destination address is loopback. Source address can only be loopback if there is a bridging/routing supported in the device and as of now our stack does not support bridging/routing. And from an external device only the destination address can be loopback. Please suggest if I am missing anything here.

source/FreeRTOS_IPv6.c
```diff
@@ -476,10 +494,9 @@ eFrameProcessingResult_t prvAllowIPPacketIPv6( const IPHeader_IPv6_t * const pxI
             eReturn = eProcessBuffer;
         }
         /* Is it the legal multicast address? */
-        else if( ( xHasUnspecifiedAddress == pdFALSE ) &&
+        else if( ( ( xHasUnspecifiedAddress == pdFALSE ) &&
+                   ( xBadIPv6Loopback( pxIPv6Header ) == pdFALSE ) ) &&
```
shubnil, line 498: Codewise it makes more sense to have xBadIPv6Loopback and xBadIPv4Loopback checks at the same level. Here, xBadIPv4Loopback check is happening at FreeRTOS_IP.c but xBadIPv6Loopback check is in IPv6 specfic file. Can we re-arrange the code.

source/FreeRTOS_IP.c
```diff
@@ -1707,14 +1707,16 @@ static eFrameProcessingResult_t prvProcessUDPPacket( NetworkBufferDescriptor_t *
...
+        {
```
shubnil, line 1714: Codewise it more readable if we have xBadIPv6Loopback and xBadIPv4Loopback checks at the same level. Here, xBadIPv4Loopback check is happening at FreeRTOS_IP.c but xBadIPv6Loopback check is in IPv6 specfic file. Can we re-arrange the code.

shubnil (changes requested):

shubnil:
Overall, 2 main concerns:
1.  The V4 and V6 loopback check arrangement is not uniform. V4 loopback check is in 'tFreeRTOS_IP.c while V6 loopback check is inside FreeRTOS_IPv6.c. Both should be at same level, otherwise the code readability and protocol separation goes for a toss.
2. Inside the loopback checks isn't it enough to have loopback check only for destination address. Do we really need to check for source addresses as well to be loopback. If yes, Please give an example of such a used case.

source/FreeRTOS_IPv4.c
htibosch: @shubnil wrote:
~~~diff
+    if( pxEndPoint != NULL )
+    {
+        BaseType_t x1 = ( xIsIPv4Loopback( pxIPHeader->ulDestinationIPAddress ) != 0 ) ? pdTRUE : pdFALSE;
+        BaseType_t x2 = ( xIsIPv4Loopback( pxIPHeader->ulSourceIPAddress ) != 0 ) ? pdTRUE : pdFALSE;
~~~

> This logic is not very clear. Isn't it enough to check only for destination address?

The check is done in `xBadIPv4Loopback()`, which is called from `prvProcessUDPPacket()`, which will be called for any incoming UDP packet.

It will return `pdTRUE` in case either the source or the destination address is a loopback address.

> As described in [RFC1122](https://www.rfc-editor.org/rfc/rfc1122#section-3.2.1.3), Section 3.2.1.3, addresses within the entire 127.0.0.0/8 block do not legitimately appear on any network anywhere.
   
Before we had a loopback interface, loopback addresses had to be dropped:

~~~
    else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
             ( ipFIRST_LOOPBACK_IPv4 <= ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) ) &&
             ( ( FreeRTOS_ntohl( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) ) < ipLAST_LOOPBACK_IPv4 ) )
    {
        eReturn = eReleaseBuffer;
    }
~~~
which would be the same as :
~~~
    else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
               xIsIPv4Loopback( pxUDPPacket->xIPHeader.ulDestinationIPAddress ) )
    {
        /* Drop all packets travelling from or to a loopback address. */
        eReturn = eReleaseBuffer;
    }
~~~

I think that we should have tested for both `ulDestinationIPAddress` and `ulSourceIPAddress` because 127.x.x.x was not yet implemented.

Now we have added a loopback device, and so 127.x.x.x addresses must be allowed, with the exception of a packet leaving or entering the host. Loopback packets may only travel internally. So that is why:
~~~
    if( x1 != x2 )
    {
        /* Either the source or the destination address is an IPv4 loopback address. */
        xReturn = pdTRUE;
    }
~~~
when either the destination or the source address is a loopback address, the packet must be dropped.
All other packets may be processed.

source/FreeRTOS_IPv6.c
htibosch: > Line no 65 and 243 both have a constant IPv6 address, however one is just const and other is static const. We should ideally have same for both.

Good point. `xIPv6UnspecifiedAddress[]` is quite new and it is put halfway the file among the functions. The unspecified address is the same as the ANY address `FreeRTOS_in6addr_any[]`, which already exists for a long time, and which is global. Let's use that in stead.

~~~diff
     const struct xIPv6_Address FreeRTOS_in6addr_any = { 0 };
     const struct xIPv6_Address FreeRTOS_in6addr_loopback = { { 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 0U, 1U } };
-    static const struct xIPv6_Address xIPv6UnspecifiedAddress = { { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 } };
~~~

source/FreeRTOS_IP.c
htibosch: I made this test conditional:
~~~diff
+#if ( ipconfigUSE_IPv4 != 0 )
     if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
         ( usLength > ( FreeRTOS_ntohs( pxUDPPacket->xIPHeader.usLength ) - uxIPHeaderSizePacket( pxNetworkBuffer ) ) ) )
     {
         eReturn = eReleaseBuffer;
     }
     else
+#endif /* ( ipconfigUSE_IPv4 != 0 ) */
~~~

source/FreeRTOS_IP.c
htibosch: I agree, `!=` leads to more mistakes than the `==` comparison`.
~~~diff
         else if( ( pxUDPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
-                 ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) != pdFALSE ) )
+                 ( xBadIPv4Loopback( &( pxUDPPacket->xIPHeader ) ) == pdTRUE ) )
~~~

source/FreeRTOS_IP.c
htibosch: I moved the IPv4 loopback checks from `FreeRTOS_IP.c` to `FreeRTOS_IPv4.c`, see my comment below.

source/FreeRTOS_IPv6.c
htibosch: I already answered to this question here above.

source/FreeRTOS_IPv6.c
htibosch: I answered this and rearranged the code like this:

`prvAllowIPPacketIPv4()` calls `xBadIPv4Loopback()`
`prvAllowIPPacketIPv6()` calls `xBadIPv6Loopback()`

htibosch (commented):
Thanks for your review, it makes sense. I will do one more integration test after applying the mentioned changed.

moninom1 (approved):

moninom1:
Hi @htibosch , Along with fixing the Unit tests we might also need to add new tests for the new functions for 100% coverage.

source/FreeRTOS_IP.c
shubnil: Thanks

source/FreeRTOS_IP.c
shubnil: Thanks

shubnil (commented):
Thanks for the changes. Everything else looks good.
However, I would like understand a little more on Packets will loopback as Source address. What will be the use case for this? Is this for the packets originated from the loopback interface? In that case the packet should not go out of the system. Please suggest if this understanding is correct.  If this is the reason, then I am good with the change.

htibosch:
> ```diff
> section 3.2.1.3
> ```

htibosch:
> However, I would like to understand a little more on Packets will loopback as Source address. What will be the use case for this? Is this for the packets originated from the loopback interface? In that case the packet should not go out of the system.

Yes indeed, that is why the source addresses are checked as well.

In `prvAllowIPPacketIPv4()`, a packet has been received from one of the network interfaces, either from a loopback, or from an "external" interface.

When the target address is a loopback address, it must originate from a loopback interface. And vice versa: when a packet is from a loopback interface, it may only be sent to a loopback interface.

A packet from a non-loopback to a non-loopback address is OK, e.g. from `192.168.2.20` to `192.168.2.1`.

And finally, mixed-type packets, e.g. `192.168.2.100` going to `127.0.0.0`, or vice versa, should be dropped.

So that is why the XOR test is used.

unit-test / coverage: I will try to complete that.

shubnil (approved):
Thanks Hein. I am fine with the changes.

moninom1 (approved):

test/unit-test/FreeRTOS_IPv6/ut.cmake
```diff
@@ -15,7 +15,10 @@ list(APPEND mock_list
             "${MODULE_ROOT_DIR}/test/FreeRTOS-Kernel/include/queue.h"
             "${MODULE_ROOT_DIR}/test/FreeRTOS-Kernel/include/event_groups.h"
             "${CMAKE_BINARY_DIR}/Annexed_TCP/NetworkBufferManagement.h"
+			"${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_IP.h"
             "${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_IPv6_Utils.h"
+			"${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_IP_Private.h"
+			"${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_Routing.h"
```
ActoryOu, line 21:
```suggestion
            "${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_IP.h"
            "${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_IPv6_Utils.h"
            "${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_IP_Private.h"
            "${CMAKE_BINARY_DIR}/Annexed_TCP/FreeRTOS_Routing.h"
```
A minor request to fix the indent here.

test/unit-test/FreeRTOS_IPv6_ConfigDriverCheckChecksum/FreeRTOS_IPv6_ConfigDriverCheckChecksum_stubs.c
```diff
@@ -40,7 +40,7 @@
 
 /* ===========================  EXTERN VARIABLES  =========================== */
 
-extern const struct xIPv6_Address xIPv6UnspecifiedAddress;
+extern const struct xIPv6_Address FreeRTOS_in6addr_any;
```
ActoryOu, line 43:
```suggestion
```
Can we remove this? I think this is useless in this file.

ActoryOu (commented):
Left 2 minor comments.

ActoryOu (approved):

source/FreeRTOS_IPv4.c
```diff
@@ -260,6 +312,13 @@ enum eFrameProcessingResult prvAllowIPPacketIPv4( const struct xIP_PACKET * cons
             /* Can not handle, unknown or invalid header version. */
             eReturn = eReleaseBuffer;
         }
+        else if( ( pxIPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
```
tony-josi-aws, line 315: `pxIPPacket->xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE`

This checks seems to be redundant. `prvAllowIPPacketIPv4` is only called when `pxIPPacket->xEthernetHeader.usFrameType` is `ipIPv4_FRAME_TYPE` from the `prvProcessIPPacket()`.

htibosch: Yes good idea, it was tested already. I also adapted the test in `FreeRTOS_IPv4_utest.c`.

moninom1 (approved):

htibosch:
Hm, the last change affects more ut's than I expected. I will first solve that locally.

moninom1 (approved):

tony-josi-aws (approved):

shubnil (approved):

PR #1022, opened 2023-09-10
The new loopback network interface

htibosch:
Description
-----------
In the earlier IPv4-only library, there has been a loopback functionality. It needed adaptation to the network interface: outgoing packets for 127.0.0.0 would be passed back to the IP-task.

The current implementation will add a new Loopback interface. It will make sure that all "outgoing" packets will be bounced back to the IP-task. The interface will get two endpoint: one for IPv4 and one for IPv6.

Test Steps
-----------
I created a demo that starts at most 4 tasks: UDPv4, UDPv6, TCPv4, and TCPv6. Each task opens two sockets that will connect and exchanges data packets.

Checklist:
----------

Related Issue
-----------
As for now, the driver needs the changes from [PR #1020](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1020).

### DIFF
--- a/source/portable/NetworkInterface/loopback/loopbackNetworkInterface.c
+++ b/source/portable/NetworkInterface/loopback/loopbackNetworkInterface.c
@@ -1,0 +1,184 @@
+/*
+ * FreeRTOS+TCP V2.3.1
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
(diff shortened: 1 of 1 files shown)


rawalexe:
Greetings,
Thank you so much for your contribution, I'll pass this along to the team and have some one review it/

source/portable/NetworkInterface/loopback/loopbackNetworkInterface.c
```diff
@@ -0,0 +1,184 @@
...
+                if( xIsIPv6Loopback( &( pxDescriptor->xIPAddress ) ) != pdFALSE )
+                {
+                    vNDRefreshCacheEntry( &xMACAddress, &( pxDescriptor->xIPAddress.xIP_IPv6 ), pxDescriptor->pxEndPoint );
+                }
+            #endif
+        }
+        else
+        {
+            #if ( ipconfigUSE_IPv4 != 0 )
+                if( xIsIPv4Loopback( pxDescriptor->xIPAddress.ulIP_IPv4 ) )
```
tony-josi-aws, line 151: I couldn't find a definition for `xIsIPv4Loopback`, as we didn't had that function defined in the stack is it missed?

tony-josi-aws: Marking this as resolved as #1020 adds the mentioned change.

shubnil (commented):
Thanks Hein. This is a good PR. I have now started thinking what is a good way to implement a loopback:
1. Software (Network Interface Loopback) as in this PR
2. Hardware Loopback (MAC or PHY loopback) - All MAC and PHY loopback registers have standard offsets from base address and hence this code will also be generic for all Hardwares.
3. Have Both the above and give the configuration option to the user.

tony-josi-aws (approved):

shubnil:
I think we can merge the PR as a first step and create an internal ticket for adding Hardware loopback as a next step enhancement.

shubnil (approved):

htibosch:
> create an internal ticket for adding Hardware loopback

I think that "Hardware loopback" is a modus of the EMAC in which the outgoing packets are coupled back to the incoming stream. I used it sometimes as a testing mode.

The loopback network interface that I created works without hardware. The loopback is realised by sending [`eNetworkRxEvent` messages](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/loopback/loopbackNetworkInterface.c#L170) to the IP-stack.

I do not see a way in which a hardware loopback can help us realise a loopback device.

PR #1027, opened 2023-09-18
Use GCC-compatible alignment specifier in TM4C NetworkInterface.c

jonathangjertsen:
Use GCC-compatible alignment specifier in TM4C NetworkInterface.c

Description
-----------
Replace the CCS-only pragma with GCC attribute. Fixing this along with issue #1206 will make this file build without warnings on the ARM GNU toolchain.

The CCS compiler does support this attribute and the manual states that DATA_ALIGN pragma is equivalent to the aligned attribute. https://www.ti.com/lit/ug/spnu151w/spnu151w.pdf?ts=1695011722091

Test Steps
-----------

- Compile an application without the change, save the binary somewhere
- Compile with this change, but with an extra newline after the declaration to ensure `__LINE__` has the same value in each `configASSERT` statement

Check that the two binaries are identical.

Checklist:
----------

Related Issue
-----------

N/A


kstribrnAmzn:
Solid change and I love the testing details you included on your PR. I've approved and will find a PRBR shortly so we can get this change in ASAP.

kstribrnAmzn (approved):

AniruddhaKanhere (approved):

PR #1030, opened 2023-09-25
Fix network down up process

bjsowa:
Description
-----------
I couldn't find a lot of information about the process of network going down (only some of the ports even support it) but from what I understand, it should work like this:
1. When the port detects that the link is down, it informs the network stack by calling `FreeRTOS_NetworkDown` or by sending `eNetworkDownEvent` to IP task.
2. The network stack stops any ongoing DHCP or RA process, clears ARP, etc
3. The `pfInitialize` function is periodically polled by the network stack until it returns pdPASS which signals that the link went up again.
4. The DHCP/RA processes are restarted.

During port development I noticed 2 problems:
1. The DHCP/RA timers are not stopped when the network goes down. 
2. If all interfaces are up at some point, the network stack will not poll the `pfInitialize` function when an interface goes down as the `xAllNetworksUp` flag is set and network timer is not longer checked.

In this PR I fixed 1. by disabling the DHCP_RA timer for all endpoints when the network goes down. I also fixed 2. by removing `xAllNetworksUp` altogether as I didn't have any better idea. Let me know what you think.

Test Steps
-----------
Use a port that support network down event and try disconnecting it from the network and reconnecting it. If the port is implemented correctly, the DHCP process should be stopped when the link goes down and restarted when the link goes up again.

Checklist:
----------


source/FreeRTOS_IP_Timers.c
```diff
@@ -55,16 +55,6 @@
...
 #include "FreeRTOS_Routing.h"
 #include "FreeRTOS_DNS.h"
-/*-----------------------------------------------------------*/
-
-/** @brief 'xAllNetworksUp' becomes pdTRUE as soon as all network interfaces have
- * been initialised. */
-/* MISRA Ref 8.9.1 [File scoped variables] */
-/* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-89 */
-/* coverity[misra_c_2012_rule_8_9_violation] */
-static BaseType_t xAllNetworksUp = pdFALSE;
```
tony-josi-aws, line 65: Instead of removing `xAllNetworksUp` maybe we could have a set/reset API to update the state of the variable when the network goes down maybe from `prvProcessNetworkDownEvent`.
This could avoid unnecesary looping through the interfaces whenever the network timer expires.

bjsowa: I left `xAllNetworksUp` intact and added a Private API for setting it to pdFALSE.

source/FreeRTOS_IP_Utils.c
```diff
@@ -861,6 +861,22 @@ void prvProcessNetworkDownEvent( struct xNetworkInterface * pxInterface )
          * treat network down as a "delivery problem" and flush the ARP cache for this
          *  interface. */
         FreeRTOS_ClearARP( pxEndPoint );
+
+        #if ( ( ipconfigUSE_DHCP != 0 ) || ( ipconfigUSE_DHCPv6 != 0 ) )
+            if( END_POINT_USES_DHCP( pxEndPoint ) )
+            {
+                /* Stop the DHCP process for this end-point. */
+                vIPSetDHCP_RATimerEnableState( pxEndPoint, pdFALSE );
```
htibosch, line 869: When you try to end a DHCP procedure just by disabling the timer, there is a risk that the DHCP socket won't be closed.
In DHCP/IPv4, the socket has a reference count called `xDHCPSocketUserCount`. It should be closed by calling `prvCloseDHCPSocket`.
For DHCPv6, this is much the same, the socket is called `xDHCPv6Socket`, and it is protected by `xDHCPv6SocketUserCount`, and closed by calling `prvCloseDHCPv6Socket()`.

source/FreeRTOS_IP_Timers.c
```diff
@@ -604,3 +604,13 @@ void vIPSetARPResolutionTimerEnableState( BaseType_t xEnableState )
 
 #endif /* ipconfigDNS_USE_CALLBACKS == 1 */
 /*-----------------------------------------------------------*/
+
+/**
+ * @brief Mark that at least one interface is down so that the 'xNetworkTimer' is checked.
+ *        Whenever the timer expires, all interfaces that are down will get a new NetworkDown
+ *        event.
+ */
+void vSetNotAllNetworksUp( void )
```
moninom1, line 613: Can you please update the function name to be more clearer.

bjsowa: any suggestions?

source/FreeRTOS_IP_Utils.c
bjsowa: I added `vDHCPStop` and `vDHCPv6Stop` functions that, besides disabling the timer, also call `prvCloseDHCPSocket` to attempt to close the socket. 

I also noticed that IPv4 implementation of DHCP is not aware which endpoints opened the socket and calling `prvCloseDHCPSocket` will decrease `xDHCPSocketUserCount` even when the endpoint did not open it, whereas the DHCPv6 implementation stores the socket locally for each endpoint. That's why I decided to change DHCPv4 implementation to be more in line with DHCPv6.

source/FreeRTOS_IP_Timers.c
moninom1: We can make it generic, pass the argument. vSetAllNetworksUp with input parameter as xIsAllNetworksUp

bjsowa: how about now?

source/FreeRTOS_IP_Timers.c
moninom1: Yeah seems fine. You might have to add Unit Test for the new function you have added for code coverage.

source/FreeRTOS_IP_Timers.c
moninom1: Yeah seems better. You might have to add Unit Test for the new function you have added for code coverage.

bjsowa: I'll fix the existing unit tests and add more tests to cover new functions once the solution is accepted.

moninom1 (approved):

source/FreeRTOS_DHCP.c
```diff
@@ -1507,7 +1532,7 @@
         /* MISRA Ref 11.4.1 [Socket error and integer to pointer conversion] */
```
htibosch, line 1532: This MISRA comment can be safely removed. It was necessary earlier when the socket was compared with `FREERTOS_INVALID_SOCKET` directly. Now `xSocketValid()` is used in stead.

source/FreeRTOS_DHCP.c
```diff
@@ -887,6 +909,7 @@
                 xAddress.sin_port = ( uint16_t ) dhcpCLIENT_PORT_IPv4;
```
htibosch, line 909: I think it wouldn't hurt if all fields of `xAddress` will be set, as is also done in `prvCreateDHCPv6Socket()`.
~~~diff
+    memset( &xAddress, 0, sizeof( xAddress ) );
+    xAddress.sin_family = FREERTOS_AF_INET4;
+    xAddress.sin_len = ( uint8_t ) sizeof( xAddress );
     xAddress.sin_port = ( uint16_t ) dhcpCLIENT_PORT_IPv4;
~~~

htibosch (approved):
Hello @bjsowa, thank you for repairing the open/close of the DHCP client socket.

I only have two minor remarks, beside that I approve the PR.

Maybe in a future PR, we could remove the `configASSERTS()` from `prvCreateDHCPv6Socket().
They test the results of `FreeRTOS_socket()` and of `FreeRTOS_bind()`. It is not unthinkable that one of these functions will ever fail. Calling assert will just stop the application.

Thanks, Hein

source/FreeRTOS_DHCP.c
bjsowa: done

source/FreeRTOS_DHCP.c
bjsowa: done

bjsowa:
I fixed the existing unit tests and added my own for 100% coverage.

The only problem left is the CBMC proof that's failing. I would gladly appreciate someones help with it as I have no idea how to even run these tests locally.

tony-josi-aws:
@bjsowa, Thanks for the PR and fixing all the other checks. I can take care of fixing the CBMC proof failures.

moninom1 (approved):

source/FreeRTOS_DHCP.c
htibosch: > done

Thank you @bjsowa for the PR. It's good work and very much appreciated. I already approved it yesterday.
Hein

tony-josi-aws (approved):

shubnil (approved):

PR #1039, opened 2023-11-01
Fixes the allocation size when sending a neighbor solicitation packet…

evpopov:
Description
-----------
When a neighbor solicitation needs to be send, the original code at FreeRTOS_ARP.c:591 allocates
```uxNeededSize = ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + sizeof( ICMPRouterSolicitation_IPv6_t );``` Which is insufficient and incorrect. 
A few lines later, the code calls ```vNDSendNeighbourSolicitation( pxTempBuffer, pxIPAddress );```
which insists on the size being ```uxNeededSize = ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + sizeof( ICMPHeader_IPv6_t );``` and since it is not, vNDSendNeighbourSolicitation() reallocates the buffer and moves on.

The above just wastes CPU and opens the door for the second allocation to fail if the system is limited in RAM. 
The original calculated size is simply incorrect and I've substituted it with the much more eye-pleasing ```sizeof( ICMPPacket_IPv6_t );```

Note: I think I understand the occasional use of the ipSIZE_OF_XXXXX macros, but in this case mixing them with the use of sizeof() made no sense to me so I opted to go with the much easier to read sizeof( ICMPPacket_IPv6_t ). After all, that is exacly what the code will cast the pointer to, so why bother with a summation of defines.
Please double-check my findings as I did not have too much time to spend on this.

Test Steps
-----------
Happens every time an IPv6 SYN packet is received from a host that is not in the ND cache.

### DIFF
--- a/source/FreeRTOS_ARP.c
+++ b/source/FreeRTOS_ARP.c
@@ -588,7 +588,7 @@ BaseType_t xCheckRequiresARPResolution( const NetworkBufferDescriptor_t * pxNetw
                                NetworkBufferDescriptor_t * pxTempBuffer;
                                size_t uxNeededSize;
 
-                               uxNeededSize = ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + sizeof( ICMPRouterSolicitation_IPv6_t );
+                               uxNeededSize = sizeof( ICMPPacket_IPv6_t );
                                pxTempBuffer = pxGetNetworkBufferWithDescriptor( BUFFER_FROM_WHERE_CALL( 199 ) uxNeededSize, 0U );
 
                                if( pxTempBuffer != NULL )

### CODE REVIEW

**Review by shubnil — APPROVED** [review]

**Review by tony-josi-aws — APPROVED** [review]

**Review by moninom1 — APPROVED** [review]

File: source/FreeRTOS_ARP.c [hunk 1]
```diff
@@ -588,7 +588,7 @@ BaseType_t xCheckRequiresARPResolution( const NetworkBufferDescriptor_t * pxNetw
                                NetworkBufferDescriptor_t * pxTempBuffer;
                                size_t uxNeededSize;
 
-                               uxNeededSize = ipSIZE_OF_ETH_HEADER + ipSIZE_OF_IPv6_HEADER + sizeof( ICMPRouterSolicitation_IPv6_t );
+                               uxNeededSize = sizeof( ICMPPacket_IPv6_t );
```
htibosch, line 591: Yes, I am also a fan of the `sizeof()` expression.

htibosch (approved):
That is a good catch, thank you @evpopov !

PR #1040, opened 2023-11-01
Changes the ND cache full behavior when the Neighbor Discovery cache fills up.

evpopov:
Description
-----------
Imagine the following scenario:
1. The ND cache becomes full.
2. pxARPWaitingNetworkBuffer is NULL as is normally the case.
3. Our DUT ( device under test ) receives a TCP connection from an unknown IPv6 host.
4. DUT parses the SYN packet, but since it needs ND resolution and the source IP is not found in the ND cache, pxARPWaitingNetworkBuffer is pointed to the new SYN packet and the DUT sends out a Neighbor Solicitation packet.
5. The unknown host responds with a Neighbor Advertisement.
6. DUT receives the Neighbor Advertisement packet, but it cannot store the information from it in the ND cache because the DN cache is full.
7. DUT also notices that the incoming Neighbor Advertisement packet matches the SYN packet that is waiting in pxARPWaitingNetworkBuffer and as expected, queues that waiting SYN packet back to the IP task to be processed for the second time.
8. The IP task receives the message with the SYN packet and the process repeats from step 4

The above results in a storm of Neighbor Solicitation and Advertisement packets.

Test Steps
-----------

1. Set ipconfigND_CACHE_ENTRIES to a small value like 4
2. Connect to a network with multiple devices that speak IPv6 and preferably a network with just dumb switches so that there is plenty of multicast ICMPv6 traffic
3. Wait a few minutes while monitoring the cache with FreeRTOS_PrintNDCache()
4. Once the cache is full, open a TCP connection from a host not in the cache to the DUT.
5. Observe the multitude of ICMPv6 traffic.

Proposed Solution
-----------
Fist of all, this PR fixes a small typo in FreeRTOS_PrintNDCache() where ```ipconfigARP_CACHE_ENTRIES``` was being used instead of ```ipconfigND_CACHE_ENTRIES```

Second, to address the bigger issue, I changed vNDRefreshCacheEntry()'s behaviour when the cache gets full. Instead of doing nothing, this PR overwrites the oldest entry in the cache. This is not a fool-proof solution but should be fairly safe.
The issue described above is caused by a buffered packet whose parsing is delayed until the source IP is resolved. When the IP is resolved, the packet get's parsed, but there is no 100% guarantee that the ND entry is still present. This scenario allows for the packet to get delayed yet again and so on....

Another possible solution would be to detect the situation and clear out pxARPWaitingNetworkBuffer. I was not able to accomplish this cleanly and there is also another issue with that solution. The problem is that if we clear pxARPWaitingNetworkBuffer every time the ND cache gets filled, it would be impossible to open a new TCP connection as long as the cache is full and I don't like that idea.
Another "fix" would be to make sure the ND cache is very hard to fill up, but this does not solve the issue, it only postpones it.

Let me know what you think.

### DIFF
--- a/source/FreeRTOS_ND.c
+++ b/source/FreeRTOS_ND.c
@@ -285,6 +285,8 @@
     {
         BaseType_t x;
         BaseType_t xFreeEntry = -1, xEntryFound = -1;
+        BaseType_t xOldestValue = ipconfigMAX_ARP_AGE + 1;
+        BaseType_t xOldestEntry = 0;
 
         /* For each entry in the ND cache table. */
         for( x = 0; x < ipconfigND_CACHE_ENTRIES; x++ )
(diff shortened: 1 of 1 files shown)


source/FreeRTOS_ND.c
```diff
@@ -304,29 +306,45 @@
...
-            xEntryFound = xFreeEntry;
+            if( xFreeEntry >= 0 )
+            {
+                xEntryFound = xFreeEntry;
+            }
+            else
+            {
+                /* No free location. Overwrite the oldest. */
+                xEntryFound = xOldestEntry;
+                FreeRTOS_printf( ( "vNDRefreshCacheEntry: Cache FULL! Overwriting oldest entry %i with %02X-%02X-%02X-%02X-%02X-%02X\n", xEntryFound, pxMACAddress->ucBytes[ 0 ], pxMACAddress->ucBytes[ 1 ], pxMACAddress->ucBytes[ 2 ], pxMACAddress->ucBytes[ 3 ], pxMACAddress->ucBytes[ 4 ], pxMACAddress->ucBytes[ 5 ] ) );
```
tony-josi-aws, line 336: To avoid the warnings, we could consider casting `xEntryFound` to `int`.

`FreeRTOS_printf( ( "%d", ( int ) xEntryFound ) );`

evpopov: Got it. Let me know if you see anything else that you want me to change.

ActoryOu (approved):

tony-josi-aws (approved):

shubnil (approved):

PR #1041, opened 2023-11-07
Merge Buffer Allocation Files and Update Network Interfaces

HTRamsey:
Merge Buffer Allocation Files and Update Network Interfaces

Description
-----------
Use a config macro `ipconfigBUFFER_ALLOC_STATIC` to determine the allocation method instead, making it easier to maintain the code and switch between the methods. Also updates the network interfaces to use a universal static allocation method unless overridden. Dramatically reduces duplicated code while improving the functionality.

A couple tests fail because they relied on hot swapping the buffer allocation method. May need some assistance on figuring a way around that.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


source/portable/BufferManagement/BufferAllocation.c
```diff
@@ -0,0 +1,647 @@
...
+        return pxNetworkBufferCopy;
+    }
+
+#endif /* if ipconfigIS_DISABLED( ipconfigBUFFER_ALLOC_STATIC ) */
+
+/*-----------------------------------------------------------*/
+
+#if ipconfigIS_ENABLED( ipconfigBUFFER_ALLOC_STATIC )
+
+    __attribute__( ( weak ) ) void vNetworkBufferAllocateRAMToBuffers( NetworkBufferDescriptor_t pxNetworkBuffers[ ipconfigNUM_NETWORK_BUFFER_DESCRIPTORS ] )
```
tony-josi-aws, line 306: I think its better to keep the usage of compiler/target specific extensions such `weak`, `section`, etc.  as off this file to help not prevent build on compilers that doesnt support those syntax/extenstions.

htibosch: @tony-josi-aws, would you propose a macro like:
~~~
#define ipconfigWEAK_FUNCTION __attribute__( ( weak ) )

ipconfigWEAK_FUNCTION void vNetworkBufferAllocateRAMToBuffers(
                            NetworkBufferDescriptor_t pxNetworkBuffers[] )
~~~
?

In the kernel, there is a `portWEAK_SYMBOL`, but it is hardly used, only when using IAR.

Or make the declaration conditional:
~~~
#if( ipconfigINCLUDE_ALLOCATE_RAM_TO_BUFFERS_FUNCTION
    void vNetworkBufferAllocateRAMToBuffers(
          NetworkBufferDescriptor_t pxNetworkBuffers[] )
    {
    }
#endif
~~~

tony-josi-aws: Having macros might require us to add macros for all the extentions like  `weak`, `section`, `aligned`, etc thats used in the function and some compilers might follow entirely different syntax which might make it harder to maintain these macros.

I think second option is better (making the declaration conditional) or the old way of keeping the definition in the network interface file.

moninom1: I prefer the old way of keeping it in network interface file, is there any strong reason to move it to bufferallocation.c?

htibosch: Yes you're right, we also use `section` and `aligned`.

I have never heard users complain about the `__attribute__ ( ( aligned( 32 ) ) )` yet. Possibly people change the code themselves.

The same about `__attribute__( ( section( ".first_data" ) )` in the SAM project. `.first_data` is allocated in the first section of RAM, which is non-cacheable.

holden-zenith: The idea about making the definition conditional using a config value like `ipconfigINCLUDE_ALLOCATE_RAM_TO_BUFFERS_FUNCTION` was my original thought as well. If the kernel doesn't get updated to better support `portWEAK_SYMBOL` then perhaps the conditional is the better solution.

amazonKamath:
Thank you for the PR @holden-zenith. This is a good change for maintainability. However, we've identified a potential backward compatibility issue where existing applications will require updating the filenames within their build systems. We are, therefore, thinking to not take it as of now and consider it again for the next major version update.  Your thoughts?

holden-zenith:
@amazonKamath sure, makes sense. If you think this will be worthwhile in the future, should I leave it as draft or should I just close it?

HTRamsey:
Should this PR be divided up where this one doesn't touch the Network Interfaces, and instead leaves `vNetworkInterfaceAllocateRAMToBuffers` to be defined by the Network Interfaces?

tony-josi-aws:
@HTRamsey, the main challenge with taking in the changes that are part of this PR is that it breaks backward compatibility as the file path/names are changed.

So we might have to hold this change until the next major release.

HTRamsey:
@tony-josi-aws No problem with that, but my question still applies. I could separate out the parts of the PR that don't change file names, or keep the file names the same and move the common buffer allocation code elsewhere?

tony-josi-aws:
@HTRamsey, by common buffer allocation code I suppose you are referring to keeping a  common `vNetworkBufferAllocateRAMToBuffers` definition by having a new `ipconfigINCLUDE_ALLOCATE_RAM_TO_BUFFERS_FUNCTION ` macro, which could also break the backward compatibility with some of the exisitng projects that might need a custom definition for the `vNetworkBufferAllocateRAMToBuffers`.

HTRamsey:
@tony-josi-aws not quite, I was referring to saving the `vNetworkBufferAllocateRAMToBuffers` for a second PR that works on the Network Interfaces. This PR could keep the same files, keep all the code I have under `#if ( ipconfigIS_DISABLED( ipconfigBUFFER_ALLOC_STATIC ) )` in BufferAllocation_2, everything with `#if ( ipconfigIS_ENABLED( ipconfigBUFFER_ALLOC_STATIC ) )` in BufferAllocation_1, and everything else ( all the helper functions and such ) can be moved elsewhere so that you still remove all duplicated code and maintain the same file structure.

Then later on do the breaking changes if desired.

PR #1048, opened 2023-11-27
Replace ipconfigHAS_PRINTF by FreeRTOS_debug_printf in the ipARP_REPLY debug print

microcris:
Replace ipconfigHAS_PRINTF by FreeRTOS_debug_printf in the ipARP_REPLY debug print

Description
-----------
The comment in the "#endif" directive states that it belongs to "ipconfigHAS_DEBUG_PRINTF". I'm changing the "#if" directive conditional expression to "ipconfigHAS_DEBUG_PRINTF". And besides that, it is quite annoying to have "ipARP_REPLY from c0a8026dip to c0a8026dip end-point c0a80272ip" all the time int the serial console.

Test Steps
-----------
It is just print replacement.
The existing comment states that " /* Some extra logging while still testing. */". In order to have that extra logging, let us must enable the ipconfigHAS_DEBUG_PRINTF flag.

Checklist:
----------
Not applicable

Related Issue
-----------
Not applicable

### DIFF
--- a/source/FreeRTOS_ARP.c
+++ b/source/FreeRTOS_ARP.c
@@ -259,13 +259,13 @@ static TickType_t xLastGratuitousARPTime = 0U;
                 traceARP_PACKET_RECEIVED();
 
                 /* Some extra logging while still testing. */
-                #if ( ipconfigHAS_PRINTF != 0 )
+                #if ( ipconfigHAS_DEBUG_PRINTF != 0 )
                     if( pxARPHeader->usOperation == ( uint16_t ) ipARP_REPLY )
                     {
-                        FreeRTOS_printf( ( "ipARP_REPLY from %xip to %xip end-point %xip\n",
-                                           ( unsigned ) FreeRTOS_ntohl( ulSenderProtocolAddress ),
-                                           ( unsigned ) FreeRTOS_ntohl( ulTargetProtocolAddress ),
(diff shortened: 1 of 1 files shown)


tony-josi-aws (approved):

moninom1 (approved):

htibosch:
@microcris wrote:
> And besides that, it is quite annoying to have "ipARP_REPLY from c0a8026dip to c0a8026dip end-point c0a80272ip" all the time int the serial console.

As you possibly know, there is an alterative implementation of the `[vsn]printf()` functions called [printf-stdarg.c](https://github.com/htibosch/freertos_plus_projects/blob/master/plus/Common/Utilities/printf-stdarg.c). These functions are reentrant, they are thread- and interrupt-safe, and they use a small stack. They do not implement floating point numbers.

I added two special formats that allow to print IPv4 and IPv6 IP-addresses: `%xip` and `%pip`:
~~~
    uint32_t ulSourceAddress;
    printf( ( "IPv4 address is %xip\n", ulSourceAddress ) );

    IPv6_Address_t xSourceAddress;
    printf( ( "IPv6 address is %pip\n", ( void * ) xSourceAddress.ucBytes ) );

/* Output: */
    IPv4 address is 192.168.02.38
    IPv6 address is 2600:70ff:c066::2001
~~~
So the address c0a8026dip stands for 192.168.02.38

If you link printf-stdarg.c into your project, you will also see the proper IP-addresses.

htibosch (approved):
Thank you @microcris for this PR. I agree that an ARP reply is not a very interesting event, and it should be printed at debug level only.

Skptak (approved):

PR #1049, opened 2023-11-28
Fix MISRA issue

moninom1:
Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


ActoryOu (approved):

source/FreeRTOS_Sockets.c
```diff
@@ -1926,8 +1926,17 @@ BaseType_t vSocketBind( FreeRTOS_Socket_t * pxSocket,
         if( pxAddress == NULL )
         {
             pxAddress = &xAddress;
-            /* Put the port to zero to be assigned later. */
-            pxAddress->sin_port = 0U;
+            /* Clear the address: */
+            ( void ) memset( pxAddress, 0, sizeof( struct freertos_sockaddr ) );
+
+            if( pxSocket->bits.bIsIPv6 == pdFALSE_UNSIGNED )
```
htibosch, line 1932: Sorry, my bad, should be `pdTRUE_UNSIGNED`.

Does this compile with all combinations of `ipconfigUSE_IPv4/6`?

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -3119,7 +3122,14 @@
  */
 
 #ifndef FreeRTOS_flush_logging
-    #define FreeRTOS_flush_logging()    if( ipconfigHAS_PRINTF || ipconfigHAS_DEBUG_PRINTF ) do {} while( ipFALSE_BOOL )
+    #define FreeRTOS_flush_logging()                     \
```
htibosch, line 3125: I wouldn't mind if we remove `FreeRTOS_flush_logging()` totally. It is only called from `FreeRTOS_TCP_WIN.c`, and it was useful while debugging, many years ago.

source/FreeRTOS_DNS.c
```diff
@@ -1082,9 +1082,7 @@
         /* MISRA Ref 11.3.1 [Misaligned access] */
         /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-113 */
         /* coverity[misra_c_2012_rule_11_3_violation] */
-        const DNSMessage_t * pxDNSMessageHeader =
-            ( ( const DNSMessage_t * )
-              pxReceiveBuffer->pucPayloadBuffer );
+        const DNSMessage_t * pxDNSMessageHeader = ( ( const DNSMessage_t * ) pxReceiveBuffer->pucPayloadBuffer );
```
htibosch, line 1085: Can the extra brackets be removed?

~~~diff
-const DNSMessage_t * pxDNSMessageHeader = ( ( const DNSMessage_t * ) pxReceiveBuffer->pucPayloadBuffer );
+const DNSMessage_t * pxDNSMessageHeader = ( const DNSMessage_t * ) pxReceiveBuffer->pucPayloadBuffer;
~~~

htibosch (commented):
Thank you for these MISRA changes. I only had 3 minor remarks. Most importantly: shall we remove `FreeRTOS_flush_logging()`?

source/FreeRTOS_Sockets.c
moninom1: Oh i missed it, let me update it. Yes it compiles with all the combination. Thanks

source/include/FreeRTOSIPConfigDefaults.h
moninom1: Yeah removing it is also fine from MISRA point of view, also as you mentioned it is only used in TCP_WIN.

htibosch (approved):
@moninom1, I think nobody will miss the function `FreeRTOS_flush_logging()`. But it doesn't bother neither.

Thanks for this PR

ActoryOu (approved):

tony-josi-aws (approved):

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -3119,7 +3122,14 @@
...
+    if( ipconfigHAS_PRINTF || ipconfigHAS_DEBUG_PRINTF ) \
+    {                                                    \
+        do {} while( ipFALSE_BOOL );                     \
+    }                                                    \
+    else                                                 \
```
AniruddhaKanhere, line 3130: I don't think we need this else here.

AniruddhaKanhere (dismissed):
Approved with a minor comment

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -3064,7 +3067,7 @@
 
 #ifndef FreeRTOS_debug_printf
```
htibosch, line 3068: When I wrote [PR #1051](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1051), I didn't know you had planned to do the change as well.

My proposal would be like this:
~~~
#ifndef FreeRTOS_debug_printf
    #if ( ipconfigHAS_DEBUG_PRINTF == 1 ) && defined( configPRINTF )
        #define FreeRTOS_debug_printf( MSG )    do { configPRINTF( MSG ); } while( ipFALSE_BOOL )
    #else
    #endif
#endif

#ifndef FreeRTOS_printf
    #if ( ipconfigHAS_PRINTF == 1 ) && defined( configPRINTF )
        #define FreeRTOS_printf( MSG )    do { configPRINTF( MSG ); } while( ipFALSE_BOOL )
    #else
    #endif
#endif
~~~

source/include/FreeRTOSIPConfigDefaults.h
moninom1: Else is what was the MISRA error, it does not allow without an else.

htibosch: I think that coverity doesn't like this statement:
~~~
if( ipconfigHAS_PRINTF || ipconfigHAS_DEBUG_PRINTF )
~~~
because the condition is not variable.

Why not like this:
~~~
#ifndef FreeRTOS_flush_logging
    #if( ipconfigHAS_PRINTF == ipconfigENABLE ) ||
       ( ipconfigHAS_DEBUG_PRINTF == ipconfigENABLE )
        #define FreeRTOS_flush_logging() do {} while( ipFALSE_BOOL );
    #endif
#endif
~~~

moninom1: MISRA Rule 15_6 : The body of the "then" branch of the "if" statement is not a compound statement. This was the error, so with if else is expected. Sure @htibosch let me update according to your suggestion and [PR 1051](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1051)

source/include/FreeRTOSIPConfigDefaults.h
moninom1: Let me merge 1051 PR changes in this PR.

tony-josi-aws (approved):

AniruddhaKanhere (approved):

PR #782, opened 2023-03-12
Improve Default Macros

holden-zenith:
Improve Default Macros

Description
-----------

- Cleaned, organized and added documentation to both IPTraceMacroDefaults.h & FreeRTOSIPConfigDefaults.h.
- Removed a few macros that were duplicated in FreeRTOSIPConfigDefaults.h.
- Delete FreeRTOS_errno_TCP.h as all but pdFREERTOS_ERRNO_EAFNOSUPPORT have been included in projdefs.h for quite some time. A PR has been created in [FreeRTOS-Kernel](https://github.com/FreeRTOS/FreeRTOS-Kernel/pull/642/files) to add the last remaining macro. Until then, it resides in FreeRTOSIPConfigDefaults.h.
- Created a new file, FreeRTOSIPDeprecatedDefinitions.h, that contains the checks for usage of deprecated macros. This is similar to FreeRTOS-Kernel's deprecated_definitions.h.
- Added a few missing deprecated macro checks to FreeRTOSIPDeprecatedDefinitions.h.
- Include IPTraceMacroDefaults.h in FreeRTOSIPConfigDefaults.h below the tracing enabling macros and removed all inclusions of IPTraceMacroDefaults.h elsewhere. This ensures the tracing enabling macros are always defined before including the tracing implementation macros.
- Removed some unused deprecated macros (i.e. ipconfigRAND32) from the tests' FreeRTOSIPConfig.h files.

Test Steps
-----------

N/A

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [x] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

N/A

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1051, opened 2023-12-01
Remove MISRA and compiler warnings from FreeRTOSIPConfigDefaults.h

htibosch:
Description
-----------

The changes made to FreeRTOSIPConfigDefaults.h in [PR #782](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/782) were very good and useful, but they may trigger an important compiler warning:

~~~
warning: suggest braces around empty body in an 'if' statement [-Wempty-body]
~~~

In other words: it may hide the next statement when `configPRINTF` is defined as empty:
~~~
#ifndef FreeRTOS_printf
    #ifdef configPRINTF
        #define FreeRTOS_printf( MSG )    if( ipconfigHAS_PRINTF ) configPRINTF( MSG )
    #else
    #endif
#endif
~~~

And also MISRA would get upset about the if statement condition which is always true/false in
~~~
if( ipconfigHAS_PRINTF ) configPRINTF( MSG )
~~~

I would like to replace it with this:
~~~
#ifndef FreeRTOS_printf
    #if ( ipconfigHAS_PRINTF == 1 ) && defined( configPRINTF )
        #define FreeRTOS_printf( MSG )    do { configPRINTF( MSG ); } while( ipFALSE_BOOL )
    #else
    #endif
#endif

#ifndef FreeRTOS_debug_printf
    #if ( ipconfigHAS_DEBUG_PRINTF == 1 ) && defined( configPRINTF )
        #define FreeRTOS_debug_printf( MSG )    do { configPRINTF( MSG ); } while( ipFALSE_BOOL )
    #else
    #endif
#endif
~~~

Test Steps
-----------
Define `ipconfigHAS_PRINTF=1` without defining `FreeRTOS_printf()` and run the compiler.
Or define `ipconfigHAS_DEBUG_PRINTF=1` without defining `FreeRTOS_debug_printf()`.
Also interesting is to try this with an empty `configPRINTF()` in your FreeRTOSConfig.h.

Checklist:
----------

### DIFF
--- a/source/include/FreeRTOSIPConfigDefaults.h
+++ b/source/include/FreeRTOSIPConfigDefaults.h
@@ -3063,8 +3063,8 @@
 #endif
 
 #ifndef FreeRTOS_debug_printf
-    #ifdef configPRINTF
-        #define FreeRTOS_debug_printf( MSG )    if( ipconfigHAS_DEBUG_PRINTF ) configPRINTF( MSG )
+    #if ( ipconfigHAS_DEBUG_PRINTF == 1 ) && defined( configPRINTF )
+        #define FreeRTOS_debug_printf( MSG )    do { configPRINTF( MSG ); } while( ipFALSE_BOOL )
     #else
         #define FreeRTOS_debug_printf( MSG )    do {} while( ipFALSE_BOOL )
     #endif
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (approved):

AniruddhaKanhere:
The links seem to be accessible. Not sure why the link verifier is failing. Maybe a change in GitHub runners to not allow them to reach gnu.org?

Similarly, for the proof-ci, not sure why this simple change will cause 4 CBMC proofs to fail.

PR #1056, opened 2023-12-03
Network Interface Cleanup

HTRamsey:
Description
-----------
Brief attempt to clean up some of the Network Interfaces and standardize them on config values

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -106,14 +106,15 @@
 
 /*---------------------------------------------------------------------------*/
 
-/*
- * pdFREERTOS_ERRNO_EAFNOSUPPORT
```
htibosch, line 110: Something is strange here: this diff shows that FreeRTOS_IP.c has changed: a comment about `pdFREERTOS_ERRNO_EAFNOSUPPORT` was removed.

But when I compare [FreeRTOS:main:FreeRTOS_IP.c](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/FreeRTOS_IP.c) with [holden-zenith:dev-network-interface/FreeRTOS_IP.c](https://github.com/holden-zenith/FreeRTOS-Plus-TCP/blob/dev-network-interface/source/FreeRTOS_IP.c), I see zero differences.

holden-zenith: that comment was in the config defaults. I had originally intended for it to be used to check if the kernel was up to date but I don't think it ever got used anyway.

source/portable/NetworkInterface/STM32Fxx/NetworkInterface.c
```diff
@@ -628,7 +600,7 @@ BaseType_t xSTM32F_NetworkInterfaceInitialise( NetworkInterface_t * pxInterface
     }
     else
     {
-        if( xPhyObject.ulLinkStatusMask != 0U )
+        if( xSTM32F_GetPhyLinkStatus( pxInterface ) != pdFALSE )
```
htibosch, line 603: Can you please revert this change?
The driver is testing the field `ulLinkStatusMask` which is being maintained by the EMAC task.

htibosch:
Hi @holden-zenith, thank you for this clean-up PR. It looks at all network interfaces and let them use the same set of macros.

I miss a summary in "__Description__" here above.

I went through all changes and noted the following:

Cleaning up static instances of `xLLMNR_MACAddress` and `xMDNS_MacAddress`, although the latter should be spelled as `xMDNS_MACAddress`.
~~~c
    extern const MACAddress_t xMDNS_MacAddress;     // Mac
    extern const MACAddress_t xMDNS_MACAddressIPv6; // MAC
~~~

The following "local" macros will be deprecated:

`niEMAC_HANDLER_TASK_NAME`
`niEMAC_HANDLER_TASK_STACK_SIZE`
`niEMAC_HANDLER_TASK_PRIORITY`
`nwRX_TASK_STACK_SIZE`

Macros now in use:

● `EMAC_HANDLER_TASK_NAME` is defined in NetworkInterface.h as "EMAC". It would be nice if the name can be set by the user.
Also, it would be nice to see a specific EMAC name like "EMAC_STM32Fx".

● `ipconfigEMAC_TASK_STACK_SIZE` with a default of ( 8 \* configMINIMAL_STACK_SIZE ) words. See remark below.

● `ipconfigEMAC_HANDLER_TASK_PRIORITY`, defaults to high which is good.

About the stack size of the EMAC task: 
---

There was a PR #435 : "Rename configEMAC_TASK_STACK_SIZE to the more correct macro name ipconfigEMAC_TASK_STACK_SIZE_WORDS."

Mind the extension `_WORDS` in `ipconfigEMAC_TASK_STACK_SIZE_WORDS`.

The stack sizes for the EMAC tasks vary between `(2*configMINIMAL_STACK_SIZE)` and `(8*configMINIMAL_STACK_SIZE)`.
So it would be safe to set a default of 8 times the minimal stack size, except that users with little memory will complain: 4 KByte in stead of 1 KByte is used for the stack of the EMACtask. 3 KByte can be a lot!!!
We never came up with a satisfying solution, but wasting 3 KB on a small CPU is also not very satisfactory.


You replaced `EMAC_CONSIDER_FRAME_FOR_PROCESSING` with `ipCONSIDER_FRAME_FOR_PROCESSING`, but why not everywhere?

In NXP1060\NetworkInterface.c you `s/ipCONSIDER_FRAME_FOR_PROCESSING/EMAC_CONSIDER_FRAME_FOR_PROCESSING/`. That is the other way around.

In \linux\NetworkInterface.c, you change `s/EMAC_CONSIDER_FRAME_FOR_PROCESSING/ipCONSIDER_FRAME_FOR_PROCESSING/`.
In all other network interfaces, `EMAC_CONSIDER_FRAME_FOR_PROCESSING` is still used. Why?

In LPC17xx\NetworkInterface.c, you define `ipCONSIDER_FRAME_FOR_PROCESSING` without ever using it?

In the interface, use `xPhyObject.ulLinkStatusMask` in stead of calling the access function `GetPhyLinkStatus()` or so. I don't see an advantage of that. `xxxGetPhyLinkStatus()` should access mentioned field.

Added a new trace macro: `iptraceEMAC_TASK_STARTING()`. Great.

The macros EMAC_IF_RX_EVENT are now declared in `NetworkInterface.h` in stead of in the interfaces.
But STM32Hxx\NetworkInterface.c is still defining the EMAC events. Why, forgotten?

Some strange code \libslirp\MBuffNetworkInterface.c :
~~~c
    if( pxDriverCtx->xInterfaceState == pdFALSE )
    {
        pxDriverCtx->xInterfaceState = pdFALSE;


        if( xPhyObject.ulLinkStatusMask != 0U )
~~~
Next time I will give inline comments as usual.

holden-zenith:
@htibosch 

`EMAC_HANDLER_TASK_NAME` can be moved and changed to `ipconfigEMAC_HANDLER_TASK_NAME`.

`ipconfigEMAC_TASK_STACK_SIZE` can be renamed to `ipconfigEMAC_TASK_STACK_SIZE_WORDS`.

If the stack sizes are meant to be user definable than you could either undef and redefine `ipconfigEMAC_TASK_STACK_SIZE` or perhaps have a NetworkInterfaceConfigDefaults.h file that gets included at the beginning of FreeRTOSIPConfigDefaults.h which would define values such as `ipconfigEMAC_TASK_STACK_SIZE`, `ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM`, `ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM`, `ipconfigETHERNET_DRIVER_FILTERS_FRAME_TYPES`, `ipconfigETHERNET_DRIVER_FILTERS_PACKETS`, etc which are NetworkInterface dependent. Otherwise if they are just to be definable by NetworkInterface then you could set a constant value at the beginning of each NetworkInterface file, very similar to how `xBufferAllocFixedSize` is used in the two different BufferAllocation files currently.

For `EMAC_CONSIDER_FRAME_FOR_PROCESSING`, this actually didn't replace `ipCONSIDER_FRAME_FOR_PROCESSING` but is the inverse of it. One is meant to be used in the Network Interfaces and the other is used in FreeRTOS_IP.c. Currently `ipCONSIDER_FRAME_FOR_PROCESSING` is used in both instances but are defined as inverses of each other which is a bit confusing. I don't entirely follow your comments on NXP1060, linux, and NXP1060 as they all just use the `EMAC_CONSIDER_FRAME_FOR_PROCESSING` as defined in NetworkInterface.h and don't define either `ipCONSIDER_FRAME_FOR_PROCESSING` or `EMAC_CONSIDER_FRAME_FOR_PROCESSING`.

The reason I used `GetPhyLinkStatus()` instead of `xPhyObject.ulLinkStatusMask` is because not all interfaces have implemented `xPhyObject` yet and `xGetPhyLinkStatus()` is compatible when using multiple interfaces because of the `NetworkInterface_t` parameter.

You are correct about STM32Hxx and those definitions need to be deleted.
 
The code in libslirp seemed redundant so I removed the `pxDriverCtx->xInterfaceState = pdFALSE;` part after checking the interface is already down.

The general idea of all this in my mind was to slim down and make each NetworkInterface file as similar as possible to eventually use a common NetworkInterface file with the unique hardware portions defined using a custom port NetworkInterface file, similar to what I mentioned when I asked about using the CMSIS standard drivers.

holden-zenith:
@htibosch Should I break this up into smaller, more easily reviewable changes? There are a few things that don't affect running code at all.

tony-josi-aws:
@HTRamsey 

Thanks for taking time to create this PR.

>If the stack sizes are meant to be user definable than you could either undef and redefine ipconfigEMAC_TASK_STACK_SIZE or perhaps have a NetworkInterfaceConfigDefaults.h file that gets included at the beginning of FreeRTOSIPConfigDefaults.h which would define values such as ipconfigEMAC_TASK_STACK_SIZE, ipconfigDRIVER_INCLUDED_RX_IP_CHECKSUM, ipconfigDRIVER_INCLUDED_TX_IP_CHECKSUM, ipconfigETHERNET_DRIVER_FILTERS_FRAME_TYPES, ipconfigETHERNET_DRIVER_FILTERS_PACKETS, etc which are NetworkInterface dependent.

This seems to be a good idea for keeping network interface specific configurations, but as it involves adding a new file to each network interfaces it might break some of the existing projects. 

Just wondering if you had a chance to build/validate the changes on any of these interfaces.

HTRamsey:
@tony-josi-aws I can only do so for a couple of the interfaces unfortunately. There are a few changes not involving running code that I think could be made safely though, which is why I was wondering about breaking up this PR into smaller ones. 

- There are still checks for `PHY_LS_HIGH_CHECK_TIME_MS` in several of the interfaces, which is handled in FreeRTOSIPDeprecatedDefinitions.h now. 
- The definitions for `ipconfigPHY_LS_HIGH_CHECK_TIME_MS` will never be included since it is defined in FreeRTOSIPConfigDefaults.h.
- The `iptraceEMAC_TASK_STARTING();` could safely be added if it is added to IPTraceMacroDefaults.h.
- The config files I deleted are no longer needed as they don't do any configuration anymore in v4.0 and thus aren't used anywhere in the tests (the standard source files are used instead of the copies in the unit-test folder).

As far as including the standard config macros that change things like stack size and such, these should be held off until more extensive testing can be done, which unfortunately is a bit of a nightmare considering how many interfaces there are. You could do individual PRs for each interface but I believe parallel PRs like that are discouraged.

HTRamsey:
@tony-josi-aws did you have any thoughts on this? Or perhaps any other areas that I could help with?

PR #1061, opened 2023-12-10
Set ipBUFFER_PADDING to 14 bytes by default on 64 bit targets v2

htibosch:
Description
-----------
While looking at old un-merged PRs I saw [PR #547](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/547): "Set ipBUFFER_PADDING to 14 bytes by default on 64 bit targets".

Mentioned PR had the following description:
If `ipconfigBUFFER_PADDING` is not defined in FreeRTOSIPConfig.h, builds for 64 bit targets fail. Add a default value of 14 bytes `( 12U + ipconfigPACKET_FILLER_SIZE )` for `ipBUFFER_PADDING` to satisfy the alignment check in FreeRTOS_IP_Utils.c.

The change in FreeRTOS_IP.h:
~~~diff
     /* Use setting from FreeRTOS if defined and non-zero */
     #if defined( ipconfigBUFFER_PADDING ) && ( ipconfigBUFFER_PADDING != 0 )
         #define ipBUFFER_PADDING    ipconfigBUFFER_PADDING
+    #elif ( UINTPTR_MAX > 0xFFFFFFFF )
+        #define ipBUFFER_PADDING    ( 12U + ipconfigPACKET_FILLER_SIZE )
     #else
         #define ipBUFFER_PADDING    ( 8U + ipconfigPACKET_FILLER_SIZE )
     #endif
~~~

The PR was reviewed, approved, and then somehow forgotten.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -985,6 +985,8 @@ void vPreCheckConfigs( void )
     {
         size_t uxSize;
 
+        /* Check if ipBUFFER_PADDING has a minimum size, depending on the platform.
+         * See FreeRTOS_IP.h for more details. */
         #if ( UINTPTR_MAX > 0xFFFFFFFFU )
 
             /*
@@ -997,8 +999,12 @@ void vPreCheckConfigs( void )
             configASSERT( ipBUFFER_PADDING >= 10U );
         #endif /* UINTPTR_MAX > 0xFFFFFFFFU */
 
-        /* And it must have this strange alignment: */
-        configASSERT( ( ( ( ipBUFFER_PADDING ) + 2U ) % 4U ) == 0U );
+        /*
+         * The size of the Ethernet header (14) plus ipBUFFER_PADDING should be a
+         * multiple of 32 bits, in order to get aligned access to all uint32_t
+         * fields in the protocol headers.
+         */
+        configASSERT( ( ( ( ipSIZE_OF_ETH_HEADER ) + ( ipBUFFER_PADDING ) ) % 4U ) == 0U );
 
         /* LCOV_EXCL_BR_START */
         uxSize = ipconfigNETWORK_MTU;
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -983,18 +983,22 @@ void vPreCheckConfigs( void )
 
     #if ( configASSERT_DEFINED == 1 )
     {
-        volatile size_t uxSize = sizeof( uintptr_t );
+        size_t uxSize;
 
-        if( uxSize == 8U )
-        {
-            /* This is a 64-bit platform, make sure there is enough space in
-             * pucEthernetBuffer to store a pointer and also make sure that the value of
-             * ipconfigBUFFER_PADDING is such that (ipconfigBUFFER_PADDING + ipSIZE_OF_ETH_HEADER) is a
-             * 32 bit (4 byte) aligned value, so that when incrementing the ethernet buffer with
-             * (ipconfigBUFFER_PADDING + ipSIZE_OF_ETH_HEADER) bytes it lands in a 32 bit aligned address
-             * which lets us efficiently access 32 bit values later in the packet. */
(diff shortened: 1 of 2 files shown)


source/FreeRTOS_IP_Utils.c
```diff
@@ -983,18 +983,22 @@ void vPreCheckConfigs( void )
...
+             * This is a 64-bit platform, make sure there is enough space in
+             * pucEthernetBuffer to store a pointer.
+             */
+            configASSERT( ipBUFFER_PADDING >= 14U );
+        #else
+            /* This is a 32-bit platform. */
+            configASSERT( ipBUFFER_PADDING >= 10U );
+        #endif /* UINTPTR_MAX > 0xFFFFFFFFU */
+
+        /* And it must have this strange alignment: */
```
tony-josi-aws, line 1000: Can we add that the purpose of this assert is to make sure that the contents of the network packets after the ethernet header lands in a 32 bit aligned address as a comment?

I think, changing the assert to: `configASSERT( ( ( ( ipBUFFER_PADDING ) + ( ipSIZE_OF_ETH_HEADER ) ) % 4U ) == 0U );` would also make it more clear to the reader.

source/FreeRTOS_IP_Utils.c
```diff
@@ -983,18 +983,22 @@ void vPreCheckConfigs( void )
...
+        configASSERT( ( ( ( ipBUFFER_PADDING ) + 2U ) % 4U ) == 0U );
```
tony-josi-aws, line 1001: I think, changing the assert to: `configASSERT( ( ( ( ipBUFFER_PADDING ) + ( ipSIZE_OF_ETH_HEADER ) ) % 4U ) == 0U );` would also make it more clear to the reader.

source/include/FreeRTOS_IP.h
```diff
@@ -97,22 +97,29 @@ extern uint32_t ulApplicationGetNextSequenceNumber( uint32_t ulSourceAddress,
...
- *  uchar16_t type;
- *  << IP-header >>     // word-aligned
+ *
+ * +---------+-----------+---------+
+ * | Offset  | Alignment | Length  |
+ * | 32 | 64 | 32  | 64  | 32 | 64 |
+ * |----|----|-----|-----|----|----|
+ * | 0  | 0  | 4   | 8   | 4  | 8  | uchar_8 * pointer;     // Points to the 'NetworkBufferDescriptor_t'.
+ * | 4  | 8  | 4   | 8   | 2  | 2  | uchar_8   filler[];    // To give the +2 byte offset.
+ * | 6  | 10 | 4+2 | 8+2 | 6  | 6  | uchar_8   dest_mac[6]; // Destination address.
```
tony-josi-aws, line 110: In the case of 32 bit platforms the `ipBUFFER_PADDING` is `( 8U + ipconfigPACKET_FILLER_SIZE )` by default. And `ipconfigPACKET_FILLER_SIZE` is 2 by default. In that case shouldn't the `dest_mac` start at an offset of 10 bytes from the start of the network buffer. Similarly 14 bytes for 64 bit platforms?

source/FreeRTOS_IP_Utils.c
moninom1: On the same line , can we convert all the numbers used above for comparison, to a descriptive MACROs to be more clear.

source/include/FreeRTOS_IP.h
htibosch: There were indeed some typos in the table in the comments. Hopefully all correct now.

source/FreeRTOS_IP_Utils.c
htibosch: We have explained it already in other locations. I added a short comment to the asserts:
~~~diff
+    /* Check if ipBUFFER_PADDING has a minimum size, depending on the platform.
+     * See FreeRTOS_IP.h for more details. */
~~~

source/FreeRTOS_IP_Utils.c
htibosch: Good suggestion to calculate the sum of these two macros :
~~~diff
+configASSERT( ( ( ( ipSIZE_OF_ETH_HEADER ) + ( ipBUFFER_PADDING ) ) % 4U ) == 0U );
~~~
and make sure the result is a multiple of 4 bytes.

tony-josi-aws (approved):

aggarg (approved):

PR #1019, opened 2023-09-05
Multicast Support

evpopov:
I'm starting a PR in order to quit procrastinating on this subject and to align my ideas with the rest of the community. This is my proposal for how multicast can be added. Please feel free to comment.
This is nowhere near done, but it is functional and I am using it in my projects. 
If you want to to try this, you will have to modify your network driver to receive multicast packets. I only have ATSAME70 hardware so that is the only network driver that this PR updates.
As of the time of this PR's creation, the code only supports IPv4. It also supports multiple groups per socket, but this will be removed in the interest of lower CPU load. The goal is for 1 socket to only be able to subscribe to 1 multicast group.

Commit details:
Adds 2 function pointers to the network interface struct that handle adding and removing  multicast MAC addresses. Updates IGMP to use function pointers through the network interface. Makes the Add/Remove Multicast functions private to NetworkInterface.c They are now used through pointers in the NetworkInterface_t struct. Improves the SAME70 driver to handle adding/removing muticast MAC addresses

PR #1065, opened 2023-12-20
Adds NetIf functions for filtering MAC addresses

evpopov:
Description
-----------
Thanks you @htibosch for suggesting this functionality be placed in a PR of it's own, so here it is...
In summary, this PR adds functionality that can be used on it's own and is also the basis of #1019 
There are 3 parts to this PR

1. Adds `ipconfigMULTICAST_MAC_FILTERING`. When this is enabled, `NetworkInterface_t` gets two new function pointers. Those are used to control which multicast MAC addresses should be received by the EMAC hardware.
2. Adds functions to the SAME70 network driver that allow modification of what multicast MAC addresses are being received by the hardware. This implementation utilizes the 64bit hash match register that is present in the SAME70/V71 microcontrollers. Those changes to the SAME70 driver can serve as a primer for implementing the MAC filtering feature on other platforms. Please read the comments and try to contribute support for other platforms. I can only work on the SAME70 port.
3. Moves the registering of the solicited-node multicast address from the network driver to `vIPNetworkUpCalls()`. This again applies to the SAME70 driver. Other network drivers need to be updated to NOT register the solicited node MAC address.

More details about bullet point 3 above.....
IPv6 end-points that have a link-local or global address must be able to receive the so-called solicited node address. This is a multicast IPv6 address that corresponds to the end-point's own IP address. As far as I understand things, being able to receive this special multicast address allows detection of duplicate addresses. I believe the same duplicate IP detection is also used in DHCPv6 as well. In any case, the solicited-node address can change any time the end-point's IPv6 address changes and because of this we cannot "set and forget" the solicited-node MAC address in the network driver's init function.
In this PR, I have added some code to the `vIPNetworkUpCalls()`  Any time an IPv6 end-point goes UP, I calculate the solicited node multicast MAC address and instruct the network driver to begin receiving it. I have not added the complementary code that stops receiving the "old" multicast MAC. This is in my todo list.

If this PR is received well, I'm hoping to base #1019 onto this PR because #1019 depends on this code.
Hopefully, we can expand this PR with more network drivers. That would in turn ensure that #1019 also supports as many hardware platforms as possible.

p.s. I'm leaving this a work-in-progress for now and as usual, let me know what you think.

Questions 1:
According to the coding guidelines, which one of these is considered better?
```
#if ( ipconfigMULTICAST_MAC_FILTERING == ipconfigENABLE )
```
or
```
#if ( ipconfigMULTICAST_MAC_FILTERING == 1 )
```
or something else maybe?

Question 2:
I have a feeling that hiding the changes of this PR behind `ipconfigMULTICAST_MAC_FILTERING`  is kind of useless. I think eventually, all platforms should support this functionality and I don't thnk `ipconfigMULTICAST_MAC_FILTERING`  should be optional and disabled by default. Multicasts are highly needed especially with IPv6 and it feels like I'm adding a useless defines and bloating the code.
Another reason I think we don't need `ipconfigMULTICAST_MAC_FILTERING` is because even if a platform doesn't support the two functions that this PR adds, `NetworkInterface_t` is always memset to zero which causes the pointers to the functions to be NULL and every code in this PR and in #1019 checks those pointers before calling them. In essence, nothing bad will happen if the function pointers are just added without an ipconfig define.
Thoughts?


source/FreeRTOS_ND.c
```diff
@@ -77,9 +77,9 @@
     /* MISRA Ref 8.9.1 [File scoped variables] */
     /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-89 */
     /* coverity[misra_c_2012_rule_8_9_violation] */
-    static const uint8_t pcLOCAL_ALL_NODES_MULTICAST_IP[ ipSIZE_OF_IPv6_ADDRESS ] = { 0xff, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 }; /* ff02:1 */
+    static const uint8_t pcLOCAL_ALL_NODES_MULTICAST_IP[ ipSIZE_OF_IPv6_ADDRESS ] = { 0xff, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 }; /* ff02::1 */
```
htibosch, line 80: Please use the U suffix as in: `0xffU, 0x02U, 0x00U` because these are `uint8_t`.
( I know that a few more `uint8`_t arrays also need this change, like e.g. `xLLMNR_MacAddress` )

source/FreeRTOS_ND.c
```diff
@@ -77,9 +77,9 @@
...
 /** @brief All nodes on the local network segment: MAC address. */
-    static const uint8_t pcLOCAL_ALL_NODES_MULTICAST_MAC[ ipMAC_ADDRESS_LENGTH_BYTES ] = { 0x33, 0x33, 0x00, 0x00, 0x00, 0x01 };
+    const uint8_t pcLOCAL_ALL_NODES_MULTICAST_MAC[ ipMAC_ADDRESS_LENGTH_BYTES ] = { 0x33, 0x33, 0x00, 0x00, 0x00, 0x01 };
```
htibosch, line 82: Please use the U suffix also

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -3318,6 +3318,21 @@
 
 /*---------------------------------------------------------------------------*/
 
+/*
+ * ipconfigMULTICAST_MAC_FILTERING
```
htibosch, line 3322: I commented earlier about the word MULTICAST: in most cases, the hash table can also be used for unicast filtering. The SAME70 has two bits to enable 2 types of filtering:
~~~
#define GMAC_NCFGR_MTIHEN (0x1u << 6) /**< (GMAC_NCFGR) Multicast Hash Enable */
#define GMAC_NCFGR_UNIHEN (0x1u << 7) /**< (GMAC_NCFGR) Unicast Hash Enable */
~~~
I saw the same possiblity on other platforms.

So why not call it `ipconfigMAC_FILTERING`? And how should `pfAddMulticastMAC()` be called?
~~~
pfAddMulticastMAC()    => pfAddAllowedMAC()
pfRemoveMulticastMAC() => pfRemoveAllowedMAC()
~~~

And yes, we can consider to leave out the entire `ipconfigMULTICAST_MAC_FILTERING` item.

source/FreeRTOS_IP.c
```diff
@@ -631,6 +631,36 @@ TaskHandle_t FreeRTOS_GetIPTaskHandle( void )
...
+             * end-point. For unicast addresses, generate the solicited-node multicast address that corresponds
+             * to the address and generate an MLD report for it.
+             * ToDo: Figure out what the proper place is to remove multicast addresses that are no longer valid. For
+             * example when a DHCPv6 lease expires, on network down.... etc. */
+            xAddressType = xIPv6_GetIPType( &( pxEndPoint->ipv6_settings.xIPAddress ) );
+
+            if( ( xAddressType == eIPv6_LinkLocal ) || ( xAddressType == eIPv6_SiteLocal ) || ( xAddressType == eIPv6_Global ) )
+            {
+                /* Tell the network driver to begin receiving this MAC address */
+                if( pxEndPoint->pxNetworkInterface && ( pxEndPoint->pxNetworkInterface->pfAddMulticastMAC != NULL ) )
```
htibosch, line 649: Suggestion:
~~~diff
-if( pxEndPoint->pxNetworkInterface && ( pxEndPoint->pxNetworkInterface->pfAddMulticastMAC != NULL ) )
+if( ( pxEndPoint->pxNetworkInterface != NULL ) && ( pxEndPoint->pxNetworkInterface->pfAddMulticastMAC != NULL ) )
~~~
to avoid warnings

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -805,29 +819,94 @@ static uint16_t prvGenerateCRC16( const uint8_t * pucAddress )
...
+    }
+}
+
+/**
+ * @brief Removes a multicast mac address to the GMAC's hash register. Note that the
+ * MAC address will actually be removed only if the corresponding hash bit bin counter reaches zero.
+ *
+ * @param[in] pucMacAddress: A pointer to the multicast MAC Address in question.
+ */
+static void prvRemoveMulticastMACAddress( const uint8_t * pucMacAddress )
```
htibosch, line 875: I already saw this function and I couldn't have done it better.

But once again, the hash table can also be used to filter unicast addresses, so the function should get a different name, like e.g. `prvRemoveAllowedMACAddress()`.

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -805,29 +819,94 @@ static uint16_t prvGenerateCRC16( const uint8_t * pucAddress )
...
+    /* If the bin counter corresponding to this mac address is already non-zero, */
+    /* a multicast address with the same hash has already been added, so there's nothing more to do. */
+    if( 0 == prvAddressHashCounters[ ucBinNumber ] )
     {
-        /* 0 .. 31 */
-        GMAC->GMAC_HRB |= ulMask;
+        /* This bin counter is zero, so this is the first time we are registering a MAC with this hash. */
+        prvAddressHashBitMask |= ( uint64_t ) ( ( uint64_t ) 1 << ucBinNumber );
+        gmac_set_hash64( GMAC, prvAddressHashBitMask );
+        gmac_enable_multicast_hash( GMAC, true );
```
htibosch, line 859: Mind you that some Atmel implementations of `gmac_enable_multicast_hash()` are erroneous:
~~~
static inline void gmac_enable_multicast_hash(Gmac* p_gmac, uint8_t uc_enable)
{
	if (uc_enable) {
		p_gmac->GMAC_NCFGR |= GMAC_NCFGR_UNIHEN;
	} else {
		p_gmac->GMAC_NCFGR &= ~GMAC_NCFGR_UNIHEN;
	}
}
~~~
`GMAC_NCFGR_MTIHEN` should be used in stead of `GMAC_NCFGR_UNIHEN`.

The function `gmac_enable_unicast_hash()` was never created, I don't know why.

Shouldn't we enable/disable both multicast and unicast filtering?

htibosch (commented):
That looks good, thank you @evpopov.
I have a few remarks. Most importantly is that the word multicast is misleading, we adding MAC address filtering, both for unicast and multicast addresses.

source/FreeRTOS_IP.c
```diff
@@ -631,6 +631,36 @@ TaskHandle_t FreeRTOS_GetIPTaskHandle( void )
  */
 void vIPNetworkUpCalls( struct xNetworkEndPoint * pxEndPoint )
 {
+    #if ( ipconfigMULTICAST_MAC_FILTERING == ipconfigENABLE )
```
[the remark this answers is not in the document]
evpopov: I had not seen that macro. Thanks, I will start using that everywhere in my PRs!

source/FreeRTOS_ND.c
evpopov: Got it. This was someone else's code so I didn't look at the numbers at all. I just needed to remove the `static`. 
I will fix this momentarily.

source/include/FreeRTOSIPConfigDefaults.h
evpopov: WOW, during all of our discussions I have been thinking that the hash register can be used either for multicast OR unicast hash matching, but now I see what you mean. I read the datasheet again and actually tried matching a unicast MAC and you are 100% right. I will rename the functions as per your request and will add some more code in them that enables GMAC_NCFGR_MTIHEN and GMAC_NCFGR_UNIHEN when needed.

As for the ipconfigMULTICAST_MAC_FILTERING define. I will rename it and leave it for a few days in case someone comes up with a really compelling reason to keep it. If nobody does, I will gladly get rid of it.

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
evpopov: Thanks for reminding me about that one. I had a note about this and I do enable the proper multicast bit in `prvGMACInit()` but I somehow forgot to fix the calls you mentioned.

BTW, is there a good reason not to fix/add those functions in gmac_SAM.c? I know it is copied from the Advanced Software Framework, but we know these are errors for sure.... Should't I just fix them and move on with using proper functions like `gmac_enable_multicast_hash()` and `gmac_enable_unicast_hash()` instead of being forced to resort to setting the individual bits like `GMAC_NCFGR_UNIHEN`?

evpopov:
Update:
I modified the `prvAddAllowedMACAddress()` and `prvRemoveAllowedMACAddress()` functions and added a bunch of notes and documentation. The reason for the extensive comments is in case those functions get used as templates for other ports. I can of course get rid of them if the team thinks they are excessive.
I also modified `prvGMACInit()` to clean up all specific match and hash match registers and then iterate over all end-points of the network interface and call `prvAddAllowedMACAddress()` for their MAC address. That last change means that we now just have to properly fill-out our end-points and the network driver will handle the MAC addresses whether all end-points use the same MAC or unique MAC addresses.

Let me know what you all think

p.s. @htibosch is this close to how you were envisioning these filtering functions?

source/include/FreeRTOS_Routing.h
```diff
@@ -62,6 +67,25 @@
         NetworkInterfaceInitialiseFunction_t pfInitialise; /**< This function will be called upon initialisation and repeated until it returns pdPASS. */
         NetworkInterfaceOutputFunction_t pfOutput;         /**< This function is supposed to send out a packet. */
         GetPhyLinkStatusFunction_t pfGetPhyLinkStatus;     /**< This function will return pdTRUE as long as the PHY Link Status is high. */
+        #if ( ipconfigIS_ENABLED( ipconfigMAC_FILTERING ) )
+
+            /* The network driver is responsible for keeping track of which MAC addresses ( unicast or multicast )
```
htibosch, line 72: Your comment is still talking a lot about multicast, whereas the same logic applies to unicast.
I have a proposal:
~~~
/*
 * The EMAC probably has a 64-bit hash table. For every MAC-address
 * the hash value (mod 64) is calculated.  pfAddAllowedMAC() will set the bit,
 * and pfRemoveAllowedMAC() will clear the bit.  However, it may
 * happen that multiple MAC-addresses result in accessing the same bit.
 * A quick and dirty implementation option is to receive all MAC
 * addresses and set both pfAddAllowedMAC and pfRemoveAllowedMAC
 * to NULL, resulting in a promiscuous interface.
 * The optimal way to implement this functionality depends on the
 * specific hardware EMAC.
 * For one implementation option, check out:
 * "portable/NetworkInterface/DriverSAM/NetworkInterface.c".
 * It counts the number of times that a particular bit was set/cleared.
 * When the counter reaches zero, the has value will be disabled.
 */
~~~
Please correct my proposal, as your proficiency in English is surely better than mine.

source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -3295,6 +3295,21 @@
...
 /*---------------------------------------------------------------------------*/
 
+/*
+ * ipconfigMAC_FILTERING
+ *
+ * Type: BaseType_t ( ipconfigENABLE | ipconfigDISABLE )
+ *
+ * When set to ipconfigENABLE, this macro will adds two functions
+ * to the network interface. Those functions allow adding/removing multicast
+ * MAC addresses to/from the hardware EMAC. It is up to the portable network
```
htibosch, line 3305:
~~~diff
-    this macro will adds two functions
+    this macro will add two functions
~~~

~~~diff
-    adding/removing multicast MAC addresses
+    adding/removing MAC addresses
~~~

htibosch (commented):
@evpopov, although I haven't tested it yet, that looks very good to me.

I only had some proposals for change regarding the `/* */` comments.

evpopov:
Thanks for the input @htibosch 
I completely rewrote the comment describing the two filter functions. I tried striking a balance between what you wanted to say and what I wanted to say.
I also removed the ipconfigMAC_FILTERING altogether. Just to reiterate, All calls to the new functions are checked for null before the call is made, so adding the function pointers to NetworkInterface_t is safe even for ports that don't implement the filter functions yet.

I believe this PR is ready for review. 
I am not able to implement any CI testing and I'd be very thankful if someone can help with that.

htibosch:
In FreeRTOS_IP.c and in FreeRTOS_IP_Utils.c you are assuming that `ipconfigUSE_IPv6` is always enabled, please make it conditional. The variable `xAddressType` shall only be define when `ipconfigUSE_IPv6` is enabled.

FreeRTOS_IP.c
~~~diff
void vIPNetworkUpCalls( struct xNetworkEndPoint * pxEndPoint )
{
-        IPv6_Type_t xAddressType;
+    #if ( ipconfigUSE_IPv6 != 0 )
         if( pxEndPoint->bits.bIPv6 == pdTRUE_UNSIGNED )
         {
             /* Now that the network is up, pxEndPoint->ipv6_settings should hold the actual address of this
              * end-point. For unicast addresses, generate the respective solicited-node multicast address.
              * Note that the check below guards against the loopback address, the unspecified address,
              * and against the weird scenario of someone assigning a multicast address to the end-point. */
-            xAddressType = xIPv6_GetIPType( &( pxEndPoint->ipv6_settings.xIPAddress ) );
+            IPv6_Type_t xAddressType = xIPv6_GetIPType( &( pxEndPoint->ipv6_settings.xIPAddress ) );
             ...
         }
+    #endif /* ( ipconfigUSE_IPv6 != 0 ) */
~~~

FreeRTOS_IP_Utils.c
~~~diff
+    #if( ipconfigUSE_IPv6 != 0 )
         if( pxEndPoint->bits.bIPv6 == pdTRUE_UNSIGNED )
         {
-            xAddressType = xIPv6_GetIPType( &( pxEndPoint->ipv6_settings.xIPAddress ) );
+            IPv6_Type_t xAddressType = xIPv6_GetIPType( &( pxEndPoint->ipv6_settings.xIPAddress ) );
             ...
         }
+    #endif /* ( ipconfigUSE_IPv6 != 0 ) */
~~~

Unit testing: after the above changes, the UT's will run well. We do have to look at the coverage scores though:

image

evpopov:
Got it.
I'm currently working on ipv4/ipv6 separation in the multicast PR and just forgot to do it here. Thanks for reminding me.
Hopefully my changes make it better. I also added some more comments to the else cases with the hope that it helps with the unit tests.

htibosch:
@evpopov , here are changes needed to satisfy unit-testing and to reach a coverage of 100%:

[Unit-testing_PR_1065.zip](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/13853567/Unit-testing_PR_1065.zip)

tony-josi-aws:
Thanks @htibosch for the unit test fix. @evpopov can you help in updating the PR.

source/FreeRTOS_IP_Utils.c
```diff
@@ -835,6 +835,51 @@ void prvProcessNetworkDownEvent( struct xNetworkInterface * pxInterface )
     {
         /* The bit 'bEndPointUp' stays low until vIPNetworkUpCalls() is called. */
         pxEndPoint->bits.bEndPointUp = pdFALSE_UNSIGNED;
+
```
AniruddhaKanhere, line 838: Since this block of code is exactly same as the code in FreeRTOS_IP.c would it be better to just move this to a function and call it in both the places?
I am saying this because that way it becomes easier to manage just in case we ever need to change anything.
Let me know if that makes sense.

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -700,53 +701,72 @@ static BaseType_t prvGMACInit( NetworkInterface_t * pxInterface )
...
-     * OR set both to ~0H to receive all multicast packets. */
-    GMAC->GMAC_HRB = 0U; /* Hash Register Bottom. */
-    GMAC->GMAC_HRT = 0U; /* Hash Register Top. */
+    /* The call to gmac_dev_init() above writes gmac_option.uc_mac_addr to the first
+     * MAC address register by calling gmac_set_address( GMAC, 0, p_opt->uc_mac_addr )
+     * This driver however uses the prvAddAllowedMACAddress() to manipulate the specific match
+     * and hash match registers, so undo the setting of the first specific MAC register. */
+
+    /* Disable all specific MAC address match registers */
+    for( int i = 0; i < GMACSA_NUMBER; i++ )
```
AniruddhaKanhere, line 720: I'd prefer if we do not use int. It is signed and the size is undefined. Maybe use a `size_t` or `uint32_t` or even `UBaseType_t`.
Or is there a specific reason that we want an int?

AniruddhaKanhere: I'd prefer size_t over all since `i` is used as an index of an array.

AniruddhaKanhere (commented):
Thank you @evpopov for this change!
And thanks a ton @htibosch for your insightful reviews!

I have a few comments - nothing major, just some nit picks. Can you please take a look?

source/FreeRTOS_IP_Utils.c
evpopov: I agree. Let me try and clean it a bit and what it looks like with all that repeating code pulled into it's own function.

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
evpopov: I agree completely, but please help me out with the following:
In cases like that I regularly consider using UBaseType_t because it's supposed to be most efficient type for the platform, so it makes sense to use it, right? But then I should prefix the variable with 'ux' and then the very readable 'i' that a lot of programmers are used to should be called 'uxi'.
In essence I'm choosing readability because in cases like those for loops, the size and signedness doesn't really matter.
That's my reasoning. Please tell me how you feel about that silly "uxi" name. I'm out of town for a few days, but once I'm back I'll make those changes right away.

AniruddhaKanhere: Yes indeed, `uxi` will sound silly. No arguments there :)

I would not use `UBaseType_t` here though as this value is used as an index whose every possible value is supposed to fit in a `size_t` which may or may not be same as `UBaseType_t`.

To that end, I suggest we replace `i` with `xIndex`? I know that `for( int i ... )` is very familiar but indices should not be signed or they can be accidentally misused. In this case I know that we will never reach a value big enough which would cause the int to be negative, but it is a matter of consistency - which paves the path of secure and reliable code :)

Let me know if that makes sense.

htibosch: > I'd prefer if we do not use int. It is signed and the size is undefined. Maybe use a `size_t` or `uint32_t` or even `UBaseType_t`. Or is there a specific reason that we want an int?

The for-loop before C99 did not allow an inline declaration:
~~~diff
-    for( int i = 0; i < max; i++ ) {}
+    int i;
+    for( i = 0; i < max; i++ ) {}
~~~
MISRA doesn't like the use of `int`: `size_t` and eg. `uint32_t` are preferred.

The C Standard says that array indexes are (signed) integers. For a simple loop, my choice would be:
~~~
    BaseType_t xIndex;
    for( xIndex = 0; xIndex < max; xIndex++ ) {}
~~~
or when `max` is an unsigned number:
~~~
    size_t uxIndex;
    for( uxIndex = 0U; uxIndex < sizeof( xArray ); uxIndex++ ) {}
~~~

evpopov: @AniruddhaKanhere and @htibosch, Thank you both for sharing your opinion and most of all for explaining your reasoning. Please take a look at my latest commit and let me know if you like it better that way.
As far as the indexing, both of you had slightly different opinion, so I had to compromise. I went with `size_t xIndex`, but also remove one inline declaration from one of the for loops.
@AniruddhaKanhere, I consolidated the code that deals with the solicited-node addresses and MLD reports into a single function. It is a bit large and clunky but get's the job done of not repeating things like the IPv6 multicast address generation.
Let me know if you like it better this way.

source/FreeRTOS_IP_Utils.c
```diff
@@ -1764,3 +1773,131 @@ uint16_t usChar2u16( const uint8_t * pucPtr )
...
+ * @param[in] xNetworkGoingUp pdTRUE when the network goes UP, pdFALSE when the network goes DOWN.
+ */
+#if ( ipconfigIS_ENABLED( ipconfigUSE_IPv6 ) )
+    void vManageSolicitedNodeAddress( struct xNetworkEndPoint * pxEndPoint,
+                                      BaseType_t xNetworkGoingUp )
+    {
+        /* _EP_: I did my best to verify that both pxEndPoint and pxEndPoint->pxNetworkInterface cannot be
+         * NULL but I'm leaving the checks below until I get a second opinion. */
+        configASSERT( pxEndPoint != NULL );
+        configASSERT( pxEndPoint->pxNetworkInterface != NULL );
```
AniruddhaKanhere, line 1796: I think we can leave the asserts in the code. Anyways in production code the ASSERTS are supposed to be compiled out so should not be a problem 🙂

AniruddhaKanhere (commented):
Thank you for making these changes @evpopov and thanks for responding back with comments!

I have a few small nit-picks (one of them due to my mistake in providing you with wrong code - apologies).

Can you please take a look?

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -685,12 +689,10 @@ static BaseType_t prvSAM_NetworkInterfaceOutput( NetworkInterface_t * pxInterfac
 static BaseType_t prvGMACInit( NetworkInterface_t * pxInterface )
 {
     NetworkEndPoint_t * pxEndPoint;
+    size_t xIndex;
```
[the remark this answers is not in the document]
evpopov: Well, here's what [the FreeRTOS coding standard](https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html#Testing) says:
> Variables of type size_t are also prefixed x

Is FreeRTOS+TCP following a different standard? Is is written down somewhere? That could save us a lot of this back and forth.
I will address all of your comments on Monday when I get to my development PC. In the next few days, I will try to go over the entire PR again and fix as much of these small details as possible.

p.s As a side note, I'm starting to get pulled in different directions at work and may have to slow down my involvement with the PR. Hopefully we can button this up fairly soon.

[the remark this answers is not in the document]
htibosch: > Well, here's what [the FreeRTOS coding standard](https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html#Testing) says:
> 
>> Variables of type size_t are also prefixed x

I am surprised to see that. I would have expected `ux` because `size_t` is unsigned.
`UBaseType_t` variables start with `ux`:
~~~
UBaseType_t uxIndex;
~~~

> Is FreeRTOS+TCP following a different standard? Is is written down somewhere? That could save us a lot of this back and forth.

FreeRTOS+TCP and FreeRTOS+FAT were meant to be written in the FreeRTOS style. Today there are still differences, and little by little they will disappear. What we don't like are huge PR's just to change the variable prefixes. That make it difficult to compare source versions.

source/FreeRTOS_IP_Utils.c
htibosch: +1
As ` vManageSolicitedNodeAddress()` is only called internally, this `configASSERT()` is a useful and safe test for as long as we are testing.

source/include/FreeRTOS_IP_Utils.h
```diff
@@ -103,6 +103,16 @@ void vPreCheckConfigs( void );
  */
 void prvProcessNetworkDownEvent( struct xNetworkInterface * pxInterface );
 
+#if ( ipconfigIS_ENABLED( ipconfigUSE_IPv6 ) )
+
+/**
+ * @brief Called for IPv6 end-points on network up/down events.
+ * The function manages the MAC address and MLD report associated with the end-point's address.
+ */
```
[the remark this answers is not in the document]
evpopov: @AniruddhaKanhere I see I forgot the parameters which brings me to a question. Where am I supposed to put the full function description? With the declaration, with the definition, or in both places?
I didn't want to just copy/paste a  large description so I mostly wrote the decription in the .c file.
Thoughts?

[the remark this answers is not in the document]
evpopov: Oh an I just realized something else. Since `vManageSolicitedNodeAddress()` only deals with IPv6 things, maybe I should have placed it in FreeRTOS_IPv6_Utils.c
Do you agree?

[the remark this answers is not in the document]
AniruddhaKanhere: I think the description should go in the .c file based on the brief review of +TCP code that I did.

Yes, I agree that we should move the `vManageSolicitedNodeAddress()` to the IPv6_Utils file. What do you think @htibosch?

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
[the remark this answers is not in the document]
AniruddhaKanhere: For now I think we should stick to `ux` since +TCP uses that 🙂

evpopov:
Updated xIndex to uxIndex
Moved `vManageSolicitedNodeAddress()` to FreeRTOS_IPv6_Utils.c
I'm ready for the next round of nit-picking.

HTRamsey:
I am implementing the stm32 network interface multicast stuff as you go, and there are a couple things that can be used across Network Interfaces. I know this is really trivial but can we get the `ucIsMulticast = pucMacAddress[ 0 ] & 1U;` as a helper somewhere? Making `pcLOCAL_ALL_NODES_MULTICAST_MAC` accessible was another but you already got that. Maybe providing an implementation of the CRC calculation as well?

evpopov:
@HTRamsey 
We sure can. Just to make sure I understand you correctly, do you mean something like this?
```
#define isMulticastMAC( MacBytes )     ( ( MacBytes[ 0 ] & 0U ) != 0U ) 
#define isUnicastMAC( MacBytes )       ( ( MacBytes[ 0 ] & 0U ) == 0U ) 
```
If this is what you mean, I think the best place would be FreeRTOS_IP_Common.h or FreeRTOS_IP_Utils.h.... or maybe NetworkInterface.h

Now that you brought that up, I think I will also benefit from such a macro in a future PR that I'm working on.
Thoughts anyone?

HTRamsey:
> > ....Maybe providing an implementation of the CRC calculation as well?
> 
> Are you asking about the protocol checksum calculation or are you referring to the crc function that is used calculating the hash match register bit?

The hash match. I'm gonna show my naivety here but is there any reason all of the interfaces wouldn't just use CRC32 with 0x04C11DB7 as the poly? Why is DriverSam using CRC16?

htibosch:
@HTRamsey wrote:
> The hash match. I'm gonna show my naivety here but is there any reason all of the interfaces wouldn't just use CRC32 with 0x04C11DB7 as the poly? Why is DriverSam using CRC16?

Here is a third one: LPC18xx, which uses a [32-bit CRC](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/4471af64de1d481374881917bbba8736b1217530/source/portable/NetworkInterface/LPC18xx/NetworkInterface.c#L930) but with a different poly (`0xEDB88320`).
STM32 indeed uses:
~~~
/* IEEE 802.3 CRC32 polynomial - 0x04C11DB7 */
~~~
So I think that the hash/CRC calculations should stay in their networkinterface.c

HTRamsey:
@htibosch I believe that is the same as the rest, just manually writing the bit reversal. The current stm32h7 interface does it like this instead `prvRevBits32( 0x04C11DB7 )`. If they are all different in general then yeah I agree leaving it in the network interface is best but if it's like a 99% use the standard 0x04C11DB7 then I figured a helper would be nice.

evpopov:
I added a MAC unicast/multicast check macro to NetworkInterface.h as requested by @HTRamsey 
I decided to put it there because NetworkInterface.h is the lowest level I could find. Everything else is IP and up. NetworkInterface.h is also pretty empty,

While I was at it, I added a bunch of macros and defines to the SAME70 driver in order to greatly improve the readability of all hash register manipulation code.

@AniruddhaKanhere and @htibosch let me know if you disagree with my changes to NetworkInterface.h or if you feel they need improvement.

HTRamsey:
@evpopov One more suggestion, I think probably `NetworkInterface_t * pxInterface` should be a parameter to `prvAddAllowedMACAddress` and `prvRemoveAllowedMACAddress` as well like the other interface functions have. I like the idea of storing an interface's data in a struct in its pvArgument, so having the interface as a parameter could let you retrieve that data there if necessary (like the counters and index and such).

evpopov:
@HTRamsey, I don't think this is needed. These are static functions inside NetworkInterface.c and they are only exposed to the outside world through pointers inside the `NetworkInterface_t` struct. Therefore, if you somehow have a pointer to the network interface, you have pointers to those functions.
Hopefully I understood you correctly. If not, please give me some more details.

HTRamsey:
@evpopov It's just for the purpose of being able to access pvArgument inside of the NetworkInterface struct. Take a look at the Zynq NetworkInterface and you will see why this is necessary. It has multiple instances of the same interface, and stores the index in pvArgument (Although it probably should have been implemented with a struct pointer stored in pvArgument instead, especially if we assume there can be multiple types of interfaces, which would require an index per type of interface rather than a general one). So to get the particular counters, register, etc for that particular interface instance you need access to pvArgument from the interface struct.

htibosch:
@HTRamsey wrote:
> Take a look at the Zynq NetworkInterface and you will see why this is necessary.

I agree that it is useful to pass a pointer to `NetworkInterface_t` to the MAC filtering functions.
~~~c
typedef void ( * NetworkInterfaceMACFilterFunction_t )
(
    NetworkBufferDescriptor_t * const pxNetworkBuffer,
    const uint8_t * pucMacAddressBytes
);
~~~
Some devices like Xilinx have multiple independent EMAC's, each with their own MAC filter rules.

evpopov:
@HTRamsey I see. I must have been tired last night but now I understand exactly what you mean.
@htibosch I think you mean to say:
```
typedef void ( * NetworkInterfaceMACFilterFunction_t ) ( struct xNetworkInterface * pxDescriptor,
                                                         const uint8_t * pucMacAddressBytes );

``` 
but I get the idea.
@HTRamsey I agree with calling the parameter pxInterface, but for consistency with the other functions, I will stick to pxDescriptor.

Gentlemen, since this PR adds functions to the TCP stack's API, I believe it is of critical importance to get things right the first time. If we don't fix these details now we will have to deal with them as backwards compatibility when we fix them later.
I greatly appreciate this type of input.

evpopov:
@htibosch I hate to bother you, but when you have a moment, would you please take a look at the unit tests. I updated the function definitions and declarations, but I have no idea what I'm doing.
Thanks in advance!

source/include/FreeRTOS_Routing.h
```diff
@@ -54,6 +54,10 @@
 /* Return true as long as the LinkStatus on the PHY is present. */
     typedef BaseType_t ( * GetPhyLinkStatusFunction_t ) ( struct xNetworkInterface * pxDescriptor );
 
+/* Functions that manipulate what MAC addresses are received by this interface */
+    typedef void ( * NetworkInterfaceMACFilterFunction_t ) ( struct xNetworkInterface * pxDescriptor,
```
htibosch, line 58: I think you mean `pxNetworkInterface` in stead of `pxDescriptor`?

htibosch: Oops, looking at the original source ai See that this mistake is made 4 times.
I propose to replace:
~~~
struct xNetworkInterface * pxDescriptor;
~~~
with:
~~~
struct xNetworkInterface * pNetworkInterface;
~~~

evpopov: Do you mean replace it for just the MAC filtering functions or all of them including
`NetworkInterfaceInitialiseFunction_t`
`NetworkInterfaceOutputFunction_t`
`GetPhyLinkStatusFunction_t`

HTRamsey:
@evpopov you check for overflow when incrementing the hash counters but not the match counters.

evpopov:
@HTRamsey Nice find! Thanks!

amazonKamath:
@evpopov  and everyone active on this PR, we are just wrapping up a round of testing on the main HEAD and will be pushing out a v4.1.0  either this week or early next week. Till then we will close the merge window. We are excluding this PR out of v4.1.0  and will club it potentially in a next release together with your main multicast PR #1019 and some other contributions and features/bug fixes.  Thank you.

htibosch (changes requested):
Hello @evpopov , I found 3 possible mistakes in your code. Please check.
Did you check the changes when using a network interface that leaves `pfAddAllowedMAC` and `prvRemoveAllowedMACAddress` to NULL?

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -796,29 +848,273 @@ static uint16_t prvGenerateCRC16( const uint8_t * pucAddress )
...
     {
-        /* 0 .. 31 */
-        GMAC->GMAC_HRB |= ulMask;
+        if( ( prvSpecificMatchCounters[ uxIndex ] > 0U ) && ( ulSAB == GMAC->GMAC_SA[ uxIndex ].GMAC_SAB ) && ( ulSAT == GMAC->GMAC_SA[ uxIndex ].GMAC_SAT ) )
+        {
+            /* Exact match! Decrement the counter unless it's maxed out. */
+            if( prvSpecificMatchCounters[ uxIndex ] < UINT8_MAX )
+            {
+                prvSpecificMatchCounters[ uxIndex ]--;
+            }
```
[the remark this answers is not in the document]
evpopov: @htibosch This is by design. Once these counters reach UINT8_MAX they should never be decremented. The reason is that we don't know how many times we have incremented past the maximum, therefore we don't know how many times we can decrement before we clear the corresponding hash register bit or exact match register.
I thought long and hard on the size of these counter variables and I believe uint8_t is plenty large for these counters. The hash register has 64 bits and I did not want to waste RAM by having 64 uint16_t or uint32_t counters.
I can't think of any normal use case that will make these counters ever approach UINT8_MAX but please let me know if you disagree.

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -796,29 +848,273 @@ static uint16_t prvGenerateCRC16( const uint8_t * pucAddress )
...
+    ucHashBit = prvGenerateCRC16( pucMacAddress );
+    FreeRTOS_debug_printf( "prvAddAllowedMACAddress: pxIf %p, %02X-%02X-%02X-%02X-%02X-%02X hash-bit %u",
+                           pxInterface,
+                           pucMacAddress[ 0 ], pucMacAddress[ 1 ], pucMacAddress[ 2 ], pucMacAddress[ 3 ], pucMacAddress[ 4 ], pucMacAddress[ 5 ],
+                           ucHashBit );
+
+    /* Calculate what the specific match registers would look like for this MAC address so that
+     * we can check if this MAC address is already present in one of the specific match registers. */
+    ulSAB = ( pucMacAddress[ 3 ] << 24 ) | ( pucMacAddress[ 2 ] << 16 ) | ( pucMacAddress[ 1 ] << 8 ) | ( pucMacAddress[ 0 ] );
+    ulSAT = ( pucMacAddress[ 5 ] << 8 ) | ( pucMacAddress[ 4 ] );
```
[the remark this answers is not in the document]
evpopov: OK, I will cast to uint32_t

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
```diff
@@ -796,29 +848,273 @@ static uint16_t prvGenerateCRC16( const uint8_t * pucAddress )
...
+        if( prvAddressHashCounters[ ucHashBit ] > 0 )
+        {
+            /* If so many multicasts with the same hash were added that the bin counter was maxed out, */
+            /* we don't really know how many times we can decrement before actually unregistering this hash. */
+            /* Because of this, if the bin counter ever maxes out, we can never unregister the hash. */
+            if( prvAddressHashCounters[ ucHashBit ] < UINT8_MAX )
+            {
+                prvAddressHashCounters[ ucHashBit ]--;
+
+                if( 0 == prvAddressHashCounters[ ucHashBit ] )
```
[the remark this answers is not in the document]
evpopov: Same as above. This is by design.

source/portable/NetworkInterface/DriverSAM/NetworkInterface.c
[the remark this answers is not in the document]
evpopov: I have [this comment](https://github.com/evpopov/FreeRTOS-Plus-TCP/blob/d154f0e3cd9ac16ef08a54363153d70ecd46d8bb/source/portable/NetworkInterface/DriverSAM/NetworkInterface.c#L206-L214) explaining it in the code as well

source/FreeRTOS_IPv6_Utils.c
```diff
@@ -285,6 +285,141 @@ size_t usGetExtensionHeaderLength( const uint8_t * pucEthernetBuffer,
...
+
+    /* do{}while(0) to allow for the use of break statements */
+    do
+    {
+        IPv6_Type_t xAddressType = xIPv6_GetIPType( &( pxEndPoint->ipv6_settings.xIPAddress ) );
+        MACAddress_t xMACAddress = { {
+                                         0x33U,
+                                         0x33U,
+                                         0xFFU,
+                                         pxEndPoint->ipv6_settings.xIPAddress.ucBytes[ 13 ],
```
tony-josi-aws, line 314: Initializer lists must be constant compile time expressions.

source/FreeRTOS_IPv6_Utils.c
```diff
@@ -285,6 +285,141 @@ size_t usGetExtensionHeaderLength( const uint8_t * pucEthernetBuffer,
...
+        }
+        else
+        {
+            if( pxEndPoint->pxNetworkInterface->pfRemoveAllowedMAC != NULL )
+            {
+                pxEndPoint->pxNetworkInterface->pfRemoveAllowedMAC( pxEndPoint->pxNetworkInterface, xMACAddress.ucBytes );
+            }
+        }
+
+        #if ( ipconfigIS_ENABLED( ipconfigSUPPORT_IP_MULTICAST ) )
```
tony-josi-aws, line 349: Is this part of the code supposed to be added as part of PR #1019 ?

source/FreeRTOS_IPv6_Utils.c
```diff
@@ -285,6 +285,141 @@ size_t usGetExtensionHeaderLength( const uint8_t * pucEthernetBuffer,
...
+ * with it. This function manages both the MAC address and the MLD report associated with
+ * the end-point's solicited-node address. On network UP, this function registers the MAC address
+ * with the network driver's filter and creates and MLD report for the UPv6 multicast address.
+ * On network DOWN, the function unregisters the MAC address and removes the MLD report.
+ * This is a "convenience" function that keeps all these tasks under one roof for easier maintenance.
+ *
+ * @param[in] pxEndPoint The end-point for which a network up/down event is being handled.
+ * @param[in] xNetworkGoingUp pdTRUE when the network goes UP, pdFALSE when the network goes DOWN.
+ */
+void vManageSolicitedNodeAddress( struct xNetworkEndPoint * pxEndPoint,
```
tony-josi-aws, line 300: `struct xNetworkEndPoint * pxEndPoint` -> `const struct xNetworkEndPoint * pxEndPoint`

source/FreeRTOS_IPv6_Utils.c
evpopov: I rearranged things a bit to address your concern and also separate declarations and code.

source/FreeRTOS_IPv6_Utils.c
evpopov: Yes, you're correct. Code removed.

source/FreeRTOS_IPv6_Utils.c
evpopov: done

evpopov:
@tony-josi-aws sorry it took me two weeks to address your latest comments. I have been busy with other tasks lately. Let me know if there is anything else that I can change to improve this PR.

tony-josi-aws (approved):

shubnil (approved):

PR #1067, opened 2023-12-28
Stream Buffer Const Correctness

HTRamsey:
Stream Buffer Const Correctness

Description
-----------
Applies const correctness to FreeRTOS_Stream_Buffer
Removes a few unneeded imports
Small cleanup of comments

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


source/FreeRTOS_Stream_Buffer.c
```diff
@@ -53,13 +53,11 @@
  * @return The space between uxLower and uxUpper, which equals to the distance
  *         minus 1.
  */
-size_t uxStreamBufferSpace( const StreamBuffer_t * pxBuffer,
-                            const size_t uxLower,
-                            const size_t uxUpper )
+size_t uxStreamBufferSpace( const StreamBuffer_t * const pxBuffer,
+                            size_t uxLower,
+                            size_t uxUpper )
```
ActoryOu, line 58: Why do you remove const attribute for uxLower and uxUpper?

n9wxu: In general, we prefer the const qualifier where arguments are not modified inside a function.
TL;DR
MISRA advises (17.8) that arguments should not be modified.
Richard Barry reminds us that FreeRTOS gets build on 18+ compilers and sometimes these qualifiers can cause warnings in a few compilers.  So a blanket rule is difficult to apply.

source/FreeRTOS_Stream_Buffer.c
```diff
@@ -381,19 +355,20 @@ size_t uxStreamBufferGet( StreamBuffer_t * pxBuffer,
     }
 
     /* Use the minimum of the wanted bytes and the available bytes. */
-    uxCount = FreeRTOS_min_size_t( uxSize, uxMaxCount );
+    uxByteCount = FreeRTOS_min_size_t( uxSize, uxByteCount );
```
htibosch, line 358: Same remark as earlier, please do not change the value of the function parameter `uxByteCount`.

source/FreeRTOS_Stream_Buffer.c
```diff
@@ -281,20 +256,21 @@ size_t uxStreamBufferAdd( StreamBuffer_t * pxBuffer,
 
     /* The number of bytes that can be written is the minimum of the number of
      * bytes requested and the number available. */
-    uxCount = FreeRTOS_min_size_t( uxSpace, uxCount );
+    uxByteCount = FreeRTOS_min_size_t( uxSpace, uxByteCount );
```
htibosch, line 259: MISRA doesn't like when we change the value of a parameter, like here in :
~~~
    uxByteCount = FreeRTOS_min_size_t( uxSpace, uxByteCount );
~~~
That is why the help variable `uxCount` was introduced.

source/FreeRTOS_Stream_Buffer.c
```diff
@@ -174,24 +159,18 @@ void vStreamBufferClear( StreamBuffer_t * pxBuffer )
...
-                           size_t uxCount )
+void vStreamBufferMoveMid( StreamBuffer_t * const pxBuffer,
+                           const size_t uxCount )
 {
     /* Increment uxMid, but no further than uxHead */
-    size_t uxSize = uxStreamBufferMidSpace( pxBuffer );
-    size_t uxMid = pxBuffer->uxMid;
-    size_t uxMoveCount = uxCount;
+    const size_t uxSize = uxStreamBufferMidSpace( pxBuffer );
+    const size_t uxMoveCount = ( uxCount > uxSize ) ? uxSize : uxCount;
```
htibosch, line 167: I prefer the original implementation, which was easier to read. Conditional operators make the code "more dense", and harder to debug.

htibosch (commented):
Thank you @holden-zenith for adding the const keywords.

This addition is also good:
~~~
    size_t uxLength = pxBuffer->LENGTH;
    ...
    if( uxCount >= uxLength )
    {
        uxCount -= uxLength;
    }
~~~
Although `LENGTH` will never change, it is good to use a copy.

FreeRTOS_Stream_Buffer.c has never received much attention since its creation. It just worked. We did add unit-testing and MISRA checks though.

I have 3 remarks, most importantly is follow the MISRA rule that forbids to change the value of a parameter.

source/FreeRTOS_Stream_Buffer.c
holden-zenith: So this relates to what I was asking @ActoryOu, is there any reason to have const written for pass by value parameters? For example, there is 
`size_t uxStreamBufferSpace( const StreamBuffer_t * const pxBuffer,
                            const size_t uxLower,
                            const size_t uxUpper );`
but if the uxLower & uxUpper parameters can't be changed anyways due to MISRA, then I suppose there is no value in having the const for uxLower and uxUpper? I'm just trying to keep things consistent as best as possible.

htibosch: You wrote:

> but if the uxLower & uxUpper parameters can't be changed anyways due to MISRA, then I suppose there is no value in having the const for uxLower and uxUpper?

I agree that it doesn't add much value to add the const keyword to parameters that are passed by-value. But it doesn't hurt either.

source/FreeRTOS_Stream_Buffer.c
ActoryOu: As @n9wxu said, that may not add much value to add const to the parameter that is called by value. But it also violates MISRA rule. I'd suggest to add it back.

holden-zenith: @ActoryOu for consistency's sake, should all pass by value parameters be marked as const then? I guess I'm just curious why a couple of these functions have const parameters but most others dont.

ActoryOu: Applogies that I take this discussion into wrong direction.

After checking, I was saying `MISRA rule 8.13 -- A pointer should point to a const-qualified type whenever possible` and that does not relate to call by value parameter. So, I'm ok to remove that.

I'd suggest to add this change log into PR description to summarize what you're doing in this PR. IMHO, it's going far comparing to what it's saying.

Thank you.

source/FreeRTOS_Stream_Buffer.c
moninom1: Yeah just to add, it violates MISRA 17.8 A function parameters should not be modified. 
So it is expected to copy the param to local object and modify it. Hence uxCount was introduced.

tony-josi-aws (approved):

moninom1 (approved):

ActoryOu (approved):

AniruddhaKanhere (approved):
Thank you @HTRamsey for these changes. 
They look good and I am approving them!

PR #1068, opened 2023-12-30
Fix declaration of socket functions without TCP

HTRamsey:
Fix declaration of socket ID functions without TCP

Description
-----------
These functions were limited to when TCP was enabled only, but weren't reliant on TCP.
Side note - Shouldn't these functions be named like an API function such as FreeRTOS_SetSocketID/FreeRTOS_GetSocketID?
And perhaps there should be a config like ipconfigSUPPORT_SOCKET_ID or something to remove the void * pvSocketID in the xSOCKET struct if not needed.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


htibosch (approved):
@holden-zenith , thank for noticing this. You are right, socket ID's can also be used in UDP-only applications.

The socket ID is used in a wide variety of situations. UDP sockets also have callbacks, and users can assign an ID.

C++ programmers will typically assign the address of an object to the socket ID.

This feature is always available, it does not depend on some ipconfig flag. We decided to do so because the memory foot print is extremely small, and we didn't like yet another `#if`.

> Shouldn't these functions be named like an API function such as FreeRTOS_SetSocketID/FreeRTOS_GetSocketID?

Yes maybe we should have done so, although the name would get long: `FreeRTOS_SocketGetSocketID()`.

The `FreeRTOS_` prefix was originally only used for the standard BSD functions:
~~~
    FreeRTOS_socket()
    FreeRTOS_listen()
    FreeRTOS_bind()
    FreeRTOS_recv()
    FreeRTOS_send()
~~~
I often have ported a program to Linux by defining :
~~~
    #define FreeRTOS_socket  socket
    #define FreeRTOS_listen  listen
~~~

and vice versa:

~~~
    #define socket  FreeRTOS_socket
    #define listen  FreeRTOS_listen
~~~

tony-josi-aws (approved):

Issue #1077, opened 2024-01-19
[BUG] The endpoint with the NULL address is operated in prvTCPReturnPacket_IPV4 function


XIAO-Q7:
**Describe the bug**

When a endpoint enables DHCP and sets a AUTO IP. Receiving a tcp socket packet from a non-AUTOIP network segment causes the endpoint with a null address to be operated on in the prvTCPReturnPacket_IPV4 function.

**Target**

Any

**To Reproduce**
![Plus_TCP_BUG](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/assets/62824695/281294a7-3f64-4f05-9d0c-1ddd32798084)

- Configure: An endpoint enable DHCP and set a AUTO IP

In my case, I'm running Plus TCP on an STM32 and built a TCP echo server, employing a computer as the echo client. When I send an echo frame to the server, the microcontroller runs into HardFault_Handler.

**Expected behavior**

Server correctly returns echo frames.

I have set a special IP address for the computer in this case, not the AUTOIP segment. If PLUS TCP can't automatically find an entry point to send frames. It can not send the frame. I don't want the chip to enter HardFault_Handler.

**Additional context**

Through my debugging, I found that in the prvTCPReturnPacket_IPV4 function, the line of code here gives an endpoint to a null address.

```C
eResult = eARPGetCacheEntry( &ulDestinationIPAddress, &xMACAddress, &( pxNetworkBuffer->pxEndPoint ) );
```

and later operates this endpoint with a null address to send frames, eventually causing the chip to run into HardFault_Handler.

```C
NetworkInterface_t * pxInterface = pxNetworkBuffer->pxEndPoint->pxNetworkInterface;
( void ) pxInterface->pfOutput( pxInterface, pxNetworkBuffer, xDoRelease );
```

htibosch:
Hello @XIAO-Q7 , thanks a lot for reporting this.
I'm trying to reproduce it, and a fix is on its way.
Thanks

tony-josi-aws:
@XIAO-Q7, we have merged #1078 that should be fixing the issue reported, if possible, please verify if the fix works for you by pulling the latest changes from main.

XIAO-Q7:
Yes, it worked. Thank you for your help.

tony-josi-aws:
Thanks for verifying.

PR #1078, opened 2024-01-22
Do not clear pxEndPoint in TCPReturnPacket()

htibosch:
Description
-----------
In response to issue #1077 : 

@XIAO-Q7, thank you for reporting the issue and for the research that you did. Also thanks to @evpopov who encountered the same problem.

I managed to reproduce the problem if I use the following configuration:

- DUT has IP-address 192.168.2.127
- DUT has no gateway (0.0.0.0)
- The TCP remote party has IP address 172.16.0.1

DUT tries to connect to remote device, but `eARPGetCacheEntry()` clears the endpoint, which leads to a dereference of NULL.

Here is a summary of the proposed change:
~~~diff
-     eResult = eARPGetCacheEntry( &ulDestinationIPAddress, &xMACAddress,
-                                  &( pxNetworkBuffer->pxEndPoint ) );
+     eResult = eARPGetCacheEntry( &ulDestinationIPAddress, &xMACAddress,
+                                  &pxEndPoint );

     if( eResult == eARPCacheHit )
     {
         pvCopySource = &xMACAddress;
+        pxNetworkBuffer->pxEndPoint = pxEndPoint;
     }
     else
     {
         pvCopySource = &pxEthernetHeader->xSourceAddress;
     }

+    if( pxNetworkBuffer->pxEndPoint == NULL )
+    {
+        break;
+    }
~~~

I hadn't had the time to check the code coverage yet.

Test Steps
-----------
Use a similar configuration and try to connect to the remote device.

Checklist:
----------


source/FreeRTOS_TCP_Transmission_IPv4.c
```diff
@@ -298,10 +305,18 @@ void prvTCPReturnPacket_IPV4( FreeRTOS_Socket_t * pxSocket,
...
-                /* Nothing to do: the buffer has been passed to DMA and will be released after use */
+                xDoRelease = pdFALSE;
+                /* The buffer has been passed to DMA and will be released after use */
             }
         } /* if( pxNetworkBuffer != NULL ) */
     } while( ipFALSE_BOOL );
+
+    if( xDoRelease == pdTRUE )
+    {
+        vReleaseNetworkBufferAndDescriptor( pxNetworkBuffer );
```
tony-josi-aws, line 316: As we are releasing the buffer here, the call to `vReleaseNetworkBufferAndDescriptor` [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/FreeRTOS_TCP_Transmission_IPv4.c#L171) doesnt seem to be required.

tony-josi-aws (approved):

source/FreeRTOS_TCP_Transmission_IPv4.c
```diff
@@ -234,19 +230,24 @@ void prvTCPReturnPacket_IPV4( FreeRTOS_Socket_t * pxSocket,
...
             {
                 pvCopySource = &xMACAddress;
+                pxNetworkBuffer->pxEndPoint = pxEndPoint;
             }
             else
             {
                 pvCopySource = &pxEthernetHeader->xSourceAddress;
             }
 
+            if( pxNetworkBuffer->pxEndPoint == NULL )
```
shubnil, line 246: Initially I had some confusion whether eARPGetCacheEntry will return a eARPCacheHit or eARPCacheMiss in this schenario and if the check (pxNetworkBuffer->pxEndPoint == NULL ) needs to be inside the return value check or outside as it is now. However, on some more exploration and internal discussions, it is clear that eARPGetCacheEntry will return eARPCacheMiss and the check can be outside and it will not have any other side-effect. With this I am good with the change.

shubnil (approved):

PR #1081, opened 2024-01-24
CMake Testing Improvements

HTRamsey:
CMake Testing Improvements

Description
-----------
There is a lot more to be done, but here is a start.
- Add missing network interfaces
- Remove unused "config files" that are no longer actually config files
- Fix some file names in unit testing
- Standardize required cmake version with kernel
- Applied a sort on some lists that revealed missing entries, alternatively use GLOB instead of manually listing files and risking it getting out of date.
- There is mixed usage of C90 & C99, I believe the kernel removed the requirement for C90. Thoughts on this?

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


.gitignore
```diff
@@ -17,3 +17,5 @@ __pycache__/
 **/*.d
 **/*.o
 **/*.su
+
+CMakeLists.txt.user
```
Skptak, line 21: Is this needed still? Or was this something you had added for local testing?

source/CMakeLists.txt
```diff
@@ -1,11 +1,9 @@
 add_library( freertos_plus_tcp STATIC )
 
-set_property(TARGET freertos_plus_tcp PROPERTY C_STANDARD 90)
+set_property(TARGET freertos_plus_tcp PROPERTY C_STANDARD 99)
```
Skptak, line 3: @tony-josi-aws @amazonKamath
Do we make any claims that FreeRTOS+TCP is C90 Compliant? If so can we add this claim to the [README.md](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP#freertos-plus-tcp-library) on this repo, and the [FreeRTOS+TCP](https://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/index.html) landing page?

If we don't make any claims about C90 compliance I see no reason not to accept this change to C99 here.

Skptak (commented):
Hey @HTRamsey, thanks for submitting these changes! I'm a fan of the change to the filenames that you've done, and thank you for adding the missing interfaces!

The team is currently in the middle of getting ready for the V4.1.0 release, where we are currently in a "PR merge pause" while finalizing it. Because of this it might be a bit until this gets fully merged in, just to give you a heads up :)

source/CMakeLists.txt
HTRamsey: I wasn't sure about this either, however there are a few things that aren't C90 compliant like unnamed unions and commas at the end of enums.

.gitignore
HTRamsey: Not really needed, just useful when you're running tests. I can remove it.

source/CMakeLists.txt
amazonKamath: @HTRamsey, Kernel still remains at C90.  There was debate sometime last year, but we decided to remain at C90.  For the TCP stack, it seems like the Source/* is at C90 and with pedantic flag enabled. We should be able to catch if there is some non-compliant code, however it looks like that is not the case for Portable/*.   My colleague @tony-josi-aws  is investigating and we will get back to you soon.  Thank you.

amazonKamath: @HTRamsey  FYI, We identified code which has creeped in over the last year(s) which are not C90 compliant.  However, we are trying to figure out what is preventing the build system from throwing the error even though the pedantic flag is set and the standard points to C90 in Cmake file.   Did you notice compile warnings/error for unnamed unions? 
Thanks!

HTRamsey: @amazonKamath yeah they're from struct xNetworkEndPoint

amazonKamath: @HTRamsey  did the compiler complain about it for you?

tony-josi-aws: @HTRamsey, the CI build checks were failing to report C90 incompatible code in the TCP library header files due to a [CMake configuration (SYSTEM keyword)](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/826e261a50b2296d8c79587a3927884fc401dcd4/source/CMakeLists.txt#L93) that was set. As we have identified what is causing the build checks to not enforce the C90 standard, we are finding the best way to remove the incompatibility without any breaking changes.

HTRamsey: [tcp_test.txt](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/14122577/tcp_test.txt)

HTRamsey:
This will pass after #1090

tony-josi-aws (dismissed):

.gitignore
moninom1: Can we remove this if not needed?

shubnil (approved):

tony-josi-aws (approved):

moninom1 (approved):

shubnil (approved):

PR #1086, opened 2024-02-01
Fix for the incoming IPv4 destination IP check

evpopov:
Description
-----------
When `ipconfigETHERNET_DRIVER_FILTERS_PACKETS` is disabled, the stack needs to filter out packets based on their destination IP. In reality it doesn't.

Test Steps
-----------
1. ```#define ipconfigETHERNET_DRIVER_FILTERS_PACKETS (0)```
2. ```#define ipconfigETHERNET_DRIVER_FILTERS_FRAME_TYPES (1)``` Note below.
3. Modify your network driver to allow all packets ( promiscuous mode ).
4. Inject an IPv4 packet with any unicast source and unicast destination addresses. Such packets can an will occur in the wild even with modern switches.
5. Note that `prvAllowIPPacketIPv4()` returns `eProcessBuffer` for the injected packet and processing continues as if the packet was destined for the DUT.
6. Once the change from this PR is applied, the DUT correctly drops the packets for which it does not have an end-point.

Random Thoughts
----------
I don't know where that NULL check came from but as of right now, any received packet will have a non-NULL `pxNetworkBuffer->pxEndPoint` as it propagates through the stack.

Note on: I understand that in promiscuous mode, `ipconfigETHERNET_DRIVER_FILTERS_FRAME_TYPES` should probably be disabled, however the default is ipconfigENABLED and the IP layer should be capable of filtering out packets regardless of whether the lower layers did their job or not.

Let me know what you think


paulbartell:
@evpopov Thanks for the PR. I'll get this over to one of our FreeRTOS+TCP experts for review.

source/FreeRTOS_IPv4.c
```diff
@@ -314,7 +314,6 @@ enum eFrameProcessingResult prvAllowIPPacketIPv4( const struct xIP_PACKET * cons
         }
         else if(
             ( FreeRTOS_FindEndPointOnIP_IPv4( ulDestinationIPAddress, 4 ) == NULL ) &&
-            ( pxNetworkBuffer->pxEndPoint == NULL ) &&
             /* Is it an IPv4 broadcast address x.x.x.255 ? */
```
tony-josi-aws, line 317: The loopback packets might also get dropped here.

shubnil: As Tony pointed out this will drop the loopback packets as well. However, it also makes sense not to have the ( pxNetworkBuffer->pxEndPoint == NULL ) check at line 317. So, for me a good solution would be to change "xBadIPv4Loopback" at line 309 to "xIPv4Loopback" and then inside that have a check for "xBadIPv4Loopback", this way the loopback packets are not dropped as well.

else if( xIPv4Loopback( &( pxIPPacket->xIPHeader ) ) == pdTRUE )
{
    if( xBadIPv4Loopback( &( pxIPPacket->xIPHeader ) ) == pdTRUE )
    {
        /* The local loopback addresses must never appear outside a host. See RFC 1122
             * section 3.2.1.3. */
            eReturn = eReleaseBuffer;
    }
}

htibosch: @tony-josi-aws wrote:
> The loopback packets might also get dropped here.

That is a good observation

I agree with @shubnil. When I added the function `xBadIPv4Loopback()`, I didn't like the word "bad" because it is a negation. So yes I agree with his suggestion.
Let's call it `xAllowIPv4Loopback()`

evpopov: @tony-josi-aws Thanks for catching that. I completely forgot about loopback packets.
After considering you comment, I was going to substitute
```
( pxNetworkBuffer->pxEndPoint == NULL ) &&
```
with 
```
( xIsIPv4Loopback( ulDestinationIPAddress ) == pdFALSE ) &&
```
Let's call this "Option 1" It should work because the `if( xBadIPv4Loopback( &( pxIPPacket->xIPHeader ) ) == pdTRUE )` above check already took care of the situations where only the source OR the destination was a loopback address, so we can get away with just checking the destination.

@shubnil I like your idea because it keeps the checks a little better organized by keeping all loopback-related checks in one spot.
If I understand you correctly, you are suggesting that I create a new function:
```
BaseType_t xIPv4Loopback( const IPHeader_t * const pxIPHeader )
{
    if( xIsIPv4Loopback( pxIPHeader->ulDestinationIPAddress ) ||
        xIsIPv4Loopback( pxIPHeader->ulSourceIPAddress ) )
    {
        /* Either the source or the destination address is an IPv4 loopback address. */
        xReturn = pdTRUE;
    }

    return xReturn;
}
```
Let's call this "Option 2" It is better organized, but requires a new function and feels a little more wasteful because of the added function calls.

Instead of adding a function, I could just rewrite the else if() clause as such:
```
else if( xIsIPv4Loopback( &( pxIPPacket->xIPHeader.ulDestinationIPAddress ) ) == pdTRUE ||
               xIsIPv4Loopback( &( pxIPPacket->xIPHeader.ulSourceIPAddress ) ) == pdTRUE )
{
	/* source OR destination is a loopback address. Make sure the BOTH are */
	if( xBadIPv4Loopback( &( pxIPPacket->xIPHeader ) ) == pdTRUE )
	{
		/* The local loopback addresses must never appear outside a host. See RFC 1122
		* section 3.2.1.3. */
		eReturn = eReleaseBuffer;
	}
}
```
Let's call this "Option 3" It is a little harder to read, but there is no new function and all loopback tests are in one spot. Also, for most non-loopback packets the compiler can probably skip ahead after just one check.

Guys, I'm actually perfectly fine with all 3 options and I don't have much of a preference for any one of them. Let me know which one you want me to implement.

@htibosch 
If I rename `xBadIPv4Loopback()` to `xAllowIPv4Loopback()` I must also invert its returns. The long if()-elseif() clause in `prvAllowIPPacketIPv4()` gets messed up.
I would have to change the elseif() to check for pdFALSE, and also make `xAllowIPv4Loopback()` return pdTRUE for good loopback packets **as well as** all non-loopback packets and I don't like that last part.


p.s. I will do some actual test with loopback packets next week. Thank you ALL for the feedback!

shubnil: Thanks @evpopov  for sharing the various.
Option 3 looks the most clean and crisp solution to me.

Thanks,
Shub

evpopov:
@shubnil  and @tony-josi-aws 
I'm closing this PR because it modifies a public API function's parameters.
I will wait for #1086 to get merged because it deals with the same part of the source and then I will open a new PR that achieves the same goal as this one here, but without modifying public API functions.

htibosch (approved):
We already talked about this upcoming change and I agree with it. Thanks!

tony-josi-aws (approved):

shubnil (approved):

Issue #1088, opened 2024-02-02
[BUG] ARP sends packets over the wrong network interfaces.


evpopov:
**Describe the bug**
When multiple interfaces and endpoints are present, ARP leaks out one NetIf's IPv4 addresses over the other NetIfs.

**Target**
- ALL

**To Reproduce**
![ARP_Leakage](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/assets/25212327/5ef305b0-5b62-4b0e-841d-5eb65a7ca75e)
Check out "before.pcapng" in the attached zip file The DUT send out Gratuitous ARP's which in the capture look like ARP requests. the protocol addresses for these ARP requests are 10.238.238.10 and 192.168.21.47. Those last two addresses belong to the Fake0 interface and should not finding their way to different interfaces. 

**Expected behavior**
After a quick fix, "after.pcapng" from the .zip file shows the resulting ARP packets. I'm pretty sure this should be the correct behavior. where we only see ARP packets associated with 172.30.30.26

**Wireshark logs**
[ARP_Leakage.zip](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/14140394/ARP_Leakage.zip)

I have a fix for this and was planning on starting a PR. Before I do that though, I'd like to have a discussion on the issue. Here are my thoughts and questions:

1. Do you agree that this is a bug. I strongly believe it is. I have never seen a host do this and I can't come up with even a theoretical reason for this behavior... but I have been wrong many times before, so please let me know if you think I'm missing something.
2. The offending code is `FreeRTOS_OutputARPRequest()` which sends out an ARP request on all available IPv4 endpoints.  My first quick and dirty fix was to use `FreeRTOS_FindEndPointOnNetMask()` inside `FreeRTOS_OutputARPRequest()` and only send to the found end-point if there was one. This approach removes a for(;;) loop and one if() clause from `FreeRTOS_FindEndPointOnNetMask()` making it substantially cleaner, but I think there might be an even better fix...
[...]

PR #1092, opened 2024-02-05
Build separation for ARP/ND

HTRamsey:
Improve Build separation for IPv4/IPv6

Description
-----------
Only include FreeRTOS_ARP if IPv4 is enabled. Move everything else into routing which calls either ARP or ND. Also create a frame agnostic helper for checking the cache.

Test Steps
-----------
Need to finish adjusting a couple tests

Checklist:
----------

Related Issue
-----------


paulbartell:
Thanks for the PR. Looks like there are a few formatting and other CI checks that need to be addressed. I'll make sure to send this to the team for review as well.

HTRamsey:
@tony-josi-aws I noticed some issues with ARP/ND in this and I started to fix them but didn't want to lose focus of the point of this PR. I'll make a separate PR for that.

source/FreeRTOS_ND.c
```diff
@@ -1316,4 +1335,97 @@
...
+        /* coverity[misra_c_2012_rule_11_3_violation] */
+        IPPacket_IPv6_t * const pxIPPacket = ( ( IPPacket_IPv6_t * ) pxNetworkBuffer->pucEthernetBuffer );
+        IPHeader_IPv6_t * const pxIPHeader = &( pxIPPacket->xIPHeader );
+        const uint8_t ucNextHeader = pxIPHeader->ucNextHeader;
+
+        if( ( ucNextHeader == ipPROTOCOL_TCP ) || ( ucNextHeader == ipPROTOCOL_UDP ) )
+        {
+            IPv6_Address_t * const pxIPAddress = &( pxIPHeader->xSourceAddress );
+            const IPv6_Type_t eType = xIPv6_GetIPType( ( const IPv6_Address_t * ) pxIPAddress );
+
```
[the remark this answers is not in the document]
HTRamsey: Yeah I can keep them there, I removed them because they were repeating the exact same info output inside 'xIPv6_GetIPType' and `eNDGetCacheEntry`. Also some of the changes I have reverted for the other PR, but haven't pushed yet.

source/FreeRTOS_Routing.c
```diff
@@ -1545,3 +1551,250 @@ struct xIPv6_Couple
...
+
+                FreeRTOS_printf( ( "Looking up %xip with%s end-point\n",
+                                   ( unsigned ) FreeRTOS_ntohl( ulIPAddress ),
+                                   ( pxNetworkBuffer->pxEndPoint != NULL ) ? "" : "out" ) );
+
+                /* Add an entry to the ARP table with a null hardware address.
+                 * This allows the ARP timer to know that an ARP reply is
+                 * outstanding, and perform retransmissions if necessary. */
+                vARPRefreshCacheEntry( NULL, ulIPAddress, NULL );
+
```
tony-josi-aws, line 1687: Missing `iptracePACKET_DROPPED_TO_GENERATE_ARP`

source/FreeRTOS_Routing.c
```diff
@@ -1545,3 +1551,250 @@ struct xIPv6_Couple
...
+             pxEndPoint != NULL;
+             pxEndPoint = FreeRTOS_NextEndPoint( NULL, pxEndPoint ) )
+        {
+            if( xIPType == ( BaseType_t ) ipTYPE_IPv6 )
+            {
+                #if ipconfigIS_ENABLED( ipconfigUSE_IPv6 )
+                    if( pxEndPoint->bits.bIPv6 != 0U )
+                    {
+                        const IPv6_Type_t eEndpointType = xIPv6_GetIPType( &( pxEndPoint->ipv6_settings.xIPAddress ) );
+                        if( eEndpointType == xIPAddrType )
```
tony-josi-aws, line 1624: Why do we need endpoints that perfectly match the address type of destination or source address? For loopback addresses, this means that we need an endpoint with loopback address as its IP.

source/FreeRTOS_UDP_IPv4.c
```diff
@@ -279,7 +262,11 @@ void vProcessGeneratedUDPPacket_IPv4( NetworkBufferDescriptor_t * const pxNetwor
         }
     }
 
-    if( eReturned != eCantSendPacket )
+    if( xLostBuffer == pdTRUE )
```
tony-josi-aws, line 265: In the case of `ipIPv4_FRAME_TYPE`, `xLostBuffer` doesn't seem to be set

source/FreeRTOS_Routing.c
HTRamsey: This was intended to replace functions like pxFindLocalEndpoint, you are correct it's not valid here and is something I have removed (yet to push) to finish later to keep this PR simpler.

source/FreeRTOS_UDP_IPv4.c
HTRamsey: Currently `xStartAddressResolution` works differently for ND and ARP because `vNDSendNeighbourSolicitation` passes control of the packets (can't use FreeRTOS_OutputARPRequest because that works differently and creates new buffers). I was going to make an internal vOutputARPRequest that works like the ND so that `xStartAddressResolution` work similarly for both but I will save that for later.

HTRamsey:
Also parts of `xCheckRequiresResolution` and `xStartAddressResolution` are currently redundant so I need to reconcile duplicate functionality as well. I'll save any functionality changes for a different PR so that this mostly just helps build separation.

source/FreeRTOS_UDP_IP.c
```diff
@@ -94,33 +79,23 @@ UDPPacketHeader_t xDefaultPartUDPPacketHeader =
  */
 void vProcessGeneratedUDPPacket( NetworkBufferDescriptor_t * const pxNetworkBuffer )
 {
-    const UDPPacket_t * pxUDPPacket;
-
-    if( pxNetworkBuffer != NULL )
+    switch( usFrameType( pxNetworkBuffer ) )
```
htibosch, line 82: Also a great improvement using `usFrameType()`.

htibosch (commented):
[Holden](https://github.com/HTRamsey), thank you for this build separation. It is also a cleanup. I was aware that FreeRTOS_ARP.c still had "temporary" code that should be moved to another module.
As I already [commented](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/1088#issuecomment-1931231344), I do find it important that user can also start an ARP search for a device.
Although I have little time, I will run two demos that use IPv4/6, using your PR #1092.

source/FreeRTOS_Routing.c
```diff
@@ -1545,3 +1555,239 @@ struct xIPv6_Couple
...
+ *         The function is to be called within the function xNetworkInterfaceOutput().
+ *
+ * @param[in] pxDescriptor The network buffer which is to be checked for loop-back.
+ * @param[in] bReleaseAfterSend pdTRUE: Driver is allowed to transfer ownership of descriptor.
+ *                              pdFALSE: Driver is not allowed to take ownership of descriptor,
+ *                                       make a copy of it.
+ *
+ * @return pdTRUE/pdFALSE: There is/isn't a loopback address in the packet.
+ */
+BaseType_t xCheckLoopback( NetworkBufferDescriptor_t * const pxDescriptor,
```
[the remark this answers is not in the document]
HTRamsey: Should that interface be included unconditionally? What happens if `ipconfigETHERNET_DRIVER_FILTERS_PACKETS` is disabled and you have a loopback? In that case `xBadIPv6Loopback` never gets checked.

source/FreeRTOS_UDP_IP.c
```diff
@@ -94,33 +79,23 @@ UDPPacketHeader_t xDefaultPartUDPPacketHeader =
...
+        #if ( ipconfigUSE_IPv6 != 0 )
+            case ipIPv6_FRAME_TYPE:
+                vProcessGeneratedUDPPacket_IPv6( pxNetworkBuffer );
                 break;
-        }
+        #endif
+        default:
+            /* Must Know Frame Type first before sending here */
+            FreeRTOS_debug_printf( ( "vProcessGeneratedUDPPacket: Unsupported Packet\n" ) );
+            configASSERT( pdFALSE );
```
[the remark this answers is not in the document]
HTRamsey: I think I was expecting all invalid packets to have been ignored already somewhere in `prvProcessEthernetPacket`, `prvProcessIPPacket`, `prvAllowIPPacketIPv6`, or `prvAllowIPPacketIPv4`, in which case it would never reach here.

source/FreeRTOS_Routing.c
tony-josi-aws: @htibosch  Yes, it has now been added to the latest changes.

HTRamsey:
@tony-josi-aws I went back one more time to make it simpler and less to review

tony-josi-aws (approved):

source/FreeRTOS_UDP_IPv6.c
```diff
@@ -423,7 +423,7 @@ BaseType_t xProcessReceivedUDPPacket_IPv6( NetworkBufferDescriptor_t * pxNetwork
 
         if( pxSocket != NULL )
         {
-            if( xCheckRequiresARPResolution( pxNetworkBuffer ) == pdTRUE )
+            if( xCheckRequiresNDResolution( pxNetworkBuffer ) == pdTRUE )
             {
                 /* Mark this packet as waiting for ARP resolution. */
                 *pxIsWaitingForARPResolution = pdTRUE;
```
shubnil, line 429: Now probably it makes more sense to change the variable name as well to something like "pxIsWaitingForNDResolution",

source/FreeRTOS_IP.c
```diff
@@ -1947,7 +1947,7 @@ static eFrameProcessingResult_t prvProcessIPPacket( const IPPacket_t * pxIPPacke
              * entry. */
             if( ucProtocol != ( uint8_t ) ipPROTOCOL_UDP )
             {
-                if( xCheckRequiresARPResolution( pxNetworkBuffer ) == pdTRUE )
+                if( xCheckRequiresResolution( pxNetworkBuffer ) == pdTRUE )
                 {
                     eReturn = eWaitingARPResolution;
```
shubnil, line 1952: The enum "eWaitingARPResolution" also should be renamed to "eWaitingResolution" to be in synch with the rest of the changes.

shubnil (commented):
This is a good addition. Thanks for creating the same. I have some minor comments, otherwise I am fine with the change.

PR #1094, opened 2024-02-05
Adds a NetworkInterface_t * parameter to FreeRTOS_FindEndPointOnIP_IPv4()

evpopov:
Description
-----------
When an incoming IPv4 frame is evaluated for reception [HERE](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/c91e982b5d597a84bf957cf36f0ec53fb4a355ca/source/FreeRTOS_IPv4.c#L316), the destination IP address of the frame is compared to all endpoints that the DUT has. If a match is found, parsing continues. Otherwise, the frame is dropped.
The issue that this commit fixes is the following:
FindEndPointOnIPv4_IfaceParam
If the above happens, the packet is received. The reason is that FreeRTOS_FindEndPointOnIP_IPv4() never considers the interface on which the search is supposed to be conducted on.
I believe such packets should be dropped.

I considered that the reception of such frames may have been by design... Maybe FreeRTOS+TCP was trying to route between interfaces and allow devices on one interface to reach it's other interfaces. I posed this question on the forum [HERE](https://forums.freertos.org/t/freertos-tcp-routing-from-interface-to-interface/19112)
If receiving such frames is by design, just let me know and I'll close this PR.

I'm starting this PR as a draft for a few reasons:
1 - I need your opinion of you think this is a bug or if it was intended behavior.
2 - I need to further look into [this call](https://github.com/evpopov/FreeRTOS-Plus-TCP/blob/dfea36c26bbeba8cd54acf712b8eb8ab11940600/source/FreeRTOS_ARP.c#L247) because I'm quite sure the output interface is well known as well and can be passed to `FreeRTOS_FindEndPointOnIP_IPv4()`
3 - I need to check if FreeRTOS_FindEndPointOnIP_IPv6() needs the same parameter.

btw, I made sure that if FreeRTOS_FindEndPointOnIP_IPv4() is called with a NULL parameter for the interface, it behaves as before and searches all network interfaces. This ensures that all cases that actually required the search to happen on all interfaces, get their expected result.

### DIFF
--- a/source/FreeRTOS_ARP.c
+++ b/source/FreeRTOS_ARP.c
@@ -244,7 +244,7 @@ _static ARPCacheRow_t xARPCache[ ipconfigARP_CACHE_ENTRIES ];
                 /* Process received ARP frame to see if there is a clash. */
                 #if ( ipconfigARP_USE_CLASH_DETECTION != 0 )
                 {
-                    NetworkEndPoint_t * pxSourceEndPoint = FreeRTOS_FindEndPointOnIP_IPv4( ulSenderProtocolAddress, 2 );
+                    NetworkEndPoint_t * pxSourceEndPoint = FreeRTOS_FindEndPointOnIP_IPv4( NULL, ulSenderProtocolAddress, 2 );
 
                     if( ( pxSourceEndPoint != NULL ) && ( pxSourceEndPoint->ipv4_settings.ulIPAddress == ulSenderProtocolAddress ) )
                     {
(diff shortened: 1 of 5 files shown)


paulbartell:
@evpopov Thanks for the PR. I'll make sure to keep get this over to a team member for review once you mark it ready.

evpopov:
@shubnil and @tony-josi-aws
I'm closing this PR because it modifies a public API function's parameters.
I will wait for https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1086 to get merged because it deals with the same part of the source and then I will open a new PR that achieves the same goal as this one here, but without modifying public API functions.

Issue #1093, opened 2024-02-05
[BUG]prvProcessICMPMessage_IPv6() sends replies to the wrong Endpoint


evpopov:
**Describe the bug**
As as already discussed in #1064, I found the same type of issue when handling ICMPv6 ND packets.

**Target**
- ALL

**To Reproduce**
In my test setup I have two network interfaces. One has an IPv6 Endpoint and is connected to my PC. The second interface is actually a fake interface that pretends to be UP and is not connected to anything. At this stage, My PC has no problem pinging the first interface by it's link-local IPv6 address.
The issue arises when I configure an IPv6 link-local Endpoint **on the second interface**. As soon as I do that, I'm no longer able to ping the first interface from my PC.
What happens is, the PC sends a Neighbor Solicitation packet to the first interface, FreeRTOS+TCP receives it and sends the response out the second interface.
The problematic piece of code is: [HERE](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/6310e61e9430cb2d50b1c128c9c00b06ba96ecd4/source/FreeRTOS_ND.c#L1060)
As was the case in #1064, we are searching for an endpoint when we already have a known good one.
Changing
```
NetworkEndPoint_t * pxEndPointFound = FreeRTOS_FindEndPointOnIP_IPv6( &( pxICMPHeader_IPv6->xIPv6Address ) );
```
to
```
NetworkEndPoint_t * pxEndPointFound = pxNetworkBuffer->pxEndPoint;
```
fixes the issue but I haven't fully examined the problem and the fix. This was just a quick 30min. debugging session and sadly I have to move on.

rant-mode-on....
[...]

PR #1096, opened 2024-02-07
Neighbor Solicitation Response Fix

evpopov:
Description
-----------
IPv6 Neighbor Solicitation Responses could go out the wrong interface. Fixed by using `pxNetworkBuffer->pxEndPoint` instead of searching for and end-point.

I allowed myself the liberty of cleaning up a bit and also enforcing that pxNetworkBuffer, pxNetworkBuffer->pxInterface, and pxNetworkBuffer->pxEndPoint are non-NULL for ALL received packets.

@shubnil , @htibosch, @HTRamsey,  Please let me know what you think

Related Issue
-----------
#1093 


HTRamsey:
Assuming all buffers, interfaces, and endpoints downstream from `ipCONSIDER_FRAME_FOR_PROCESSING` massively simplifies the cyclomatic complexity of everything else and makes it more determinant, which I am a big fan of, but I don't know how the others will feel. I do believe there is a lot of redundant/wasteful checking that makes test coverage more difficult though.

source/FreeRTOS_IP.c
```diff
@@ -1544,6 +1544,9 @@ static void prvProcessEthernetPacket( NetworkBufferDescriptor_t * const pxNetwor
     eFrameProcessingResult_t eReturned = eReleaseBuffer;
 
     configASSERT( pxNetworkBuffer != NULL );
+    configASSERT( pxNetworkBuffer->pxInterface != NULL );
+    configASSERT( pxNetworkBuffer->pxEndPoint != NULL );
```
tony-josi-aws, line 1548: If packet processing relies on these non-NULL checks from here, wouldn't it be better to have them as an if check rather than `configASSERT`, since most release configurations don't define `configASSERT`.

HTRamsey: `configASSERT( pxNetworkBuffer->pxInterface != NULL );` and `configASSERT( pxNetworkBuffer->pxEndPoint != NULL );` should go just below `ipCONSIDER_FRAME_FOR_PROCESSING`. 
`configASSERT( pxNetworkBuffer != NULL );` would be fine if you add an if( pxNetworkBuffer != NULL )` in `prvHandleEthernetPacket`.

tony-josi-aws: > `configASSERT( pxNetworkBuffer->pxInterface != NULL );` and `configASSERT( pxNetworkBuffer->pxEndPoint != NULL );` should go just below `ipCONSIDER_FRAME_FOR_PROCESSING`.
If there isn't a valid interface/endpoint associated with an incoming packet, I can't find any reason to do an `ipCONSIDER_FRAME_FOR_PROCESSING` check since most of the checks in `eConsiderFrameForProcessing` are memory comparisons of the MAC address.

htibosch, line 1548: >     configASSERT( pxNetworkBuffer != NULL );
>     configASSERT( pxNetworkBuffer->pxInterface != NULL );

Some people may wonder why a network buffer has both a reference to an `pxEndPoint` and to an `pxInterface`, because when you know the `pxEndPoint`, `pxInterface` is also known.
The answer is that the reverse is not true: when a packet is received on interface `pxInterface`, we still have to find out which endpoint is handling the incoming packets.

About the added `configASSERT`'s: yes I agree, it is in line with the idea of keeping the initial `pxEndPoint` that was obtained by the network interface.

source/FreeRTOS_ND.c
```diff
@@ -1057,20 +1058,6 @@
                    {
                        size_t uxICMPSize;
                        BaseType_t xCompare;
-                       NetworkEndPoint_t * pxEndPointFound = FreeRTOS_FindEndPointOnIP_IPv6( &( pxICMPHeader_IPv6->xIPv6Address ) );
```
htibosch, line 1060: It is always nice to remove code, isn't it?
This is a good cleanup: the original `pxEndPoint` should be used to reply to the sender.

htibosch (approved):
Hello [Emil](https://github.com/evpopov), thank you for this PR. I approve it.

source/FreeRTOS_IP.c
HTRamsey: `eConsiderFrameForProcessing` does the `FreeRTOS_FindEndPointOnMAC` check which I suppose could be a last effort recovery at determining the endpoint if the NetworkInterface didn't set it correctly. Or I guess you could just require it be set anyways.

HTRamsey: Isn't this a way that @evpopov post [here](https://forums.freertos.org/t/freertos-tcp-routing-from-interface-to-interface/19112/5) would be implemented? Something like `pxNetworkBuffer->pxInterface` and `pxNetworkBuffer->pxInterface->pxEndPoint` are the interfaces and endpoints received from, and `pxNetworkBuffer->pxEndPoint` and `pxNetworkBuffer->pxEndPoint->pxNetworkInterface` are the target interfaces and endpoints?

htibosch: Very often, `configASSERT()` is called just to wake up the developer, so she/he notices that there is a fatal problem.
I always deliver my software with a working version of `configASSERT()`. Up until know I never had an assert.
One alternative is to return `eReleaseBuffer` to the caller and `FreeRTOS_debug_printf` the error.

evpopov: @tony-josi-aws I personally prefer what you are suggesting: do all the NULL checks and cleanly drop the buffer. My reasoning is that if I didn't catch something in testing, it shouldn't break the product in production. To be honest, the only reason I'm adding those asserts is because I'm mimicking what the code already does and I'm trying to follow the original intent. In my personal code, I never do asserts. I always check early, and handle the situation as cleanly as possible and certainly not by going in a forever loop.

Let me try this: Let me substitute the asserts with clean dropping of packets and reexamine where those checks need to go and you can tell me what you think.

tony-josi-aws: Since we are removing the `pxEndPoint != NULL` checks in later part of the code, I believe its safer to do the NULL check and drop the packets here.

evpopov:
Please check out my latest commit and tell me if it's a step in the right or wrong direction.

source/FreeRTOS_IP.c
```diff
@@ -667,38 +667,50 @@ void vIPNetworkUpCalls( struct xNetworkEndPoint * pxEndPoint )
...
+                /* Make it NULL to avoid using it later on. */
+                pxBuffer->pxNextBuffer = NULL;
+            }
+            #endif /* ipconfigIS_ENABLED( ipconfigUSE_LINKED_RX_MESSAGES ) */
 
             prvProcessEthernetPacket( pxBuffer );
-            pxBuffer = pxNextBuffer;
+
+            #if ( ipconfigIS_ENABLED( ipconfigUSE_LINKED_RX_MESSAGES ) )
+                pxBuffer = pxNextBuffer;
```
tony-josi-aws, line 708: `pxNextBuffer` is undeclared here.

source/FreeRTOS_IP.c
```diff
@@ -667,38 +667,50 @@ void vIPNetworkUpCalls( struct xNetworkEndPoint * pxEndPoint )
...
+     * member.  The loop below walks through the chain processing each packet
+     * in the chain in turn. */
+
+    /* The while() below is needed when ipconfigUSE_LINKED_RX_MESSAGES is enabled.
+     * The while() below acts like if() when ipconfigUSE_LINKED_RX_MESSAGES is disabled. */
+    while( pxBuffer != NULL )
+    {
+        /* The network interface should not be sending us network buffers
+         * that are not associated with valid interface and end-point. */
+        if( ( pxBuffer->pxInterface == NULL ) || ( pxBuffer->pxEndPoint == NULL ) )
```
tony-josi-aws, line 686: How about we do these checks in the `prvProcessEthernetPacket` where the `configASSERT`s were added initially.

source/FreeRTOS_IP.c
```diff
@@ -667,38 +667,48 @@ void vIPNetworkUpCalls( struct xNetworkEndPoint * pxEndPoint )
...
+        {
+            /* configASSERT( pdFALSE ); _EP_: Should be leave one of these here for developers that prefer asserts? */
+            vReleaseNetworkBufferAndDescriptor( pxBuffer );
+            break;
```
tony-josi-aws, line 690: When `ipconfigUSE_LINKED_RX_MESSAGES` is enabled, and if one of the linked RX network buffers doesn't have an interface or endpoint, the loop breaks here and the rest of the RX buffers are not processed or released, leading to buffers getting lost.

I believe `prvProcessEthernetPacket` is a better target to add these checks.

evpopov:
I squashed all the changes and force-pushed this to make it easier to look at.
@tony-josi-aws, I'm sorry but I'm out of time. If you think anything else needs to change, please feel free to make the changes.

moninom1 (approved):

shubnil (approved):

PR #1097, opened 2024-02-07
Remove macros ipLOCAL_MAC_ADDRESS and ipLOCAL_IP_ADDRESS_POINTER

HTRamsey:
Remove Old Incompatible Code

Description
-----------
Looks like these are left over from before multiple endpoints were added, and it no longer works in any useful manner.
It also helps resolve some of the linkage issues I kept having in tests due to dependencies causing multiple definitions of `xDefaultPartUDPPacketHeader`. The UDP files shouldn't be required by everything else anymore.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


test/unit-test/FreeRTOS_ARP/FreeRTOS_ARP_utest.c
```diff
@@ -592,7 +584,7 @@ void test_eARPProcessPacket_Request_SenderMACSameAsLocalMAC( void )
 
     FreeRTOS_FindEndPointOnNetMask_IgnoreAndReturn( 1 );
     eResult = eARPProcessPacket( pxNetworkBuffer );
-    TEST_ASSERT_EQUAL( eReturnEthernetFrame, eResult );
+    TEST_ASSERT_EQUAL( eReleaseBuffer, eResult );
```
HTRamsey, line 587: This test was passing incorrectly, as [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/6aa48a566e2259488179397b77c00837fa2eca07/source/FreeRTOS_ARP.c#L298) you should only get eReturnEthernetFrame if the MACs are different. And this test case was intending for them to be the same.

source/portable/NetworkInterface/STM32Fxx/NetworkInterface.c
```diff
@@ -991,19 +991,6 @@ static BaseType_t xMayAcceptPacket( uint8_t * pucEthernetBuffer )
 
         ulDestinationIPAddress = pxIPHeader->ulDestinationIPAddress;
 
-        /* Is the packet for this node? */
-        if( ( ulDestinationIPAddress != *ipLOCAL_IP_ADDRESS_POINTER ) &&
```
HTRamsey, line 995: This check was invalid with multiple endpoints.

HTRamsey:
@tony-josi-aws I assume that one failure had nothing to do with the code right?

tony-josi-aws:
> @tony-josi-aws I assume that one failure had nothing to do with the code right?

@HTRamsey, no, its just that the [GCC website](https://gcc.gnu.org/install/) link was unreachable when the check was done. Triggered a re-run.

source/FreeRTOS_DHCP.c
```diff
@@ -505,10 +505,6 @@
             EP_IPv4_SETTINGS.ulBroadcastAddress = EP_DHCPData.ulOfferedIPAddress | ~( EP_IPv4_SETTINGS.ulNetMask );
             EP_DHCPData.eDHCPState = eLeasedAddress;
 
-            /* _HT_ This macro must be removed later.
-             * It is enough to set 'EP_IPv4_SETTINGS.ulIPAddress'. */
-            *ipLOCAL_IP_ADDRESS_POINTER = EP_IPv4_SETTINGS.ulIPAddress;
```
moninom1, line 510: Agree with the clean up of ipLOCAL_IP_ADDRESS_POINTER.

moninom1 (dismissed):

moninom1:
Seems like there are a lot of UT conflicts, @HTRamsey will you please help in taking a look. Thank you

htibosch (approved):
Thank you for this clean-up. I agree with all of it, except for the PR title.

A friendlier title of this PR would have been:

    "Clean up last remains from the IPv4/single library"

Or:

    "remove macros ipLOCAL_MAC_ADDRESS and ipLOCAL_IP_ADDRESS_POINTER"

moninom1 (approved):

tony-josi-aws (approved):

shubnil:
If the definition inside the function is to restrict the scope, then I am good as long as Misra does not give any warnings.

source/FreeRTOS_TCP_State_Handling.c
```diff
@@ -990,8 +990,7 @@
         #endif /* ipconfigSUPPORT_SELECT_FUNCTION */
 
         /* And bind it to the same local port as its parent. */
-        /*TODO xAddress.sin_address.ulIP_IPv4 = *ipLOCAL_IP_ADDRESS_POINTER; */
-        xAddress.sin_port = FreeRTOS_htons( pxSocket->usLocalPort );
+        FreeRTOS_GetLocalAddress( pxSocket, &xAddress );
```
tony-josi-aws, line 993: Please make it MISRA compliant by making use of the return value. 

` The return value of a non-void function "FreeRTOS_GetLocalAddress" is unused.`

source/FreeRTOS_UDP_IPv4.c
```diff
@@ -117,6 +114,24 @@ void vProcessGeneratedUDPPacket_IPv4( NetworkBufferDescriptor_t * const pxNetwor
...
+                0x00,                               /* ucDifferentiatedServicesCode. */
+                0x00, 0x00,                         /* usLength. */
+                0x00, 0x00,                         /* usIdentification. */
+                0x00, 0x00,                         /* usFragmentOffset. */
+                ipconfigUDP_TIME_TO_LIVE,           /* ucTimeToLive */
+                ipPROTOCOL_UDP,                     /* ucProtocol. */
+                0x00, 0x00,                         /* usHeaderChecksum. */
+                0x00, 0x00, 0x00, 0x00              /* Source IP address. */
+            };
+
```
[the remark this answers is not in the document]
HTRamsey: I did it to limit scope and to be consistent with the same done [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/b72e2b6eabfaf39b9d9efad7d26167a9ca582caa/source/FreeRTOS_ARP.c#L1433). Although I didn't make it static. I'm not even sure it's worth it to keep it when there's only four non-zero bytes at initialization.

moninom1 (approved):

tony-josi-aws (approved):

shubnil (approved):

PR #1103, opened 2024-02-21
eARPGetCacheEntryGateWay() return value with no valid gateway

evpopov:
Description
-----------
When an IPv4 packet needs to be sent to a host outside the local network, we need to send this packet to the gateway.
If there is no valid gateway, we cannot send the packet.

Test Steps
-----------
1. Configure an single IPv4 endpoint with and IP of 192.168.0.1/24 and a gateway of 0.0.0.0
2. Open a TCP connection to 10.10.10.10
3. `eARPGetCacheEntryGateWay()` should result in `eCantSendPacket`


kstribrnAmzn:
The [test_eARPGetCacheEntry_GatewayAddressZero](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/test/unit-test/FreeRTOS_ARP/FreeRTOS_ARP_utest.c#L2215-L2243) unit test is failing. It's due to the test expecting a cache miss stemming from [this final lookup](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/FreeRTOS_ARP.c#L1089-L1103).

Updating [this line](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/test/unit-test/FreeRTOS_ARP/FreeRTOS_ARP_utest.c#L2241) to `TEST_ASSERT_EQUAL_MESSAGE( eCantSendPacket, eResult, "Test 9" );` will fix the failing unit tests

evpopov:
Thanks @kstribrnAmzn

kstribrnAmzn (dismissed):

kstribrnAmzn:
Tagging @shubnil and @tony-josi-aws for review

moninom1 (approved):

moninom1:
Testing in progress, will merge after that.

htibosch (approved):
@evpopov, thank you for this PR!

I tested the PR in real hardware.

Before applying this change and in case there is no IPv4 gateway defined, `eARPGetCacheEntryGateWay()` returns `eARPCacheMiss`.

After applying the change and when there is no IPv4 gateway, `eARPGetCacheEntryGateWay()` returns `eCantSendPacket`, which is correct.

That is an important difference, because when `eCantSendPacket` is returned, the TCP stack will give up sending immediately.
`eARPCacheMiss` will trigger an ARP lookup, and unnecessary waiting.

Please merge it :-)

tony-josi-aws (approved):

PR #1104, opened 2024-02-21
Prevents FreeRTOS_OutputARPRequest() sending on all interfaces and endpoints

### LINKED ISSUE #1088: [BUG] ARP sends packets over the wrong network interfaces.

evpopov:
This PR is my proposal for how we should fix #1088
I decided to keep things simple and stay away from modifying API function parameters, or adding new macros/functions.

Let me know what you think.

### DIFF
--- a/source/FreeRTOS_ARP.c
+++ b/source/FreeRTOS_ARP.c
@@ -1274,13 +1274,19 @@ static BaseType_t prvFindCacheEntry( const MACAddress_t * pxMACAddress,
         NetworkBufferDescriptor_t * pxNetworkBuffer;
         NetworkEndPoint_t * pxEndPoint;
 
-        /* Send an ARP request to every end-point which has the type IPv4,
-         * and which already has an IP-address assigned. */
-        for( pxEndPoint = FreeRTOS_FirstEndPoint( NULL );
-             pxEndPoint != NULL;
-             pxEndPoint = FreeRTOS_NextEndPoint( NULL, pxEndPoint ) )
+        pxEndPoint = FreeRTOS_FindEndPointOnNetMask( ulIPAddress, 3 );
+
+        /* Only send an ARP request if we were able to find an IPv4 end-point
+         * whose IP/mask combination matches ulIPAddress */
+        if( pxEndPoint != NULL )
         {
-            if( ( pxEndPoint->bits.bIPv6 == pdFALSE_UNSIGNED ) &&
+            /* We found a good end-point, but FreeRTOS_FindEndPointOnNetMask() never checked if we gave it
+             * an IP address of 0.0.0.0, so theoretically, we can fall through here if the user calls us with
+             * and IP of 0 and the end-point doesn't have a valid IP.
+             * FreeRTOS_FindEndPointOnNetMask() will also return a match for 255.255.255.255, but we should
+             * not output an ARP request for it. Make sure we catch those cases. */
+            if( ( ulIPAddress != 0U ) &&
+                ( ulIPAddress != ~0U ) &&
                 ( pxEndPoint->ipv4_settings.ulIPAddress != 0U ) )
             {
                 /* This is called from the context of the IP event task, so a block time


tony-josi-aws:
@evpopov, thanks for creating this PR.

We have created another PR (#1105) also targeting issue #1088. Would love to have your thoughts on the same, as it takes a slightly different approach by creating a new API, `FreeRTOS_OutputARPRequest_Multi` which requires the interface to be targeted while sending the ARP messages while keeping the `FreeRTOS_OutputARPRequest` and making it use the new API internally.

`FreeRTOS_OutputARPRequest_Multi` takes the target interface as an argument so that finding the interface is deferred to the caller of the API, while `FreeRTOS_OutputARPRequest` is used when the required interface is not readily available from the caller's context and would require a call to `FreeRTOS_FindEndPointOnNetMask`. This could save redundant search for endpoints.

evpopov:
Thanks. I commented in #1105.
This PR is no longer needed.

tony-josi-aws:
Thanks @evpopov

> I commented in #1105.
I couldn't see any comments from you there.

PR #1105, opened 2024-02-22
Fixes ARP messages from being sent through wrong endpoints

### LINKED ISSUE #1088: [BUG] ARP sends packets over the wrong network interfaces.

tony-josi-aws:
Description
-----------
This PR fixes #1088, by adding a new API `FreeRTOS_OutputARPRequest_Multi` which requires the interface to be targeted while sending the ARP messages. 

This PR also fixes some variable declarations that are present after code.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

#1088 


source/FreeRTOS_ARP.c
```diff
@@ -1338,6 +1340,24 @@ static BaseType_t prvFindCacheEntry( const MACAddress_t * pxMACAddress,
...
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Create and send an ARP request packet.
+ *
+ * @param[in] ulIPAddress A 32-bit representation of the IP-address whose
+ *                         physical (MAC) address is required.
+ */
+    void FreeRTOS_OutputARPRequest( uint32_t ulIPAddress )
```
moninom1, line 1352: It should enclosed with ipconfigIPv4_BACKWARD_COMPATIBLE flag.

source/FreeRTOS_DHCP.c
```diff
@@ -1455,7 +1447,7 @@
                  * it easier to lookup a device in a router's list of DHCP clients. */
 
                 /* Point to where the OPTION_END was stored to add data. */
-                pucPtr = &( pucUDPPayloadBuffer[ dhcpFIRST_OPTION_BYTE_OFFSET + ( *pxOptionsArraySize - 1U ) ] );
+                uint8_t * pucPtr = &( pucUDPPayloadBuffer[ dhcpFIRST_OPTION_BYTE_OFFSET + ( *pxOptionsArraySize - 1U ) ] );
```
moninom1, line 1450: uint8_t * pucPtr should be declared at line 1400, any reason for moving it here?

source/FreeRTOS_TCP_IP_IPv6.c
```diff
@@ -100,6 +100,15 @@ BaseType_t xProcessReceivedTCPPacket_IPV6( NetworkBufferDescriptor_t * pxDescrip
 {
     /* Function might modify the parameter. */
     NetworkBufferDescriptor_t * pxNetworkBuffer = pxDescriptor;
+    const ProtocolHeaders_t * pxProtocolHeaders;
```
moninom1, line 103: Good cleanup, was it coming as a warning or MISRA error?

source/FreeRTOS_ARP.c
tony-josi-aws: Hey Monika, `FreeRTOS_OutputARPRequest` is used in the stack when there isn't an endpoint readily available and have to use `FreeRTOS_FindEndPointOnNetMask` (basically what `FreeRTOS_OutputARPRequest` does). Hence, it is not a function kept for backward compatibility.

source/FreeRTOS_DHCP.c
tony-josi-aws: `pucPtr` is only used when `#if ( ipconfigDHCP_REGISTER_HOSTNAME == 1 )` evaluates to true.

source/FreeRTOS_TCP_IP_IPv6.c
tony-josi-aws: TY, its part of the C90 compliance changes not MISRA.

source/FreeRTOS_ARP.c
moninom1: Make sense.

moninom1 (dismissed):

ActoryOu (commented):
Could you help add description on why some `FreeRTOS_OutputARPRequest` are changed but some are still same?

test/cbmc/proofs/IP/ProcessEthernetPacket/ProcessEthernetPacket_harness.c
```diff
@@ -87,5 +89,8 @@ void harness()
     pxNetworkBuffer->pucEthernetBuffer = ( ( ( uint8_t * ) safeMalloc( pxNetworkBuffer->xDataLength ) ) );
     __CPROVER_assume( pxNetworkBuffer->pucEthernetBuffer != NULL );
 
+    /* Its assumed that every RX packet will have a valid network interface assigned. (Asserted in code) */
+    pxNetworkBuffer->pxInterface = &xInterface;
```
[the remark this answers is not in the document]
tony-josi-aws: In #1096 we have [made explicit checks](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1096/files#diff-8b61e00e54b55092bc30ba18e0dc9570da3218e0b816bd15d47ba3b81c054318R1556) to make sure that the interface or endpoint of an RX packet is never NULL.

source/FreeRTOS_ARP.c
```diff
@@ -1338,6 +1340,24 @@ static BaseType_t prvFindCacheEntry( const MACAddress_t * pxMACAddress,
...
+    {
+        const NetworkEndPoint_t * pxEndPoint = FreeRTOS_FindEndPointOnNetMask( ulIPAddress, 12 );
```
shubnil, line 1354: Please add a comment mentioning the below assumptions :
1. IPv4 endpoints belonging to same physical interface can have same subnet
2. IPv4 endpoints belonging to different physical interface in the system will have a different subnet.

shubnil (commented):
The code looks good. We can add the assumptions for subnet in both the code as well in the endpoint documentation.

ActoryOu (dismissed):

moninom1 (dismissed):

moninom1 (approved):

moninom1 (dismissed):

shubnil (dismissed):

htibosch:
The addition of the function `FreeRTOS_OutputARPRequest_Multi()` is very useful.

However, at the moment the function is called, we already have a reliable endpoint pointer. So I think there is no need to iterate through all endpoints.

In your PR, I see the following loop:

~~~c
void FreeRTOS_OutputARPRequest_Multi( const NetworkInterface_t * pxNetworkInterface,
                                      uint32_t ulIPAddress )
{
    /* Iterate through all endpoints that are attached to 'pxNetworkInterface'. */
    for( pxEndPoint = FreeRTOS_FirstEndPoint( pxNetworkInterface );
         pxEndPoint != NULL;
         pxEndPoint = FreeRTOS_NextEndPoint( pxNetworkInterface, pxEndPoint ) )
    {
        if( ( pxEndPoint->bits.bIPv6 == pdFALSE_UNSIGNED ) &&
            ( pxEndPoint->ipv4_settings.ulIPAddress != 0U ) )
        {
            // As for the endpoint, you are using the first IPv4 endpoint which has an IP-address
            break;
        }
    }
}
~~~

I am using the following configuration for a test:
~~~c
    SetupV4_t setup_v4_first =
    {
        .pcMACAddress = "00-11-11-11-11-07",
        .pcIPAddress = "172.16.0.104",
        .pcMask = "255.255.224.0",
        .eType = eStatic,  /* or eDHCP if you wish. */
    };
    SetupV4_t setup_v4_second =
    {
        .pcMACAddress = "00-11-11-11-11-02",
        .pcIPAddress = "192.168.2.104",
        .pcMask = "255.255.255.0",
        .pcGateway = "192.168.2.1",
        .pcDNS[0] = "118.98.44.10",
        .eType = eStatic,
    };
    xSetupEndpoint_v4( &xInterfaces[ 0 ], &( xEndPoints[ 0 ] ), &setup_v4_first );
    xSetupEndpoint_v4( &xInterfaces[ 0 ], &( xEndPoints[ 1 ] ), &setup_v4_second );
~~~
( using [net_setup.[ch]](https://github.com/htibosch/freertos_plus_projects/blob/master/plus/Framework/FreeRTOS-Plus-TCP-main/tools/tcp_utilities/net_setup.c) to configure the endpoints. )

When querying gateway "192.168.2.1" with ARP,  queries are still sent using wrong endpoints. The above function selects the first IPv4 endpoint, which is on the wrong network "172.x.x.x", in stead of "192.168.2.x".
When I swap the order of the declarations, the problem does not occur.

Your test-case: when I use two separate interfaces, the problem does not occur neither. But when I attach the 2 endpoints to the same interface, ARP does not work properly.

So my proposal is the following:
~~~c
void FreeRTOS_OutputARPRequest_Multi( const NetworkEndPoint_t * pxEndpoint,
                                      uint32_t ulIPAddress )
{
    /* This is called from the context of the IP event task, so a block time
     * must not be used. */
    pxNetworkBuffer = pxGetNetworkBufferWithDescriptor( sizeof( ARPPacket_t ), ( TickType_t ) 0U );
    // etc
}
~~~
You will have to change the calls in 4 locations, which is straightforward.

After the above change, the order of declaration of the endpoints doesn't matter anymore.

Regards

evpopov:
Hein,
I think you are referring to #1105
<https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1105> and not my PR. My
PR was #1104 <https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1104>
My PR did not add a new function, It just fixed the
old FreeRTOS_OutputARPRequest() function.
After Tony opened #1105, I closed my PR.
I suggested to Tony that he include all the changes from my PR because
after the call to FreeRTOS_FindEndPointOnNetMask() we still need to make
sure that ulIPAddress is != 0 and != ~0 before we send an ARP request. They
haven't included tose checks in #1105 yet, but I hope they do eventually.

htibosch:
@evpopov wrote:
> Hein, I think you are referring to #1105 and not my PR. My PR was #1104

Yes indeed, I was testing Tony's PR about `FreeRTOS_OutputARPRequest()`.
My remarks also apply to your #1104.

tony-josi-aws:
Hi @htibosch,

>         if( ( pxEndPoint->bits.bIPv6 == pdFALSE_UNSIGNED ) &&
>            ( pxEndPoint->ipv4_settings.ulIPAddress != 0U ) )
>        {
>           // As for the endpoint, you are using the first IPv4 endpoint which has an IP-address
>            break;
>       }

The `FreeRTOS_OutputARPRequest_Multi` API doesn't have a `break`, the requests are sent through all the endpoints of the given interface.

Though the ARP request to the first endpoint is not necessary in this case, it will be sent through the second endpoint as well.

htibosch:
Hello @tony-josi-aws,
> The `FreeRTOS_OutputARPRequest_Multi` API doesn't have a `break`, the requests are sent through all the endpoints of the given interface.
> 
> Though the ARP request to the first endpoint is not necessary in this case, it will be sent through the second endpoint as well.

I haven't had the time to check this thought:

**Without** the break, the ARP packet will be sent to each IPv4 endpoint that it up, but with a different return address. My router on 192.168.2.1 will reply to it, but the there is a problem: the ARP cache entry already has an entry for 192.168.2.1, but with a different endpoint.

**With** the break, there is 50% chance that it works for me, that depends on the order in which de endpoints were added.

source/FreeRTOS_ARP.c
```diff
@@ -1264,77 +1264,71 @@ static BaseType_t prvFindCacheEntry( const MACAddress_t * pxMACAddress,
...
  *                         physical (MAC) address is required.
  */
-    void FreeRTOS_OutputARPRequest_Multi( const NetworkInterface_t * pxNetworkInterface,
+    void FreeRTOS_OutputARPRequest_Multi( const NetworkEndPoint_t * pxEndPoint,
                                           uint32_t ulIPAddress )
     {
         NetworkBufferDescriptor_t * pxNetworkBuffer;
-        NetworkEndPoint_t * pxEndPoint;
 
         /* Send an ARP request to every end-point which has the type IPv4,
```
moninom1, line 1278: Please remove the comment.

htibosch (approved):
Thank you Tony so far, I approve the changes

test/cbmc/proofs/ARP/ARPProcessPacket/ARPProcessPacket_harness.c
```diff
@@ -8,12 +8,14 @@
...
 #include "FreeRTOS_IP_Private.h"
 #include "FreeRTOS_ARP.h"
+#include "FreeRTOS_Routing.h"
 
 /* This pointer is maintained by the IP-task. Defined in FreeRTOS_IP.c */
 extern NetworkBufferDescriptor_t * pxARPWaitingNetworkBuffer;
 
 /* This is an output function and need not be tested with this proof. */
-void FreeRTOS_OutputARPRequest( uint32_t ulIPAddress )
+void FreeRTOS_OutputARPRequest_Multi( NetworkEndPoint_t * pxEndPoint,
```
moninom1, line 17: As we are removing FreeRTOS_OutputARPRequest, don't we have the case where flow is from FreeRTOS_OutputARPRequest to -> FreeRTOS_OutputARPRequest_Multi?

moninom1 (approved):

shubnil (approved):

PR #1106, opened 2024-02-23
Fix declarations after statements

tony-josi-aws:
Description
-----------
This PR updates the codebase to remove declarations after statements that are non compliant as per the C90 standard.

Test Steps
-----------

Checklist:
----------

- ~[ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.~

Related Issue
-----------

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,9 +217,6 @@ add_compile_options(
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-extra-semi-stmt>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-missing-noreturn>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-cast-align>
-
-    # Suppressions required to build clean with GCC.
-    $<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-declaration-after-statement>
 )
 
 ########################################################################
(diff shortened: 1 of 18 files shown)


CMakeLists.txt
```diff
@@ -217,9 +217,6 @@ add_compile_options(
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-extra-semi-stmt>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-missing-noreturn>
     $<$<COMPILE_LANG_AND_ID:C,Clang>:-Wno-cast-align>
-
-    # Suppressions required to build clean with GCC.
-    $<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-declaration-after-statement>
```
moninom1, line 222: If we are removing this have you checked build with gcc

tony-josi-aws: Yes, the C90 compliance checks has been done on GCC and Clang

shubnil (commented):
The change looks good.
But, I have 2 queries.
1. Is declaration at the top enough for C90 or should we initialize to some default values as well
2. Are we just moving the declaration up and initialization is still at its place

tony-josi-aws:
1. Declaration itself is enough
2. If we are able to know the data before hand like constants we are initializing it during declaration itself, but for the cases were it depends on other variables/function calls we defer the initialization until other variables are known

shubnil (approved):

kar-rahul-aws (approved):

PR #1084, opened 2024-01-30
Improve Validation of Config

HTRamsey:
Improve Validation of Config

Description
-----------
Verifies FreeRTOSConfig.h values that are required
Uses static asserts on a few ipconfig values that can't be compile time checked normally

Test Steps
-----------

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1117, opened 2024-03-14
Add Kernel Version Check

HTRamsey:
Adds a check for minimum supported kernel version


source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -148,6 +148,10 @@
 
 /*---------------------------------------------------------------------------*/
 
+#if ( ( tskKERNEL_VERSION_MAJOR < 8 ) || ( ( tskKERNEL_VERSION_MAJOR == 8 ) && ( tskKERNEL_VERSION_MINOR < 1 ) ) )
```
AniruddhaKanhere, line 151: I suppose that this specific version is used due to the fact that before this `pdMS_TO_TICKS` did not exist which is used a lot in this library.

AniruddhaKanhere (approved):
The changes look good to me.

Thank you @HTRamsey for taking the time to contribute!

Thanks

source/include/FreeRTOSIPConfigDefaults.h
HTRamsey: It's actually from [this](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1084#issuecomment-1997748807) but looking at the changelog myself I think maybe it should be v9.0 cause that's when task notifications were added. I can try and at least see the oldest version that will compile later.

AniruddhaKanhere: Thank you! That would be great.

AniruddhaKanhere:
@ActoryOu should we be including the kernel headers here https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/df5aed9e58a72e2489b6c1936788885672c62981/test/build-combination/Header_Self_Contain/headerSelfContain.cmake#L21 before compile options as the build check is failing there.

source/include/FreeRTOSIPConfigDefaults.h
tony-josi-aws: @HTRamsey 

Task notifications are used by the IP stack only if `ipconfigSELECT_USES_NOTIFY` is enabled. Network interfaces might still use it, depending on how the EMAC task is implemented.

ActoryOu:
> @ActoryOu should we be including the kernel headers here
> 
> https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/df5aed9e58a72e2489b6c1936788885672c62981/test/build-combination/Header_Self_Contain/headerSelfContain.cmake#L21
> 
> before compile options as the build check is failing there.

Sorry for late response.
Yeah, that helps. And @HTRamsey created a better solution by #1120.
Let's do it that way.

HTRamsey:
@tony-josi-aws to your first point, the value would be in making it obvious to users why it is not compiling, instead of getting more questions on the forum. To the second point, the check is currently for v8.1, not v8.2.

PR #1118, opened 2024-03-15
DNS local lookups using mDNS repaired

htibosch:
Description
-----------
Somewhere along the way, active lookups using mDNS got broken.

When a peer answers an mDNS a lookup, in its answer, it will **not** repeat the questions that were asked by the client.

Test Steps
-----------

I did 3 kinds of lookups, each of them in 4 different ways:
A normal DNS lookup
A mDNS lookup of `laptop-hp.local`
An LLMNR lookup of `laptop-hp`

I repeated each lookup by either asking for A or for AAAA records. Also I would either use IPv4 or IPv6 to contact the peer/server, setting [`xDNS_IP_Preference`](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/4364e78a4a6413d11012cc3994a5f5c07df4af78/source/include/FreeRTOS_DNS.h#L91).

Here are the results of of mDNS/LLMNR tests:

[DNS_tests.zip](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/files/14611445/DNS_tests.zip)

Without the reparation, mDNS would not succeed.

Checklist:
----------

Related Issue
-----------


source/FreeRTOS_DNS.c
```diff
@@ -1248,17 +1248,6 @@
         /* Make sure all fields of the 'sockaddr' are cleared. */
         ( void ) memset( ( void * ) &xAddress, 0, sizeof( xAddress ) );
 
-        #if ( ipconfigUSE_IPv6 != 0 )
```
moninom1, line 1251: Is this assignment causing any issue?

htibosch: You are referring to this code:

~~~c
#if ( ipconfigUSE_IPv6 != 0 )
    if( xFamily == ( BaseType_t ) FREERTOS_AF_INET6 )
    {
        xDNS_IP_Preference = xPreferenceIPv6;
    }
    else
    {
        xDNS_IP_Preference = xPreferenceIPv4;
    }
#endif /* ( ipconfigUSE_IPv6 != 0 ) */
~~~

The global variable `xDNS_IP_Preference` was introduced to indicate the preference for a DNS server: either IPv4 or IPv6.

When I test the DNS functions, I always try out 4 combinations: contact the server with either IPv4 or IPv6, and ask for either an A (IPv4) or an AAAA (IPv6) record.

What we could do it hide the variable `xDNS_IP_Preference` and add an access function.

In answer to your question: changing `xDNS_IP_Preference` disturbs the testing process. The variable should be under control of the DUT. At least, that is how it was designed.

tony-josi-aws:
@htibosch 

> Without the reparation, mDNS would not succeed.

I tested the mDNS without the changes using ping (from a Windows PC):
```
C:\Users\tonyjosi>ping STM32F4.local

Pinging STM32F4.local [192.168.0.102] with 32 bytes of data:
Reply from 192.168.0.102: bytes=32 time=1ms TTL=64
Reply from 192.168.0.102: bytes=32 time=1ms TTL=64
Reply from 192.168.0.102: bytes=32 time=1ms TTL=64
Reply from 192.168.0.102: bytes=32 time=1ms TTL=64

Ping statistics for 192.168.0.102:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 1ms, Maximum = 1ms, Average = 1ms
```
and also with a DNS v4 query from another +TCP DUT. It seems to be working.

Wondering if the tests you performed were different.

htibosch:
> 
> and also with a DNS v4 query from another +TCP DUT. It seems to be working.
> 
> Wondering if the tests you performed were different.

My tests were about a LLMNR or mDNS lookup lookup that originates from a DUT running FreeRTOS+TCP.

The biggest problem is this code in FreeRTOS_DNS_Parser.c:

~~~
if( xSet.usQuestions == 0U )
{
    /* The IP-stack will only accept DNS replies that have a copy
     * of the questions. */
    xReturn = pdFALSE;
    break;
}
~~~
As for mDNS it is normal that the questions are **not** repeated.

tony-josi-aws:
@htibosch 

Can you help fix the CI checks?

shubnil (approved):

test/unit-test/FreeRTOS_DNS/FreeRTOS_DNS_utest.c
```diff
@@ -1040,6 +1040,8 @@ void test_FreeRTOS_getaddrinfo_a_IPv6Random_BindFailWithDot( void )
     struct xSOCKET xDNSSocket;
     NetworkEndPoint_t xEndPoint[ 5 ];
 
+    xDNS_IP_Preference = xPreferenceIPv6;
```
moninom1, line 1043: We have to manually set this because xDNS_IP_Preference is always xPreferenceIPv4?

shubnil (approved):

kar-rahul-aws (approved):

moninom1:
FreeRTOS_SetDNSIPPreference makes more sense, as now application can set it accordingly. We might have to update this API in the website as well. Thanks for the update.

moninom1 (approved):

PR #1120, opened 2024-03-20
Fix header includes

HTRamsey:
Hopefully will allow #1117 to pass
Ensures correct ordering of includes for everything.
Might be a bit pedantic, but having the core/optional include sections promotes the idea of add-on features rather than having everything intertwined. Similar to how [cyclone](https://github.com/Oryx-Embedded/CycloneTCP) is organized which I'm a big fan of.
Also, I removed the ipconfig checks for the headers because in my opinion it's cleaner to just guard the entire implementation with the ipconfig checks rather than the headers, which will have the same effect during builds but with less clutter. I'd suggest leaving them if the source files were organized in a module manner rather than a single root folder.

### DIFF
--- a/source/include/FreeRTOS_DNS_Cache.h
+++ b/source/include/FreeRTOS_DNS_Cache.h
@@ -28,13 +28,11 @@
 #ifndef FREERTOS_DNS_CACHE_H
 #define FREERTOS_DNS_CACHE_H
 
-/* FreeRTOS includes. */
-#include "FreeRTOS.h"
+/* Global Includes & Definitions. */
+#include "FreeRTOS_IP_Common.h"
 
+/* Optional FreeRTOS+TCP Includes. */
 #include "FreeRTOS_DNS_Globals.h"
-
-/* Standard includes. */
-#include <stdint.h>
 
 #if ( ( ipconfigUSE_DNS_CACHE == 1 ) && ( ipconfigUSE_DNS != 0 ) )
 
--- a/source/include/FreeRTOS_DNS_Callback.h
+++ b/source/include/FreeRTOS_DNS_Callback.h
@@ -29,16 +29,14 @@
 #ifndef FREERTOS_DNS_CALLBACK_H
 #define FREERTOS_DNS_CALLBACK_H
 
-/* FreeRTOS includes. */
-#include "FreeRTOS.h"
+/* Global Includes & Definitions. */
+#include "FreeRTOS_IP_Common.h"
 
-/* FreeRTOS+TCP includes. */
+/* Core FreeRTOS+TCP Includes. */
 #include "FreeRTOS_IP.h"
 
+/* Optional FreeRTOS+TCP Includes. */
 #include "FreeRTOS_DNS_Globals.h"
-
-/* Standard includes. */
-#include <stdint.h>
 
 /* *INDENT-OFF* */
 #ifdef __cplusplus
--- a/source/include/FreeRTOS_IPv6_Private.h
+++ b/source/include/FreeRTOS_IPv6_Private.h
@@ -28,20 +28,16 @@
 #ifndef FREERTOS_IPV6_PRIVATE_H
 #define FREERTOS_IPV6_PRIVATE_H
 
-/* Application level configuration options. */
-#include "FreeRTOSIPConfig.h"
-#include "FreeRTOSIPConfigDefaults.h"
+/* Global Includes & Definitions */
 #include "FreeRTOS_IP_Common.h"
+
+/* Core FreeRTOS+TCP Includes. */
 #include "FreeRTOS_Sockets.h"
 #include "FreeRTOS_Stream_Buffer.h"
-#if ( ipconfigUSE_TCP == 1 )
-    #include "FreeRTOS_TCP_WIN.h"
-    #include "FreeRTOS_TCP_IP.h"
-#endif
 
-#include "semphr.h"
-
-#include "event_groups.h"
+/* Optional FreeRTOS+TCP Includes. */
+#include "FreeRTOS_TCP_WIN.h"
+#include "FreeRTOS_TCP_IP.h"
 
 /* *INDENT-OFF* */
 #ifdef __cplusplus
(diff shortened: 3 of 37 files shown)


source/include/FreeRTOS_IPv6_Private.h
```diff
@@ -28,20 +28,16 @@
...
-    #include "FreeRTOS_TCP_WIN.h"
-    #include "FreeRTOS_TCP_IP.h"
-#endif
-
-#include "semphr.h"
 
-#include "event_groups.h"
+/* Optional FreeRTOS+TCP Includes. */
+#include "FreeRTOS_TCP_WIN.h"
+#include "FreeRTOS_TCP_IP.h"
```
ActoryOu, line 40:
```suggestion
#if ( ipconfigUSE_TCP == 1 )
    #include "FreeRTOS_TCP_WIN.h"
    #include "FreeRTOS_TCP_IP.h"
#endif
```
We should align the behavior on this header.

source/include/FreeRTOS_Routing.h
```diff
@@ -26,16 +26,15 @@
...
-    #endif
+/* Core FreeRTOS+TCP Includes. */
+    #include "FreeRTOS_IP.h"
 
-    #if ( ipconfigUSE_IPv6 != 0 )
-        #include "FreeRTOS_DHCPv6.h"
-    #endif
+/* Optional FreeRTOS+TCP Includes. */
+    #include "FreeRTOS_DHCP.h"
+    #include "FreeRTOS_DHCPv6.h"
```
ActoryOu, line 37:
```suggestion
    #if ( ipconfigUSE_DHCP != 0 )
        #include "FreeRTOS_DHCP.h"
    #endif

    #if ( ipconfigUSE_IPv6 != 0 )
        #include "FreeRTOS_DHCPv6.h"
    #endif
```
Align behavior. But I think we can change `ipconfigUSE_IPv6` to `ipconfigUSE_DHCPv6` here.

ActoryOu (commented):
Hi @HTRamsey,
Appreciate your contribution! I really like this change!
Left some comments.

BTW, I have tested this with #1117, and it works!

source/include/FreeRTOS_IPv6_Private.h
HTRamsey: Okay I can do that if you prefer, I had mentioned in my original post that the reason I got rid of them was I figured it was cleaner to just guard the source implementation with the config macros rather than try to catch every include everywhere. You aren't wasting flash if the function declarations are never defined.

ActoryOu: Could you show me the original post? I'd like to align the information first.

ActoryOu (dismissed):

shubnil (approved):

ActoryOu (approved):

shubnil (approved):

tony-josi-aws (approved):

source/include/FreeRTOS_DNS_Cache.h
```diff
@@ -28,14 +28,12 @@
 #ifndef FREERTOS_DNS_CACHE_H
 #define FREERTOS_DNS_CACHE_H
 
-/* FreeRTOS includes. */
-#include "FreeRTOS.h"
+/* Global Includes & Definitions. */
+#include "FreeRTOS_IP_Common.h"
```
n9wxu, line 32: adding FreeRTOS_IP_Common to bring in FreeRTOS.h and stdint.h also adds dependencies upon many other modules.  This is a bad idea and encourages spaghetti code.

source/include/FreeRTOS_DNS_Callback.h
```diff
@@ -29,17 +29,15 @@
 #ifndef FREERTOS_DNS_CALLBACK_H
 #define FREERTOS_DNS_CALLBACK_H
 
-/* FreeRTOS includes. */
-#include "FreeRTOS.h"
+/* Global Includes & Definitions. */
+#include "FreeRTOS_IP_Common.h"
```
n9wxu, line 33: adding FreeRTOS_IP_Common to bring in FreeRTOS.h also adds dependencies upon many other modules.  This is a bad idea and encourages spaghetti code.

n9wxu:
FreeRTOS_IP_Common.h is a bad idea.  There is really nothing common about that header file and adding additional headers just creates dependencies where there should not be any.

I am reverting this PR.  The common header should go away entirely with the other contents moved into new headers appropriate to the purpose.  The OP compared with CycloneTCP as a a good example.  I agree, the structure of CycloneTCP should be followed here with many small headers specific to the needed functionality.

HTRamsey:
I agree that each header should only include what it needs, which the vast majority of them will at least need
```
#include "FreeRTOS.h"
#include "task.h"

#include "FreeRTOSIPConfig.h"
#include "FreeRTOSIPConfigDefaults.h"
```
and most of the typedefs in the headers should be avoided to allow forward declarations. Also all of the source files currently basically include everything. That was all a lot more than I wanted to tackle in one PR to get the unit tests to pass for the other PRs.

cobusve:
There is probably a good case to consolidate the Config and defaults, if config has to ALWAYS be followed by defaults then it can be included in the config header file making the semantics of the dependency clear and resolving the risk of include order, but I agree we cannot make an aggregate header that includes all headers.

We are in the process of trying to restructure and we must make sure every step we take takes us closer to the target not the opposite direction.

At this point if anybody finds a header that uses something declared in a different header please fix that instance by including the header which it clearly depends on instead, that takes us closer to the goal.

We should also remove headers that are being included but are not being used wherever we see this.

HTRamsey:
@cobusve Since the config is a user provided file I guess we would have to just rely on the check [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/9f4ebd532fe7ac145bc229ab02272d4d82f36ab6/source/include/FreeRTOSIPConfigDefaults.h#L44). While I clearly did it incorrectly, the purpose of this PR was to address the issue [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/8415936029/job/23041843293) where the problem was that `FreeRTOS.h` & `task.h` were needed to be included before `FreeRTOSIPConfig.h` & `FreeRTOSIPConfigDefaults.h` for each header, which wasn't the case for FreeRTOS_DNS.h at least. An alternative is to just not do any compile time config checks. Then you wouldn't need `FreeRTOS.h` & `task.h` for the config files.

PR #1121, opened 2024-03-20
Remove ipconfigMULTI_INTERFACE

HTRamsey:
Removes ipconfigMULTI_INTERFACE, which is only referenced in the tcp_utilities and one place in the [Demos](https://github.com/FreeRTOS/FreeRTOS/blob/000c005ceed37605a3c146459acdf605ec90919e/FreeRTOS-Plus/Demo/FreeRTOS_Plus_TCP_IPv6_Demo/common/NTP/NTPDemo.c#L553). If you guys disagree then no problem but it seems to me it's not worth it to further maintain compatibility between the tcp_utilities and an older version of +TCP. If somebody has a single interface version of +TCP then they'll have the old tcp_utilities that work with it as well. I can make a PR to remove that usage from the Demos if this gets approved.

Edit: I also just realized that's the purpose of ipconfigCOMPATIBLE_WITH_SINGLE anyways

### DIFF
--- a/source/FreeRTOS_Routing.c
+++ b/source/FreeRTOS_Routing.c
@@ -170,10 +170,6 @@ struct xIPv6_Couple
             {
                 /* Other interfaces are already defined, so iterate to the end of the
                  * list. */
-
-                /*_RB_ Question - if ipconfigMULTI_INTERFACE is used to define the
-                 * maximum number of interfaces, would it be more efficient to have an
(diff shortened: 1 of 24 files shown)


ActoryOu (approved):
I like this PR as well. Looks good to me.

tony-josi-aws (approved):

shubnil (approved):

PR #1129, opened 2024-04-03
CI Workflow Update

Skptak:
Description
-----------
Update the ci.yml workflow to use echo groups with colour. Use consistent spacing as well.

Test Steps
-----------
View github action page

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 
 env:
   bashPass: \033[32;1mPASSED -
-  bashWarn: \033[33;1mWARNING -
+  bashInfo: \033[33;1mWINFO -
   bashFail: \033[31;1mFAILED -
   bashEnd:  \033[0m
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -21,24 +21,33 @@ jobs:
...
           # ${{ env.stepName }}
-          echo "::group::Install Dependencies"
-          git submodule update --checkout
-          cd tools/CMock
-          git submodule update --init vendor/unity
-          cd ../..
+          echo -e "::group::${{ env.bashInfo }} ${{ env.stepName }} ${{ env.bashEnd }}"
+
+          git submodule update --checkout --init --recursive
+          sudo apt-get update -y
```
Skptak, line 33: This is a functional change I've added to the PR. Supposed to perform a `sudo apt update` before installing dependencies, for more info refer [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners).

kstribrnAmzn (approved):

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 env:
   bashPass: \033[32;1mPASSED -
-  bashWarn: \033[33;1mWARNING -
+  bashInfo: \033[33;1mWINFO -
```
kstribrnAmzn, line 12: Ridiculous Nit-pick: I would prefer 'info' logs to be a different color than yellow (maybe [blue](https://gist.github.com/vratiu/9780109#file-bash_aliases-L16)?). I think 'warn' being yellow makes the most sense. 

Obviously not important enough to block

tony-josi-aws (approved):

.github/workflows/ci.yml
Skptak: That makes sense, it's something we'd probably want to change in all the repos that do this to be consistent though.

Issue #1136, opened 2024-04-15
[BUG] prvTCPReturnPacket_IPV4() dereferences pxNetworkBuffer outside its null pointer check


anordal:
This applies to the latest release (V4.1.0) as well as origin/HEAD as of writing.

Clang-tidy (16 and 18) complains about a couple of possible null pointer dereferences in this repository, and I think it is right about one of them. They are of the form of a null pointer check (hinting that it can be NULL) and a dereference outside the check that dereferences it anyway.

## function prvTCPReturnPacket_IPV4(), variable pxNetworkBuffer:

`source/FreeRTOS_TCP_Transmission_IPv4.c:147:44: warning: Access to field 'pucEthernetBuffer' results in a dereference of a null pointer (loaded from variable 'pxNetworkBuffer'`

This is the one I think clang-tidy is right about. Here, the unguarded dereference is accompanied by null pointer checks both before and after:

```
#if ( ipconfigZERO_COPY_TX_DRIVER != 0 )
{
    if( xDoRelease == pdFALSE )
    {
        /* A zero-copy network driver wants to pass the packet buffer
         * to DMA, so a new buffer must be created. */
        pxNetworkBuffer = pxDuplicateNetworkBufferWithDescriptor( pxNetworkBuffer, ( size_t ) pxNetworkBuffer->xDataLength );

        if( pxNetworkBuffer != NULL )
        {
            xDoRelease = pdTRUE;
        }
        else
        {
            FreeRTOS_debug_printf( ( "prvTCPReturnPacket: duplicate failed\n" ) );
        }
    }
}
#endif /* ipconfigZERO_COPY_TX_DRIVER */

// …

pxIPHeader = ( ( IPHeader_t * ) &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER ] ) );

#ifndef __COVERITY__
    if( pxNetworkBuffer != NULL ) /* LCOV_EXCL_BR_LINE the 2nd branch will never be reached */
#endif
```

Interestingly, the section above this would have made sure it isn't a null pointer if it wasn't for assigning it again here. If the assignment here (in the ipconfigZERO_COPY_TX_DRIVER section) is to uphold the same guarrantee, it is missing some error handling.

## function prvSendDHCPRequest(), variable EP_DHCPData.xDHCPSocket:

This must be a false positive, but I'll include it.
[...]

PR #1137, opened 2024-04-17
Tcp transmission: Fix nullptr dereference

anordal:
Description
-----------
The first commit moves the initialization of pxIPHeader. This fixes a possible nullpointer dereference.

~The second commit reduces the scope of pxIPHeader and other local variables.~ Dropped.

Test Steps
-----------
I am using this CMake hack to compile my project with clang-tidy:

```
# Setting *_CLANG_TIDY tells CMake to insert itself as a compiler wrapper,
# underneath Ccache, that also runs Clang-tidy. Thus, Clang-tidy is rerun
# whenever the compiler is, as determined by Ninja and Ccache. For Ccache
# invalidation to work when settings are changed, the settings must therefore be
# in the arguments, not some ".clang-tidy" file that Ccache doesn't know about.
if(CMAKE_C_COMPILER_ID MATCHES "Clang")
    set(tidy_checks
        -clang-analyzer-security.insecureAPI.*
        -clang-analyzer-deadcode.DeadStores
        # -clang-analyzer-core.NullDereference
    )
    string(JOIN "," tidy_checks_str ${tidy_checks})
    set(CMAKE_C_CLANG_TIDY
        clang-tidy
        --use-color
        --checks=${tidy_checks_str}
        --warnings-as-errors=*,${tidy_checks_str}
    )
endif()
```

Checklist:
----------

- [x] I have run clang-tidy on the changed files to confirm that the changes at least make clang-tidy happy.

Related Issue
-----------
#1136

### DIFF
--- a/source/FreeRTOS_TCP_Transmission_IPv4.c
+++ b/source/FreeRTOS_TCP_Transmission_IPv4.c
@@ -141,16 +141,16 @@ void prvTCPReturnPacket_IPV4( FreeRTOS_Socket_t * pxSocket,
         }
         #endif /* ipconfigZERO_COPY_TX_DRIVER */
 
-        /* MISRA Ref 11.3.1 [Misaligned access] */
-        /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-113 */
-        /* coverity[misra_c_2012_rule_11_3_violation] */
-        pxIPHeader = ( ( IPHeader_t * ) &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER ] ) );
-
         #ifndef __COVERITY__
             if( pxNetworkBuffer != NULL ) /* LCOV_EXCL_BR_LINE the 2nd branch will never be reached */
         #endif
         {
             NetworkInterface_t * pxInterface;
+
+            /* MISRA Ref 11.3.1 [Misaligned access] */
+            /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-113 */
+            /* coverity[misra_c_2012_rule_11_3_violation] */
+            pxIPHeader = ( ( IPHeader_t * ) &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER ] ) );
 
             /* Map the Ethernet buffer onto a TCPPacket_t struct for easy access to the fields. */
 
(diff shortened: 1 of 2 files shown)


source/FreeRTOS_TCP_Transmission_IPv4.c
```diff
@@ -141,17 +129,29 @@ void prvTCPReturnPacket_IPV4( FreeRTOS_Socket_t * pxSocket,
...
-        /* MISRA Ref 11.3.1 [Misaligned access] */
-        /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-113 */
-        /* coverity[misra_c_2012_rule_11_3_violation] */
-        pxIPHeader = ( ( IPHeader_t * ) &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER ] ) );
-
         #ifndef __COVERITY__
             if( pxNetworkBuffer != NULL ) /* LCOV_EXCL_BR_LINE the 2nd branch will never be reached */
         #endif
         {
+            TCPPacket_t * pxTCPPacket = NULL;
```
shubnil, line 136: This "do-while" loop is a single iteration loop. There is not much optimization if the declarations are inside the loop are where they were. Can we only keep the NULL pointer dereference line (line 153) inside the if check and rest of the declarations can be where they were

source/FreeRTOS_TCP_Transmission_IPv6.c
```diff
@@ -147,19 +137,30 @@ void prvTCPReturnPacket_IPV6( FreeRTOS_Socket_t * pxSocket,
...
-        /* MISRA Ref 11.3.1 [Misaligned access] */
-        /* More details at: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/MISRA.md#rule-113 */
-        /* coverity[misra_c_2012_rule_11_3_violation] */
-        pxIPHeader = ( ( IPHeader_IPv6_t * ) &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER ] ) );
-
         #ifndef __COVERITY__
             if( pxNetworkBuffer != NULL ) /* LCOV_EXCL_BR_LINE the 2nd branch will never be reached */
         #endif
         {
+            TCPPacket_IPv6_t * pxTCPPacket = NULL;
```
shubnil, line 144: Same as above

shubnil (commented):
Thanks for creating the change. This is a good catch. I have added a minor comment.

source/FreeRTOS_TCP_Transmission_IPv4.c
shubnil: Thanks @anordal . I was saying more from the point of view of issue to fix mapping. We can have the variable optimization as a different PR. Thanks for the understanding.

shubnil (approved):

tony-josi-aws (approved):

tony-josi-aws:
@anordal , thanks for creating this PR

PR #1140, opened 2024-04-26
Enabling user control of compiler analysis for non ISO statements

thirtytwobits:
Description
-----------
This change replaces the GCC-specific fix for -pedantic in FreeRTOS_Sockets.c with a configurable solution. The change has no functional impact.

Test Steps
-----------
Compiles with GCC, -pedantic, and -Werror

Checklist:
----------
- [x] I have tested by changes but have not run the existing tests.

Related Issue
-----------
(none)

# Conflicts:
#	source/FreeRTOS_Sockets.c

Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/include/FreeRTOSIPConfigDefaults.h
+++ b/source/include/FreeRTOSIPConfigDefaults.h
@@ -3374,6 +3374,43 @@ STATIC_ASSERT( ipconfigDNS_SEND_BLOCK_TIME_TICKS <= portMAX_DELAY );
 
 /*---------------------------------------------------------------------------*/
 
+/*
+ * ipconfigISO_STRICTNESS_VIOLATION_START, ipconfigISO_STRICTNESS_VIOLATION_END
+ *
+ * Type: compiler pragma injection macros
+ *
+ * These two macros enclose parts of the source that contain intentional
+ * deviations from the ISO C standard. Users, and AI (I welcome our robot
+ * overlords!), can use this to customize static analysis settings such as
+ * the `-pedantic` flag in GCC. These should appear in very few places within
+ * the FreeRTOS TCP source and should enclose only a line or two at a time.
+ * When first introduced these macros enclosed a single line of source code in
+ * the sockets implementation.
+ *
+ * GCC example
+ *
+ * In gcc, to allow the Free RTOS TCP code to compile with `-pedantic` you can
+ * define these macros as such:
+ *
+ * ```
+ * // Last tested in GCC 10
+ * #define ipconfigISO_STRICTNESS_VIOLATION_START _Pragma("GCC diagnostic push") \
+ *  _Pragma("GCC diagnostic ignored \"-Wpedantic\"")
+ *
+ * #define ipconfigISO_STRICTNESS_VIOLATION_END _Pragma("GCC diagnostic pop")
+ * ```
+ */
+
+#ifndef ipconfigISO_STRICTNESS_VIOLATION_START
+    #define ipconfigISO_STRICTNESS_VIOLATION_START
+#endif
+
+#ifndef ipconfigISO_STRICTNESS_VIOLATION_END
+    #define ipconfigISO_STRICTNESS_VIOLATION_END
+#endif
+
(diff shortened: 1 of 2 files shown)


source/include/FreeRTOSIPConfigDefaults.h
```diff
@@ -3374,6 +3374,43 @@ STATIC_ASSERT( ipconfigDNS_SEND_BLOCK_TIME_TICKS <= portMAX_DELAY );
...
+ * ipconfigISO_STRICTNESS_VIOLATION_START, ipconfigISO_STRICTNESS_VIOLATION_END
+ *
+ * Type: compiler pragma injection macros
+ *
+ * These two macros enclose parts of the source that contain intentional
+ * deviations from the ISO C standard. Users, and AI (I welcome our robot
+ * overlords!), can use this to customize static analysis settings such as
+ * the `-pedantic` flag in GCC. These should appear in very few places within
+ * the FreeRTOS TCP source and should enclose only a line or two at a time.
+ * When first introduced these macros enclosed a single line of source code in
```
jasonpcarroll, line 3387:
```suggestion
 * When first introduced, these macros enclosed a single line of source code in
```

jasonpcarroll (commented):
Looks good! I will get someone else to also take a look. You will need to update the formatting per the CI check, but that can easily be done by applying the git patch found under Artifacts->formattingChanges on the summary page of the CI run.

thirtytwobits:
Forgot semicolons. Will update sometime today.

thirtytwobits:
memcpy is a pretty heavy hand to get around some macros. For example:

https://godbolt.org/z/WePM9W85f

The macros avoid any trickery in the compiler stage isolating it to the pre-processing stage. My assumption was you'd prefer to write idiomatic and typesafe C and accept some macros as a wart rather than the inverse? Even looking at the difference between my lighter-weight hack and the current code -> https://godbolt.org/z/M1soG7rfh <- you're still generating extra instructions at -O0 (should be identical for any other level of optimization) which does show this is actively thwarting the type system for every compiler whereas the macros have no effect until/unless the user adds directives specific to their compiler.

But I'll do this either way. It's your project. I'm just visiting ;-)

htibosch:
> @thirtytwobits
> 
> Thanks for this PR. As [we discussed](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1135#issuecomment-2058703110) in #1135, what is your thought on using memcpy to avoid the warning instead of adding macros that translate to compiler specific code in the common source code?

I also thought of using `memcpy()`, but a problem will still occur when a function pointer is larger than a void pointer.

htibosch (approved):
The following assignment seems to be a mortal sin:
~~~
    typedef void (* SocketWakeupCallback_t )( Socket_t * pxSocket );
    SocketWakeupCallback_t pxUserWakeCallback;

    pxUserWakeCallback = ( SocketWakeupCallback_t ) pvOptionValue;
~~~

The problem is that we treat a function pointer as if it were a void pointer. The reason for this is that `FreeRTOS_setsockopt()` uses a void parameter:
~~~
    BaseType_t FreeRTOS_setsockopt( Socket_t xSocket,
                                    int32_t lLevel,
                                    int32_t lOptionName,
                                    const void * pvOptionValue,
                                    size_t uxOptionLength );
~~~
The option `FREERTOS_SO_WAKEUP_CALLBACK` will bind a user function to a socket.
On some platforms, the two pointers (void* versus function*) should be treated in a different way.

Thanks for this PR, I approve it.

thirtytwobits:
> The socket option is rarely set, and most `memcpy()`s are optimised to copy 4 or 8 bytes in a few instructions. But as I said, the use of `void *` is the problem.

but you'd add a `bl` statement; you would modify the runtime to work around a compiler front-end diagnostic. This seems like the wrong bias.

tony-josi-aws (approved):

chinglee-iot (approved):

aggarg (approved):

PR #1151, opened 2024-06-01
Let's fix the tests enough to run with AddressSanitizer and UB Sanitizer and enable those in CI

anordal:
This PR adds an option to build with sanitizers, fixes all AddressSanitizer and UB-Sanitizer errors (because these sanitizers can be run together) and enforces them in CI.

Motivation:
The tests were crashing when compiled with Gcc 13. With these fixes, they no longer do.

I think it would be nice for future development to run the tests with sanitizers in CI to prevent the sort of bugs I've now fixed. They were not few, which should contribute to reduce the amount of flakiness (especially crashiness) in the tests.

I have focused on what's necessary to enable these sanitizers, but there is more. Particularly uninitialized variables. The next I would do is ban it with `-Werror=maybe-uninitialized`, but that's going to reveal the rest of that iceberg.

Test Matrix
-----------

* Tests shall now pass both with `cmake -DSANITIZE=address,undefined` and `cmake -DSANITIZE=`.
* Tests shall now pass both for Gcc 11 and 13 (and hopefully 12).
* Code coverage shall be perfect on all metrics (line, function, branch) when compiling with no sanitizers (regardless of compiler).
  Fixing line coverage with sanitizers is still more work, but branch coverage is currently not attainable (presumably due to artificial branches that get inserted but not ignored).

Checklist:
----------
- [x] Code coverage: There are no changes in this PR that don't already have 100% branch coverage by unit-tests.

Related Issue
-----------
None.


tony-josi-aws:
@anordal Thanks for taking time to contribute to FreeRTOS+TCP.
Enabling address and undefined behavior sanitizers in the CI check would help improve the code quality.

source/FreeRTOS_TCP_IP.c
```diff
@@ -594,7 +601,7 @@
             FreeRTOS_debug_printf( ( "Connect[%xip:%u]: next timeout %u: %u ms\n",
                                      ( unsigned ) pxSocket->u.xTCP.xRemoteIP.ulIP_IPv4, pxSocket->u.xTCP.usRemotePort,
                                      pxSocket->u.xTCP.ucRepCount, ( unsigned ) ulDelayMs ) );
-            pxSocket->u.xTCP.usTimeout = ( uint16_t ) ipMS_TO_MIN_TICKS( ulDelayMs );
+            pxSocket->u.xTCP.usTimeout = ( uint16_t ) ulDelayMs;
```
tony-josi-aws, line 604: Wondering why the conversion of time in milliseconds to rtos ticks (`ipMS_TO_MIN_TICKS`) was removed here. `pxSocket->u.xTCP.usTimeout` is expected to have time (in ticks).

anordal: Good catch. I think I should drop this change.

I only tried to remove the clamping of 0 to 1, since I handled that in the millisecond domain, but this clamping is in the tick domain, so it's not the same.

anordal: Cutting the commit down to the essential change,
here is the new commit 7c0e0605dc
and the total diff [84710cce..10c134d8](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/compare/84710cce..10c134d8).

I think it was a bit confusing that the variable is a TickType_t while holding milliseconds.
I would consider moving the call to ipMS_TO_MIN_TICKS to each place it is assigned instead.
I don't know if optimizers are generally smart enough to delete the clamping operation in its current place, but I would expect it on the constant expressions.

tony-josi-aws:
@anordal can you help fix the merge conflict with the `test/unit-test/FreeRTOS_TCP_IP/FreeRTOS_TCP_IP_utest.c`?

anordal:
Yes. I see you fixed a use after free (ba6ba81f). I'll rebase.

shubnil (approved):

tony-josi-aws (approved):

PR #1166, opened 2024-07-02
Allow FetchContent usage in external projects

nicovanduijn:
Description
------------
My project has an existing FreeRTOS Kernel and I attempted to include this project via CMake's `FetchContent` function as suggested by the Readme. In this process I stumbled over two minor details:
- Copy-paste error in the readme, suggesting to set defines for the FAT build instead of the TCP library
- No functionality to prevent the TCP library to pull in its own copy of the kernel, as other libraries have

Test Steps
-----------
- Set up a project with bare-bones FreeRTOS
- Consume this library via CMake's `FetchContent` as suggested by the readme

Checklist:
----------
- [~] I have tested my changes. No regression in existing tests.
  -> I have obviously tested that this solution fixes *my* problem, but I'm unsure how I can test it against regressions in the upstream CI other than just opening this PR.
   -> No real code changes outside the build system, so this does not apply

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.21)
 cmake_policy(SET CMP0048 NEW) # project version
 cmake_policy(SET CMP0076 NEW) # full paths
 
@@ -174,4 +174,6 @@ add_subdirectory(source)
 add_subdirectory(tools)
 add_subdirectory(test)
 
-FetchContent_MakeAvailable(freertos_kernel cmock)
+if(PROJECT_IS_TOP_LEVEL)
+    FetchContent_MakeAvailable(freertos_kernel cmock)
+endif()
--- a/README.md
+++ b/README.md
@@ -60,13 +60,12 @@ FetchContent_Declare( freertos_plus_tcp
   - this particular example supports a native and cross-compiled build option.
 
 ```cmake
-set( FREERTOS_PLUS_FAT_DEV_SUPPORT OFF CACHE BOOL "" FORCE)
 # Select the native compile PORT
-set( FREERTOS_PLUS_FAT_PORT "POSIX" CACHE STRING "" FORCE)
-# Select the cross-compile PORT
+set( FREERTOS_PLUS_TCP_NETWORK_IF "POSIX" CACHE STRING "" FORCE)
+# Or: select a cross-compile PORT
 if (CMAKE_CROSSCOMPILING)
-  # Eg. Zynq 2019_3 version of port
-  set(FREERTOS_PLUS_FAT_PORT "ZYNQ_2019_3" CACHE STRING "" FORCE)
+  # Eg. STM32Hxx version of port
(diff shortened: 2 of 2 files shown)


kstribrnAmzn:
Thank you for submitting this PR! I took a look over it and saw nothing wrong. I'll send this to the TCP experts on the team for their review.

CMakeLists.txt
```diff
@@ -174,4 +174,6 @@ add_subdirectory(source)
 add_subdirectory(tools)
 add_subdirectory(test)
 
-FetchContent_MakeAvailable(freertos_kernel cmock)
+if(FREERTOS_PLUS_TCP_FETCH_FREERTOS)
+    FetchContent_MakeAvailable(freertos_kernel cmock)
+endif()
```
kstribrnAmzn, line 179: CMake documentation (for the latest release) mentions
>True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number (including floating point numbers). False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND.

We should keep default behavior to clone the FreeRTOS kernel. This default behavior is used in [this now failing check](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/9763074732/job/26950437825?pr=1166). Maybe negate your variable to something like `FREERTOS_PLUS_TCP_USE_LOCAL_FREERTOS`. Then your condition could be...

```
if(NOT FREERTOS_PLUS_TCP_USE_LOCAL_FREERTOS)
    FetchContent_MakeAvailable(freertos_kernel cmock)
endif()
```

nicovanduijn: Great input, indeed this breaks existing builds. Thanks for the pointer.

I flipped the logic, hopefully CI passes now.

kstribrnAmzn (approved):

CMakeLists.txt
archigup: Perhaps it makes sense to check if we are the top level CMake project and only clone FreeRTOS kernel then?

nicovanduijn: I didn't know this was possible until I googled it just now. Is doing [this](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) considered best practice?

nicovanduijn: @archigup I've changed the logic as you suggested

archigup (approved):

CMakeLists.txt
```diff
@@ -174,4 +174,6 @@ add_subdirectory(source)
 add_subdirectory(tools)
 add_subdirectory(test)
 
-FetchContent_MakeAvailable(freertos_kernel cmock)
+if(PROJECT_IS_TOP_LEVEL)
```
kstribrnAmzn, line 177: `PROJECT_IS_TOP_LEVEL` is new for CMake 3.21. I think we'll need to bump the minimum required version of CMake at the top of this file.

nicovanduijn: done.

kstribrnAmzn: Thank you!

kstribrnAmzn (approved):

n9wxu (approved):

Issue #1172, opened 2024-07-25
[Feature Request] Xilinx Zynq and Ultrascale functional differences in 4.x


JohnnyOpcode:
I'd like to migrate a MPSoC (Ultrascale) project to 4.x, but 

https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c

is not keeping pace with

https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/Zynq/NetworkInterface.c

htibosch:
Hello, thank you for your interest in FreeRTOS+TCP.

How do you mean that the Ultrascale driver is not keeping pace with the Zynq driver?

ActoryOu:
Hi @htibosch,
I think he means that the network interface of xilinx_ultrascale does not support multiple interface, which is introduced by TCP V4.

Hello @JohnnyOpcode,
We warmly welcome and appreciate contributions from the community to our repository. Your participation and code submissions are highly valued. Feel free to create PR once ready and we'll review it.

Thank you in advance for your support and involvement!

JohnnyOpcode:
> Hi @htibosch, I think he means that the network interface of xilinx_ultrascale does not support multiple interface, which is introduced by TCP V4.
> 
> Hello @JohnnyOpcode, We warmly welcome and appreciate contributions from the community to our repository. Your participation and code submissions are highly valued. Feel free to create PR once ready and we'll review it.
> 
> Thank you in advance for your support and involvement!

@ActoryOu I don't have the level of subject-matter expertise to code such a PR. I have my hands full with just the FreeRTOS robotic application (on MPSoC).

pete-pjb:
Hi @ActoryOu , @JohnnyOpcode and @htibosch ,

I did a port for this a number of months ago, but I have been too busy to contribute it...  Please find attached zip with source, if anyone has time to get it into the source tree please feel free!

[zus_gem.zip](https://github.com/user-attachments/files/16392705/zus_gem.zip)

I hope that is helpful!

Kind Regards,

Pete

htibosch:
Ah yes, now I understand it, Pete.
I will create a PR to upgrade the Ultrascale driver.
Thanks

htibosch:
I looked at the proposed new driver in detail. At the first sight it looks good.
[...]

PR #1187, opened 2024-09-11
Update Zynq Ultrascale port for V4.x and Clean up

StefanBalt:
Description
-----------

This pull request introduces the V4.x port for the Xilinx Ultrascale, which basically merges the previous implementation, the Zynq7000 port and the suggestion by @pete-pjb here #1172

**Key Changes:**
-  V4.x Port: A combination of previous work, the Zynq7000 port, and the proposal in by @pete-pjb here #1172 
- Interrupt Mapping: A new header file, x_emac_map.h, has been introduced to handle interrupt IDs. Since there is no canonical definition for EMAC interrupt IDs, the header maps interrupt IDs based on the EMAC base address.
- LLMNR Update: The LLMNR implementation now leverages the multicast MAC hash table rather than using a dedicated MAC entry.
- Removal of Zynq7000 (GEM V2) Code: Zynq7000-related code has been removed, as the ports for the Zynq7000 and Ultrascale are quite distinct. Keeping the Zynq7000 code IMHO no longer adds value and was redundant.
- Micrel PHY support is also from the .zip provided by @pete-pjb. However, this has not been tested and may require further validation.

I will provide inline comments on specific sections where I'm uncertain or need feedback.

Test Steps
-----------
I could only test a single interface with IPv4 so far. Junbo Frames and LLMNR also work fine.
I plan to test multiple interfaces on a Kria board I have at my hand soon.

Checklist:
----------

Related Issue
-----------
#1172 


source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -123,116 +126,259 @@ static void prvEMACHandlerTask( void * pvParameters );
...
 
+            #if ( ( ipconfigUSE_MDNS == 1 ) && ( ipconfigUSE_IPv6 != 0 ) )
+                XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MacAddress.ucBytes ); /* why for IPv6? */
+                XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MACAddressIPv6.ucBytes );
+
+                /* allow reception of multicast addresses programmed into hash */
+                XEmacPs_SetOptions( pxEMAC_PS, XEMACPS_MULTICAST_OPTION );
+            #endif /* ( ( ipconfigUSE_MDNS == 1 ) && ( ipconfigUSE_IPv6 != 0 ) ) */
+
+            /* In the Zynq7000 port the MAC-Address is stored on second endpoint here at position 4. Needed? */
```
StefanBalt, line 364: As mentioned in the comment Zynq7000 does this here:
```
        pxEndPoint = FreeRTOS_NextEndPoint( pxInterface, pxEndPoint );

        if( pxEndPoint != NULL )
        {
            /* If there is a second end-point, store the MAC
             * address at position 4.*/
            XEmacPs_SetMacAddress( pxEMAC_PS, ( void * ) pxEndPoint->xMACAddress.ucBytes, 4 );
        }
```
Do we also need this here?

source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -123,116 +126,259 @@ static void prvEMACHandlerTask( void * pvParameters );
...
+                }
+                #endif /* if ( ipconfigUSE_IPv6 == 0 ) */
+
+
+                /* allow reception of multicast addresses programmed into hash */
+                XEmacPs_SetOptions( pxEMAC_PS, XEMACPS_MULTICAST_OPTION );
             }
             #endif /* ipconfigUSE_LLMNR == 1 */
 
+            #if ( ( ipconfigUSE_MDNS == 1 ) && ( ipconfigUSE_IPv6 != 0 ) )
```
StefanBalt, line 356: Why only add MDNS mulicast MAC to hash table when `( ipconfigUSE_IPv6 != 0 )`?

source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -348,12 +348,19 @@ static BaseType_t xUltrascaleNetworkInterfaceInitialise( NetworkInterface_t * px
                     XEmacPs_SetHash( pxEMAC_PS, ( void * ) xLLMNR_MacAddressIPv6.ucBytes );
                 }
                 #endif /* if ( ipconfigUSE_IPv6 == 0 ) */
+
+
+                /* allow reception of multicast addresses programmed into hash */
+                XEmacPs_SetOptions( pxEMAC_PS, XEMACPS_MULTICAST_OPTION );
```
StefanBalt, line 352: These are also missing in the Zynq7000 port and most likely needed.
Should I add it? But I cannot test it.

source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -207,32 +318,62 @@ BaseType_t xNetworkInterfaceInitialise( void )
...
-                XEmacPs_SetMacAddress( pxEMAC_PS, ( void * ) xLLMNR_MACAddress, 2 );
+                #if ( ipconfigUSE_IPv6 == 0 )
+                {
+                    /* Also add LLMNR multicast MAC address. */
+                    XEmacPs_SetMacAddress( pxEMAC_PS, ( void * ) xLLMNR_MacAddress.ucBytes, 2 );
+                }
+                #else
+                {
+                    /* is this loop really required? */
+                    for( pxEndPoint = FreeRTOS_FirstEndPoint( pxInterface );
```
StefanBalt, line 332: Is it really necessary to add `ucMACAddress` to all endpoints?

I would have thought `XEmacPs_SetHash( pxEMAC_PS, ( void * ) xLLMNR_MacAddressIPv6.ucBytes );` would be sufficient.

source/portable/NetworkInterface/xilinx_ultrascale/x_emacpsif_hw.c
```diff
@@ -173,7 +181,7 @@ static void emacps_handle_error( void * arg,
                 if( ( ErrorWord & XEMACPS_RXSR_HRESPNOK_MASK ) != 0 )
                 {
                     last_err_msg = "Receive DMA error";
-                    xNetworkInterfaceInitialise();
+                    vInitialiseOnIndex( xEMACIndex );
```
StefanBalt, line 184: I am not sure whether these are any good as `vInitialiseOnIndex()` cannot be run again when it once ended up in `xEMAC_Fatal` or `xEMAC_Ready`.

htibosch (commented):
Hi @StefanBalt, that is great work! Until now the UltraScale driver would only work with the older single-interface version. You made it standard and compliant with 4.x.x.
In a later stage, I will also test it in hardware.

~~~
pxEndPoint = FreeRTOS_NextEndPoint( pxInterface, pxEndPoint );

if( pxEndPoint != NULL )
{
	/* If there is a second end-point, store the MAC
	 * address at position 4.*/
	XEmacPs_SetMacAddress( pxEMAC_PS, ( void * ) pxEndPoint->xMACAddress.ucBytes, 4 );
}
~~~
You wrote:

> Do we also need this here?

It needs to be worked out.

We can have multiple interfaces each with multiple endpoints, long with different MAC-addresses.

The peripheral has space for 4 primary MAC-addresses. All other addresses will be set in the hash table.

I wish that [PR 1019](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1019) were already approved and merged. That will contain a method of setting the MAC-addresses in a proper way.

~~~
    #if ( ( ipconfigUSE_MDNS == 1 ) && ( ipconfigUSE_IPv6 != 0 ) )
    XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MacAddress.ucBytes ); /* why for IPv6? */
    XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MACAddressIPv6.ucBytes );
~~~
> Why only add MDNS multicast MAC to hash table when ( ipconfigUSE_IPv6 != 0 )?

You are right, that is not correct.

Also setting the "Solicited-Node Multicast" addresses `{ 0x33, 0x33, 0xff, 0, 0, 0 };` only depends on `ipconfigUSE_IPv6` and not on `ipconfigUSE_LLMNR`.

~~~
    /* allow reception of multicast addresses programmed into hash */
    XEmacPs_SetOptions( pxEMAC_PS, XEMACPS_MULTICAST_OPTION );
~~~

> These are also missing in the Zynq7000 port and most likely needed.
> Should I add it?

I think so. The `XEMACPS_MULTICAST_OPTION` refers to `XEMACPS_NWCFG_MCASTHASHEN_MASK`, or MCAST (multicast) HASH (hash) EN (enable). That is indeed important.
~~~
    /* Allow multicast address filtering */
    if( ( Options & XEMACPS_MULTICAST_OPTION ) != 0x00000000U )
    {
        RegNewNetCfg |= XEMACPS_NWCFG_MCASTHASHEN_MASK;
    }
~~~

> But I cannot test it.

By enabling and playing with mDNS?

> Is it really necessary to add ucMACAddress to all endpoints?

The global `ucMACAddress` doesn't exists since release 4.x.

> I would have thought XEmacPs_SetHash( pxEMAC_PS, ( void * ) xLLMNR_MacAddressIPv6.ucBytes ); would be sufficient.

I think so too.

htibosch:
Every Ethernet peripheral, or most of them, has two ways of setting a MAC-address:

- Up to four standard MAC-addresses. Need 100% match
- A hash table with like 64 entries. Each entry is actually a bit.

Hash: when you add a MAC address whose calculated hash value is 12, then bit 12 in the hash table will get set.

( no news for most of you ).

[Emil Popov](https://github.com/evpopov) and me came up with this idea:

- First add primary MAC-addresses, i.e. the hardware addresses of the endpoint.
- Secondly, add the MAC-addresses of the multicast addresses used (e.g. mDNS, LLMNR).

The first 4 (`GMACSA_NUMBER`) addresses will go to the table in the peripheral.
All subsequent addresses will be hashed.

MAC-addresses can be unhashed too: there will be 64 usage counters. When the counter reaches zero again, the hash value will be cleared.
You find an implementation in [DriverSAM](https://github.com/evpopov/FreeRTOS-Plus-TCP/blob/MCast_PR/source/portable/NetworkInterface/DriverSAM/NetworkInterface.c). Maybe we should put it in [Common](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/main/source/portable/NetworkInterface/Common), so all drivers can use it.

For the Ultrascale port, I am thinking of using the same method.

StefanBalt:
Ah youre'right the Zynq7000 port uses `XEMACPS_NWCFG_MCASTHASHEN_MASK` directly instead of using the API function.
I added the discussed multicast fixes to the Zynq7000 port.

I noticed the promiscuous mode in the Zynq7000 port which is somewhat questionable as a default:
```
    /* As 'MCASTHASHEN' doesn't seem to work, use the promiscuous mode so that IPv6 multicast packets are received. */
    /* Allow promiscuous mode. */
    ulValue |= XEMACPS_NWCFG_COPYALLEN_MASK;
```


> 
>     But I cannot test it.
> 
> By enabling and playing with mDNS?
> 
I meant I do not have a Zynq7000 at  hand. I did test mDNS now for the Ultrascale.

I do like the idea of sharing the MAC related code in [Common](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/tree/main/source/portable/NetworkInterface/Common), however I do not have the time to do it at the moment.

My testing is complete for now here is what I did and what I think should still be tested:
- [x] IPv4 including LLMNR, mDNS, Jumbo Frames
- [x] Two endpoints (Virtual interfaces) on the same physical interface
- [ ] IPv6
- [ ] More than one physical interface
- [ ] Zynq7000 changes

tony-josi-aws:
@StefanBalt 

Can you help in fixing the CI checks:

1. To fix the spell checks [these words](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/11046727149/job/30686592267?pr=1187) [those listed in the logs `Run spellings check`] needs to be added to [this file](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/.github/.cSpellWords.txt).
2. To fix the formatting please apply [this patch](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/11046727149/artifacts/1980714949).

Please let us know if you have questions or want us to take care of the CI.

tony-josi-aws (approved):

aggarg (approved):

source/portable/NetworkInterface/Zynq/NetworkInterface.c
```diff
@@ -256,40 +256,50 @@ static BaseType_t xZynqNetworkInterfaceInitialise( NetworkInterface_t * pxInterf
...
-                        ucMACAddress[ 3 ] = pxEndPoint->ipv6_settings.xIPAddress.ucBytes[ 13 ];
-                        ucMACAddress[ 4 ] = pxEndPoint->ipv6_settings.xIPAddress.ucBytes[ 14 ];
-                        ucMACAddress[ 5 ] = pxEndPoint->ipv6_settings.xIPAddress.ucBytes[ 15 ];
-                        XEmacPs_SetHash( pxEMAC_PS, ( void * ) ucMACAddress );
-                    }
-                }
-
                 XEmacPs_SetHash( pxEMAC_PS, ( void * ) xLLMNR_MacAddressIPv6.ucBytes );
             }
             #endif /* if ( ipconfigUSE_IPv6 == 0 ) */
```
htibosch, line 267: The above is not correct, think of this:

~~~
    #if( ipconfigUSE_LLMNR == 1 )
    {
        #if( ipconfigIS_ENABLED( ipconfigUSE_IPv4 ) )
        {
            XEmacPs_SetHash( pxEMAC_PS, ( void * ) xLLMNR_MacAddress.ucBytes );
        }
        #endif

        #if( ipconfigIS_ENABLED( ipconfigUSE_IPv6 ) )
        {
            XEmacPs_SetHash( pxEMAC_PS, ( void * ) xLLMNR_MacAddressIPv6.ucBytes );
        }
        #endif
    }
    #endif /* ipconfigUSE_LLMNR == 1 */
~~~

It means that LLMNR can be enabled for IPv4 and on IPv6 at the same time.

source/portable/NetworkInterface/Zynq/NetworkInterface.c
```diff
@@ -256,40 +256,50 @@ static BaseType_t xZynqNetworkInterfaceInitialise( NetworkInterface_t * pxInterf
...
+            {
+                XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MacAddress.ucBytes );
+            }
+            #else
+            {
+                XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MacAddressIPv6.ucBytes );
+            }
+            #endif /* if ( ipconfigUSE_IPv6 == 0 ) */
+        }
+        #endif /* ( ipconfigUSE_MDNS == 1 ) */
```
htibosch, line 283: and also here:

~~~
    #if ( ipconfigIS_ENABLED( ipconfigUSE_MDNS ) )
        #if ( ipconfigIS_ENABLED( ipconfigUSE_IPv4 ) )
        {
            XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MacAddress.ucBytes );
        }
        #endif

        #if ( ipconfigIS_ENABLED( ipconfigUSE_IPv6 ) )
        {
            XEmacPs_SetHash( pxEMAC_PS, ( void * ) xMDNS_MACAddressIPv6.ucBytes );
        }
        #endif
    #endif /* ipconfigUSE_MDNS */    
~~~

MDNS can also be active on both IP-versions.

source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -320,22 +504,37 @@ BaseType_t xNetworkInterfaceOutput( NetworkBufferDescriptor_t * const pxBuffer,
...
+                ( pxICMPPacket->xIPHeader.ucNextHeader == ipPROTOCOL_ICMP_IPv6 ) )
+            {
+                /* The EMAC will calculate the checksum of the IP-header.
+                 * It can only calculate protocol checksums of UDP and TCP,
+                 * so for ICMP and other protocols it must be done manually. */
+                usGenerateProtocolChecksum( pxBuffer->pucEthernetBuffer, pxBuffer->xDataLength, pdTRUE );
+            }
+        #endif /* ipconfigUSE_IPv6 */
+
         if( ( pxPacket->xICMPPacket.xEthernetHeader.usFrameType == ipIPv4_FRAME_TYPE ) &&
```
htibosch, line 520: The driver should not bother about ipV4 frames, unless `ipconfigUSE_IPv4` is defined.

htibosch (approved):
Thank you so far, [Stefan](https://github.com/StefanBalt) for this PR. I don't have the time to test it, I hope you did.
It will still have some work to do, but let's wait for the multi-cast feature first.

tony-josi-aws:
@StefanBalt 
Since the multicast PR needs more time to be merged, can you please help in fixing the following proposals from @htibosch so that we can go ahead with merging the branch:

- [LLMNR ](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1187#discussion_r1780480251)
- [MDNS ](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1187#discussion_r1780481174)

Please feel free to let us know if you prefer us to make the suggested changes.
Thank you.

StefanBalt:
Sure, I will push a fix (probably today).

source/portable/NetworkInterface/Zynq/NetworkInterface.c
StefanBalt: Ah of course IPv4 and v6 can be both enabled. Done

source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
StefanBalt: Done

StefanBalt:
> I don't have the time to test it, I hope you did.

As mentioned earlier, I tested some of the features but not all (i.e. IPv6, DHCP and more than one physical interface missing).

tony-josi-aws (dismissed):

source/portable/NetworkInterface/xilinx_ultrascale/NetworkInterface.c
```diff
@@ -444,6 +654,56 @@ BaseType_t xGetPhyLinkStatus( void )
...
+/*-----------------------------------------------------------*/
+#if defined( ipconfigIPv4_BACKWARD_COMPATIBLE ) && ( ipconfigIPv4_BACKWARD_COMPATIBLE == 1 )
+
+/* Do not call the following function directly. It is there for downward compatibility.
+ * The function FreeRTOS_IPInit() will call it to initialise the interface and end-point
+ * objects.  See the description in FreeRTOS_Routing.h. */
+    NetworkInterface_t * pxFillInterfaceDescriptor( BaseType_t xEMACIndex,
+                                                    NetworkInterface_t * pxInterface )
+    {
+        pxUltrascale_FillInterfaceDescriptor( xEMACIndex, pxInterface );
```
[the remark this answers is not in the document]
ActoryOu: This might be missed. But overall it looks good to me.

ActoryOu (approved):

tony-josi-aws (approved):

aggarg (approved):

PR #1221, opened 2025-01-24
Improves handling of IPv4 broadcast addresses.

evpopov:
Description
-----------
The current code incorrectly assumes that broadcast addresses only end in x.x.x.255
When a frame enters the stack, all broadcast tests must be made against the 255.255.255.255 address and against the **specific broadcast of the endpoint that the frame was assigned**
When an ARP lookup is attempted, we need to check if the user is attempting to send a packet that may be considered a broadcast on any of our IPv4 endpoints. Current code was only calling an outgoing packet a "broadcast" is said packet ended in 0xFF.
This PR attempts to correct both of the above.

Additionally, I've taken the liberty to substitute ipBROADCAST_IP_ADDRESS with FREERTOS_INADDR_BROADCAST and put the new definition in a public header file. Here's my reasoning:

- INADDR_BROADCAST is a well-known definition that in my opinion is widely used. Just like INADDR_ANY. The original definition's name is obscure.
- The original definition is supposed to be "inaccessible" or "hidden" to the user because it was in FreeRTOS_IP_Private.h I did not see a good reason for this.

I have also probably broken a bunch of test cases with this and I'd really appreciate it if someone can help with fixing the old tests and maybe even add tests that verify the new changes. For example test that verifies 192.168.0.7 as a broadcast if the endpoint is 192.168.0.1/29, etc...

### DIFF
--- a/source/FreeRTOS_IPv4.c
+++ b/source/FreeRTOS_IPv4.c
@@ -224,7 +224,6 @@ BaseType_t xIsIPv4Broadcast( uint32_t ulIPAddress,
                              struct xNetworkEndPoint ** ppxEndPoint )
 {
     BaseType_t xIsBroadcast;
-    uint32_t ulEndPointBroadcast;
     NetworkEndPoint_t * pxEndPoint;
 
     /* Assign a default answer based on the "global" broadcast.	 This way
--- a/source/FreeRTOS_IPv4.c
+++ b/source/FreeRTOS_IPv4.c
@@ -211,6 +211,66 @@ BaseType_t xIsIPv4Multicast( uint32_t ulIPAddress )
 /*-----------------------------------------------------------*/
 
 /**
+ * @brief Checks if the IP address matches the global 255.255.255.255 broadcast address or
+ * the broadcast address for any of our IPv4 endpoints.
+ *
+ * @param[in] ulIPAddress The IP address being checked.
+ *
+ * @param[out] ppxEndPoint Pointer to an end-point where we store the endpoint whose broadcast address we matched. Or NULL if no IPv4 endpoints were found.
+ *
+ * @return pdTRUE if the IP address is a broadcast address or else, pdFALSE.
+ */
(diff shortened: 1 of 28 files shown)


source/FreeRTOS_IPv4.c
```diff
@@ -210,6 +210,66 @@ BaseType_t xIsIPv4Multicast( uint32_t ulIPAddress )
...
+ *
+ * @param[out] ppxEndPoint Pointer to an end-point where we store the endpoint whose broadcast address we matched. Or NULL if no IPv4 endpoints were found.
+ *
+ * @return pdTRUE if the IP address is a broadcast address or else, pdFALSE.
+ */
+BaseType_t xIsIPv4Broadcast( uint32_t ulIPAddress,
+                             struct xNetworkEndPoint ** ppxEndPoint )
+{
+    BaseType_t xIsBroadcast;
+    uint32_t ulEndPointBroadcast;
```
tony-josi-aws, line 227: Unused variable `ulEndPointBroadcast`

source/include/FreeRTOS_IP_Private.h
```diff
@@ -258,8 +258,6 @@ struct xPacketSummary
     uint16_t usProtocolBytes;              /**< The total length of the protocol data. */
 };
 
-#define ipBROADCAST_IP_ADDRESS               0xffffffffU
```
tony-josi-aws, line 261: There are few more references to `ipBROADCAST_IP_ADDRESS` in the code - STM32 network interfaces and UTs

tony-josi-aws:
Thanks @evpopov for this PR, I have added some minor comments.

evpopov:
Thanks for catching those @tony-josi-aws 
I addressed all of your comments, but I cannot run the unit tests so please verify

tony-josi-aws (approved):

kar-rahul-aws (approved):

PR #1223, opened 2025-02-04
Set MDNS AA bit while answering MDNS request.

ActoryOu:
Set MDNS AA bit while answering MDNS request.

Description
-----------

From forum question [FreeRTOS+TCP mDNS behaviour on receiving request](https://forums.freertos.org/t/freertos-tcp-mdns-behaviour-on-receiving-request/22455), the AA bit of MDNS reply is not set correctly.

Here is the description in [RFC6762 18.4](https://datatracker.ietf.org/doc/html/rfc6762#section-18.4).
> In response messages for Multicast domains, the Authoritative Answer
> bit MUST be set to one (not setting this bit would imply there’s some
> other place where “better” information may be found) and MUST be
> ignored on reception.

So in this PR, we add the AA bit for MDNS query. And keep the behavior unchanged for LLMNR query.

Test Steps
-----------

1. Add unit test to parse MDNS request.
2. Check the flag in the DNS header to see if it's 0x8400. (with QR bit and AA bit set)

Checklist:
----------

Related Issue
-----------

https://forums.freertos.org/t/freertos-tcp-mdns-behaviour-on-receiving-request/22455

### DIFF
--- a/source/FreeRTOS_DNS_Parser.c
+++ b/source/FreeRTOS_DNS_Parser.c
@@ -545,11 +545,22 @@
                             if( ( pxNetworkBuffer != NULL ) )
                             {
                                 pxAnswer = ( ( LLMNRAnswer_t * ) xSet.pucByte );
+
                                 /* We leave 'usIdentifier' and 'usQuestions' untouched */
-                                vSetField16( xSet.pxDNSMessageHeader, DNSMessage_t, usFlags, dnsLLMNR_FLAGS_IS_RESPONSE ); /* Set the response flag */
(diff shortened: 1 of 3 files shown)


htibosch:
I encounter the same problem when I lookup a device using mDNS from Linux, even though the authoritative flag is set.
Now I will try leaving out the questions asked. Which is not so easy.

htibosch:
It appears that the implementation of mDNS in FreeRTOS_DNS_Parser.c needs two changes:

- Set the authoritative flag in the reply
- Do not include Questions when replying

I tested the new version with Linux and Windows as clients, and that both worked.

I committed the changes in PR #1292.

I think that this PR can now be closed. Thank you @ActoryOu

PR #1226, opened 2025-02-16
Fix DHCP v4/v6 race condition and memory leak

tony-josi-aws:
Description
-----------
This PR fixes DHCP - v4 and v6 race condition and memory leak.
While the DHCP state machine is processing response from a DHCP server, causing it to change the DHCP state, if another DHCP server responds to the DUT, state mismatch will cause the `vDHCPProcess` function to result in an infinite loop.

Thanks [Raghav](https://forums.freertos.org/u/Raghav) for reporting this issue in the [forum post](https://forums.freertos.org/t/code-stuck-in-the-vdhcpprocess-function-mostly-when-dhcp-state-machine-is-ewaitingacknowledge/22484).

Along with this fix, the zero copy buffers were not freed in the DHCPv6 state machine as well as in DHCPv4 corner cases, those are also corrected in this PR.

Test Steps
-----------

STM32F4 - DHCP v4 tests

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/test/unit-test/FreeRTOS_DHCP/FreeRTOS_DHCP_utest.c
+++ b/test/unit-test/FreeRTOS_DHCP/FreeRTOS_DHCP_utest.c
@@ -1326,7 +1326,7 @@ void test_vDHCPProcess_eLeasedAddress_InCorrectState_Loop( void )
     pxNetworkEndPoints = pxEndPoint;
     
     /* Expect these arguments. */
-    FreeRTOS_recvfrom_Stub( FreeRTOS_recvfrom_ResetAndIncorrectStateWithSocketAlreadyCreated_LoopedCall );
+    FreeRTOS_recvfrom_Stub( FreeRTOS_recvfrom_LoopedCall );
 
     FreeRTOS_ReleaseUDPPayloadBuffer_Expect( pucUDPBuffer );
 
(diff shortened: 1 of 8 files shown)


test/unit-test/FreeRTOS_DHCP/FreeRTOS_DHCP_utest.c
```diff
@@ -341,6 +341,12 @@ void test_vDHCPProcess_ResetAndIncorrectStateWithSocketAlreadyCreated( void )
 
         /* Expect these arguments. */
         FreeRTOS_recvfrom_ExpectAndReturn( xDHCPv4Socket, NULL, 0UL, FREERTOS_ZERO_COPY + FREERTOS_MSG_PEEK, NULL, NULL, 0 );
+
+        /* Ignore the source address and source address
+         * length argument though. */
+        FreeRTOS_recvfrom_IgnoreArg_pxSourceAddress();
```
ActoryOu, line 347: Is there any test case verifying multiple DHCP server scenarios?

tony-josi-aws: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1226/files#diff-33d34c64f4cfe318de92341b204dd23da6d4594abd3a5a81d3a4f360be3a42f8R1347-R1437

ActoryOu (approved):

kar-rahul-aws (approved):

PR #1227, opened 2025-02-21
Update Zynq portable for Xilinx SDT drivers

mike919192:
Description
-----------
Sometime around 2023, Xilinx started introducing breaking changes to their drivers along with their new IDE. As an example you can look at this file where they use the macro SDT to detect whether the new or old drivers are being used: [https://github.com/Xilinx/embeddedsw/blob/master/XilinxProcessorIPLib/drivers/emacps/examples/xemacps_example_intr_dma.c](https://github.com/Xilinx/embeddedsw/blob/master/XilinxProcessorIPLib/drivers/emacps/examples/xemacps_example_intr_dma.c)

This PR is updating the Zynq portable code so that it works with both the new and old drivers.

Test Steps
-----------
I have tested on real hardware for a few days now.  Testing included using both the IDE with new drivers (Vitis 2024.2) and old drivers (Vitis Classic 2024.2).

The changes are all within the portable layer, so I don't believe they are covered by any unit tests.

Checklist:
----------

Related Issue
-----------

I posted some more details in this discussion: [https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/discussions/1225](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/discussions/1225)


tony-josi-aws:
@mike919192 

Thanks for contributing to FreeRTOS+TCP.
We will review the PR and get back to you.

source/portable/NetworkInterface/Zynq/x_emacpsif.h
```diff
@@ -38,8 +38,19 @@
...
+        #define ZYNQ_SCUGIC_0_BASEADDR          XPAR_XSCUGIC_0_BASEADDR
+        #define ZYNQ_ETHERNET_0_BASEADDR        XPAR_XEMACPS_0_BASEADDR
+        #if ( XPAR_XEMACPS_NUM_INSTANCES > 1 )
+            #define ZYNQ_ETHERNET_1_BASEADDR    XPAR_XEMACPS_1_BASEADDR
+        #endif
+    #else
+        #define ZYNQ_SCUGIC_0_BASEADDR          XPAR_PS7_SCUGIC_0_BASEADDR
+        #define ZYNQ_ETHERNET_0_BASEADDR        XPAR_PS7_ETHERNET_0_BASEADDR
+        #if ( XPAR_XEMACPS_NUM_INSTANCES > 1 )
+            #define ZYNQ_ETHERNET_1_BASEADDR    XPAR_PS7_ETHERNET_1_BASEADDR
```
tony-josi-aws, line 51: When I'm building on an enviornemnt that doesnt have `SDT` defined and `XPAR_XEMACPS_NUM_INSTANCES` > 1, there is build failure:
```
Description	Resource	Path	Location	Type
'XPAR_PS7_ETHERNET_1_DEVICE_ID' undeclared here (not in a function); did you mean 'XPAR_PS7_ETHERNET_0_DEVICE_ID'?	NetworkInterface.c	/aws_demos/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq	line 185	C/C++ Problem

```

source/portable/NetworkInterface/Zynq/NetworkInterface.c
```diff
@@ -173,14 +173,18 @@ XEmacPs_Config mac_configs[ XPAR_XEMACPS_NUM_INSTANCES ] =
...
+        #endif
+        .BaseAddress  = ZYNQ_ETHERNET_0_BASEADDR       /**< Physical base address of IPIF registers */
         },
     #if ( XPAR_XEMACPS_NUM_INSTANCES > 1 )
         [ 1 ] =
         {
-        .DeviceId    = XPAR_PS7_ETHERNET_1_DEVICE_ID, /**< Unique ID  of device */
-        .BaseAddress = XPAR_PS7_ETHERNET_1_BASEADDR   /**< Physical base address of IPIF registers */
+        #ifndef SDT
+            .DeviceId = XPAR_PS7_ETHERNET_1_DEVICE_ID, /**< Unique ID  of device */
```
tony-josi-aws, line 185: When I'm building on an environment that doesn't have SDT defined and XPAR_XEMACPS_NUM_INSTANCES > 1, there is a build failure:
```
Description	Resource	Path	Location	Type
'XPAR_PS7_ETHERNET_1_DEVICE_ID' undeclared here (not in a function); did you mean 'XPAR_PS7_ETHERNET_0_DEVICE_ID'?	NetworkInterface.c	/aws_demos/libraries/freertos_plus/standard/freertos_plus_tcp/source/portable/NetworkInterface/Zynq	line 185	C/C++ Problem

```

source/portable/NetworkInterface/Zynq/x_emacpsif.h
```diff
@@ -38,8 +38,19 @@
     extern "C" {
     #endif
 
-    #define XPAR_PS7_ETHERNET_1_DEVICE_ID    1
-    #define XPAR_PS7_ETHERNET_1_BASEADDR     0xE000C000
```
tony-josi-aws, line 42: Any particular reason these macros are removed now?

mike919192: Thanks for the comment.  This must be the cause of the other errors so I will comment here.

I should have mentioned that my dev board only has 1 ethernet, so I wasn't able to check this case.  `XPAR_PS7_ETHERNET_0_DEVICE_ID` and `XPAR_PS7_ETHERNET_0_BASEADDR` are defined in the xilinx file `xparameters.h`.  I had assumed that xilinx would provide these definitions for additional ethernet, but maybe it is a bug that they are missing from the xilinx file.  The deleted lines may have been a workaround for that.

I could look at your `xparameters.h` file if you post it.  But it sounds like `XPAR_PS7_ETHERNET_1_DEVICE_ID` is missing.  In that case it is no problem to revert the deletion.  Is `XPAR_PS7_ETHERNET_1_BASEADDR` also missing from the `xparameters.h`?

tony-josi-aws: > But it sounds like XPAR_PS7_ETHERNET_1_DEVICE_ID is missing. In that case it is no problem to revert the deletion. Is XPAR_PS7_ETHERNET_1_BASEADDR also missing from the xparameters.h?

Yes, both macros are missing; I just defined `XPAR_XEMACPS_NUM_INSTANCES` to 2 to just check the build. I believe ideally the `xparameters.h` should be updated with macros.

For backward compatibility for anyone already using the library with `XPAR_XEMACPS_NUM_INSTANCES` > 1, I believe it would be good to keep those macros optionally defined if they're not defined in `xparameters.h` maybe with `#ifndef`'s.

mike919192: I have put the defines back in with `#ifndef`.  They will be active when SDT is not defined.

source/portable/NetworkInterface/Zynq/NetworkInterface.c
mike919192: This should be resolved after adding the defines back in.

source/portable/NetworkInterface/Zynq/x_emacpsif.h
mike919192: This should be resolved after adding the defines back in.

ActoryOu (approved):

Issue #1235, opened 2025-03-19
RST with not matching sequence number causes continuously sending ACK/RST


mj-servus:
**Describe the bug**
The target and the host remain in a loop continuously sending RST and challenge ACK.

The reason is a RST packet from the HOST with the sequence number inside the current receive window, but not exactly matching the expected value.
In that case the target responds with an challenge ACK (https://datatracker.ietf.org/doc/html/rfc5961#section-3.2). The received frame is the base for the challenge ACK, so the sequence number sent by the host is used for the challenge ACK message. As the host already closed the connection the response is again a RST with the same sequence number.

The ping pong of the messages is done until an other message is sent to the host, including the expected sequence number. The RST response of the other message finally causes the connection to close.

**Target**
- FreeRTOS-Plus-TCP network stack version 4.2.0
- Development board: Proprietary board with STM32H7 CPU and RS9116 WLAN module.
- Instruction Set Architecture: cortex-m7
- Toolchain and version: GNU Arm Embedded Toolchain 10.3-2021.10

**Host**
- Host OS: Windows Server 2016 Datacenter
- Version: 1607

**To Reproduce**
- Establish a connection to the host
- Interrupt the data transfer for about 25s (WLAN disconnect)
- Cause the host to abort the connection (in our case implicit by the Websocket server)
- Continue the data transfer

**Expected behavior**
In case a RST frame with the sequence number inside the current receive window is received, one challenge ACK should be sent. In case the host responds again with an RST the connection should be closed.

**Screenshots**
![Image](https://github.com/user-attachments/assets/339426ff-07d2-418f-8392-ebf66515bdee)

**Wireshark logs**
The Wireshark logs where taken at the target.
Capture: [RST_ACK.zip](https://github.com/user-attachments/files/19335418/RST_ACK.zip)
Target: 10.1.126.248
Host: 10.1.126.11

**Additional context**
I did an experiment using the expected sequence number in the challenge ACK message. By doing so, the host responded with a RST with the expected sequence number and the connection was closed after the second RST.

mj-servus:
I added the changes of my experiment as pull request: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1236

tony-josi-aws:
Closing this issue as the PR is merged

PR #1236, opened 2025-03-20
Use the expected sequence number for challenge ACK messages

mj-servus:
That avoids avoid endless ACK/RST retransmissions.

Description
-----------
See #1235

Test Steps
-----------

Checklist:
----------

Related Issue
-----------
#1235

### DIFF
--- a/source/FreeRTOS_TCP_Transmission.c
+++ b/source/FreeRTOS_TCP_Transmission.c
@@ -1352,7 +1352,7 @@
  *        unexpected but still within the window.
  *
  * @param[in] pxNetworkBuffer The network buffer descriptor with the packet.
- * @param[in] ulCurrentSequenceNumber The current sequence number of the connection.
+ * @param[in] ulCurrentSequenceNumber The current expected sequence value by the connection.
  * @param[in] ulOurSequenceNumber The SEQ number to send out.
  *
  * @return Returns the value back from #prvTCPSendSpecialPacketHelper.
--- a/source/FreeRTOS_TCP_Transmission.c
+++ b/source/FreeRTOS_TCP_Transmission.c
@@ -1352,11 +1352,20 @@
  *        unexpected but still within the window.
  *
  * @param[in] pxNetworkBuffer The network buffer descriptor with the packet.
+ * @param[in] ulCurrentSequenceNumber The current sequence number of the connection.
  *
  * @return Returns the value back from #prvTCPSendSpecialPacketHelper.
  */
-    BaseType_t prvTCPSendChallengeAck( NetworkBufferDescriptor_t * pxNetworkBuffer )
+    BaseType_t prvTCPSendChallengeAck( NetworkBufferDescriptor_t * pxNetworkBuffer,
+                                       uint32_t ulCurrentSequenceNumber )
     {
+        ProtocolHeaders_t * pxProtocolHeaders = ( ( ProtocolHeaders_t * )
+                                                  &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER + uxIPHeaderSizePacket( pxNetworkBuffer ) ] ) );
+
+        /* Correct the sequence number to avoid an endless ACK/RST sequence. When sending the
+         * same number again the next RST will not match as well. */
+        pxProtocolHeaders->xTCPHeader.ulSequenceNumber = FreeRTOS_htonl( ulCurrentSequenceNumber );
+
         return prvTCPSendSpecialPacketHelper( pxNetworkBuffer, tcpTCP_FLAG_ACK );
     }
     /*-----------------------------------------------------------*/
--- a/source/FreeRTOS_TCP_Transmission.c
+++ b/source/FreeRTOS_TCP_Transmission.c
@@ -1353,18 +1353,26 @@
  *
  * @param[in] pxNetworkBuffer The network buffer descriptor with the packet.
  * @param[in] ulCurrentSequenceNumber The current sequence number of the connection.
+ * @param[in] ulOurSequenceNumber The SEQ number to send out.
  *
  * @return Returns the value back from #prvTCPSendSpecialPacketHelper.
  */
     BaseType_t prvTCPSendChallengeAck( NetworkBufferDescriptor_t * pxNetworkBuffer,
-                                       uint32_t ulCurrentSequenceNumber )
+                                       uint32_t ulCurrentSequenceNumber,
+                                       uint32_t ulOurSequenceNumber )
     {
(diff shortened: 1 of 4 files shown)


source/FreeRTOS_TCP_Transmission.c
```diff
@@ -1352,11 +1352,19 @@
...
  * @param[in] pxNetworkBuffer The network buffer descriptor with the packet.
+ * @param[in] ulCurrentSequenceNumber The current sequence number of the connection.
  *
  * @return Returns the value back from #prvTCPSendSpecialPacketHelper.
  */
-    BaseType_t prvTCPSendChallengeAck( NetworkBufferDescriptor_t * pxNetworkBuffer )
+    BaseType_t prvTCPSendChallengeAck( NetworkBufferDescriptor_t * pxNetworkBuffer, uint32_t ulCurrentSequenceNumber )
     {
+        ProtocolHeaders_t * pxProtocolHeaders = ( ( ProtocolHeaders_t * )
+            &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER + xIPHeaderSize( pxNetworkBuffer ) ] ) );
```
tony-josi-aws, line 1362: `xIPHeaderSize` seems not defined; I would suggest using `uxIPHeaderSizePacket` instead.

mj-servus: Thanks for the suggestion. I did the experiments with an older version.

AhmedIsmail02:
@mj-servus 

Thanks for your contribution.

I'm trying to understand the changes, from what I've seen in the wire shark capture on [issue!1235](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/1235) and AFAIK, the target device should have dropped the connection after the **second** RST packet was received as this packet contains a **sequence number** equals to the challenge ACK packet **acknowledgement number**. However the target didn't reset the connection, that's why i believe @tony-josi-aws  has suggested [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/1235#issuecomment-2739681661) that the second RST packet should be handled separately as in we need to compare the **second RST packet's sequence number** with the **challenge ACK packets's acknowledgement number**. However, I can't see this being part of the changes, am I missing something?

tony-josi-aws:
@mj-servus

Thanks for your contribution to FreeRTOS+TCP. I just tested your implementation locally and verified that it works by not leading to continuously sending ACK/RST.

Attaching PCAP and screen capture of the tests for anyone interested:

wshark

the second RST (packet 2503) from the external device (in this case, a server running on Windows OS) has triggered a closing of the socket on the FreeRTOS+TCP device:

serial_console

PCAP file: [wshark.zip](https://github.com/user-attachments/files/19416368/wshark.zip)

A further improvement to be fully complaint with [RFC 5961](https://datatracker.ietf.org/doc/html/rfc5961) [[3.2](https://datatracker.ietf.org/doc/html/rfc5961#section-3.2).  Mitigation] is to set the sequence number of the challenges ACK as per the TCP sockets `SND.NXT` (`pxSocket->u.xTCP.xTCPWindow.ulOurSequenceNumber`), similar to how acknowledge is set via `pxProtocolHeaders->xTCPHeader.ulSequenceNumber = FreeRTOS_htonl( ulCurrentSequenceNumber );`, as they are swapped inside the `prvTCPSendSpecialPacketHelper`.

In my test case, the acknowledgement number of the first RST was what was expected by the FreeRTOS+TCP device, so swapping it to the sequence number was as per RFC:

```
 TCP MUST
      send an acknowledgment (challenge ACK):

      <SEQ=SND.NXT><ACK=RCV.NXT><CTL=ACK>
```

But I believe in the case of a malicious RST this might not be the case, hence it might be more reasonable to use `pxSocket->u.xTCP.xTCPWindow.ulOurSequenceNumber` to be set as `SEG.SEQ` for the challenge ACK than swapping the first RST's ACK number as it is currently done.

So, in my opinion, it makes sense to also take `pxSocket->u.xTCP.xTCPWindow.ulOurSequenceNumber` as an argument to `prvTCPSendChallengeAck`.

Feel free to share your opinion, and if you wish to add those changes to the PR please do so or let us know.

@AhmedIsmail02 

Thanks for taking a look at this PR. 
With respect to my response to issue #1235 , the sequence number and the acknowledge number of the initial RST packet are [swapped and are used in the challenge ACK](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/FreeRTOS_TCP_Transmission_IPv4.c#L189) but the problem was that the `SEG.SEQ` and `SEG.ACK` of the resultant challenge ACK were not as per RFC and hence the second RST from the external device was not meeting the conditions for a socket close.

In this PR, @mj-servus corrects that by updating the sequence number of the first RST which is used as the base packet to form the challenge ACK. Updating the sequence number causes the acknowledgement number of the challenge ACK to be updated with that sequence number because of the swapping mentioned in the previous paragraph.

source/FreeRTOS_TCP_Transmission.c
```diff
@@ -1352,11 +1352,20 @@
...
-    BaseType_t prvTCPSendChallengeAck( NetworkBufferDescriptor_t * pxNetworkBuffer )
+    BaseType_t prvTCPSendChallengeAck( NetworkBufferDescriptor_t * pxNetworkBuffer,
+                                       uint32_t ulCurrentSequenceNumber )
     {
+        ProtocolHeaders_t * pxProtocolHeaders = ( ( ProtocolHeaders_t * )
+                                                  &( pxNetworkBuffer->pucEthernetBuffer[ ipSIZE_OF_ETH_HEADER + uxIPHeaderSizePacket( pxNetworkBuffer ) ] ) );
+
+        /* Correct the sequence number to avoid an endless ACK/RST sequence. When sending the
+         * same number again the next RST will not match as well. */
+        pxProtocolHeaders->xTCPHeader.ulSequenceNumber = FreeRTOS_htonl( ulCurrentSequenceNumber );
```
tony-josi-aws, line 1367: Would be great to mention in the comment that the sequence number will be used as the ACK number and vice versa for the challenge ACK later inside the `prvTCPSendSpecialPacketHelper`.
Also may include:

```
TCP MUST
      send an acknowledgment (challenge ACK):

      <SEQ=SND.NXT><ACK=RCV.NXT><CTL=ACK>
```

tony-josi-aws (approved):

AhmedIsmail02:
@tony-josi-aws Thanks for your detailed explanation.
Changes LGTM.

AhmedIsmail02 (approved):

ActoryOu (approved):

PR #1271, opened 2025-06-23
Add configuration option to prevent configASSERT checks on ipBUFFER_PADDING

blanco-ether:
Description
-----------
Added ipconfigSUPPRESS_BUFFER_PADDING_CHECK to prevent configASSERT checks on ipBUFFER_PADDING in FreeRTOS_IP_Utils.c/vPreCheckConfigs( void )

* Defaults to ipconfigDISABLE to build with configASSERT checks in place.
* User settable to ipconfigENABLE to eliminate configASSERT checks.

Test Steps
-----------
Observe that when source built while ipconfigSUPPRESS_BUFFER_PADDING_CHECK either not defined in user configuration or defined as ipconfigDISABLE, the configASSERTs to validate optimal settings for ipBUFFER_PADDING are also built as in existing code base.  The configASSERTS are not built when the default setting is overridden in a user configuration by defining ipconfigSUPPRESS_BUFFER_PADDING_CHECK as ipconfigENABLE.

Checklist:
----------

Related Issue
-----------
Setting ipconfigPACKET_FILLER_SIZE to 0 is necessary when porting to a MAC controller that does not have an option to pad the start of a received packet within a word boundary to optimize access to IP fields.  Setting ipconfigPACKET_FILLER_SIZE to 0 causes configASSERT checks related to validating ipBUFFER_PADDING to fail.

Forum discussion: https://forums.freertos.org/t/assert-fail-using-zero-copy-buffers-porting-freertos-plus-tcp-v3-1-0-to-v4-2-2/23254

### DIFF
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -1004,6 +1004,8 @@ void vPreCheckConfigs( void )
     {
         size_t uxSize;
 
+        #if ( ipconfigSUPPRESS_BUFFER_PADDING_CHECK == 0 )
+
         /* Check if ipBUFFER_PADDING has a minimum size, depending on the platform.
          * See FreeRTOS_IP.h for more details. */
         #if ( UINTPTR_MAX > 0xFFFFFFFFU )
@@ -1024,6 +1026,8 @@ void vPreCheckConfigs( void )
          * fields in the protocol headers.
          */
         configASSERT( ( ( ( ipSIZE_OF_ETH_HEADER ) + ( ipBUFFER_PADDING ) ) % 4U ) == 0U );
+
+        #endif
 
         /* LCOV_EXCL_BR_START */
         uxSize = ipconfigNETWORK_MTU;
--- a/source/FreeRTOS_IP_Utils.c
+++ b/source/FreeRTOS_IP_Utils.c
@@ -1006,28 +1006,27 @@ void vPreCheckConfigs( void )
 
         #if ( ipconfigSUPPRESS_BUFFER_PADDING_CHECK == 0 )
 
-        /* Check if ipBUFFER_PADDING has a minimum size, depending on the platform.
-         * See FreeRTOS_IP.h for more details. */
-        #if ( UINTPTR_MAX > 0xFFFFFFFFU )
+            /* Check if ipBUFFER_PADDING has a minimum size, depending on the platform.
+             * See FreeRTOS_IP.h for more details. */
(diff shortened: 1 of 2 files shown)


tony-josi-aws:
@blanco-ether 
Thanks for your contribution. 
Can you apply the formatting patch on the PR, which is available [here](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/15832428120/artifacts/3388755178) to fix the CI check?

tony-josi-aws:
@devprodest 
Do you mean having platform-specific alignment macros for different supported architectures?

aggarg (approved):

moninom1 (approved):

blanco-ether:
If I understand what you're asking, no.

Some more detail to help make things clearer...

A user who ports to a MAC that does not have a feature to padd the start of
a received packet (2 bytes) will likely need to
set ipconfigPACKET_FILLER_SIZE to 0
in their FreeRTOSIPConfig.h.  Maybe more specifically, when a user
implements zero copy buffers on a port to a MAC that does not support any
way to automatically
add padding.  It may be that when porting to a MAC that lacks the padding
feature but not implementing zero copy buffers they will not need to change
ipconfigPACKET_FILLER_SIZE in their FreeRTOSIPConfig.h.

In our case, I ported FreeRTOS-Plus-TCP to the MAC in the processor in our
device (Actel/Microsemi, now Microchip: A2F200) years ago and implemented
zero copy
buffers right off to save memory.  I needed to set
ipconfigPACKET_FILLER_SIZE to 0 to make it work.  I recently needed to
update to a newer version of FreeRTOS-Plus-TCP
to take advantage of a bug fix related to an orphaned socket that was
closed before being accepted caused FreeRTOS_accept to return a pointer to
freed memory.

The port to the new version required minor changes and works fine except
fails asserts because of the need to set ipconfigPACKET_FILLER_SIZE to 0.
The configASSERTs will fail when ipconfigPACKET_FILLER_SIZE is set to less
than 2.

Hope that helps.

Thanks

ActoryOu (approved):

devprodest:
> @devprodest Do you mean having platform-specific alignment macros for different supported architectures?

Yes, unfortunately it is.

tony-josi-aws:
@devprodest 
Since those are dependent on each HW platform and also design choices of the app developers, I believe it's better to keep the padding information as a config macro (we currently dont have recommended platform-specific macro definitions for any network interfaces; instead, those are in readme/comments/warnings in network interface folders) from a maintenance perspective.

Issue #1277, opened 2025-07-28
[BUG] ETH_TxPacketConfig undefined


george-palmsens:
**Describe the bug**

On version 4.3.3

When compiling the new implementation for STM32H7, `NetworkInterface.c` fails to compile with unknown type `ETH_TxPacketConfig`.

This seems correct - this type is defined in the F4 driver but not in the H7 driver.

Changing all occurances of `ETH_TxPacketConfigTypeDef` to `ETH_TxPacketConfig` in the H7 driver files compiles fine and behaves correctly.

**Target**

STM32H7xx GCC

**Host**
Windows 11

kstribrnAmzn:
Thanks for raising this bug! It seems like this must have been missed during testing when introduced. At some point in the past the structure was renamed. Looking at the latest header (which I believe to be [this](https://github.com/STMicroelectronics/stm32f4xx-hal-driver/blob/master/Inc/stm32f4xx_hal_eth.h#L111-L155)). Looks like it's safe to rename the structure.

PR #1278, opened 2025-07-28
Rename ETH_TxPacketConfigTypeDef in STM32 drivers

george-palmsens:
Address #1277 - STM32 drivers that are not the F4 do not compile `NetworkInterface.c` because of incorrect naming of `ETH_TxPacketConfigTypeDef ` (should be `ETH_TxPacketConfig`).

Description
-----------
Rename `ETH_TxPacketConfigTypeDef` to `ETH_TxPacketConfig` in all drivers.

Test Steps
-----------
It didn't compile before, it does compile now. And on our hardware it correctly accepts connections and communicates back and forth.

Checklist:
----------
- [X] I have tested my changes. No regression in existing tests.

Related Issue
-----------
- #1277


kstribrnAmzn:
Thanks for opening this PR! I'm taking a look now.

kstribrnAmzn:
I took at the [latest driver here](https://github.com/STMicroelectronics/stm32f4xx-hal-driver/blob/master/Inc/stm32f4xx_hal_eth.h#L111-L155) it seems like it would be better to swap the H4 to name the structure `ETH_TxPacketConfigTypeDef` instead of modifying all the other ports. If you do this you should only have to rename the F4 port files, legacy port files, and [this call](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/19134e1dc5adce673fa2f19f032e9245d8fc44e4/source/portable/NetworkInterface/STM32/NetworkInterface.c#L606).

What are your thoughts?

tony-josi-aws:
@george-palmsens Thanks for the PR.

The latest updated version looks good to me; as @kstribrnAmzn suggested, it seems like ST is making the structure name consistent across different HW platforms (for example, previously the v1.0.0 version of STM32H5 HAL had the struct named as [ETH_TxPacketConfig](https://github.com/STMicroelectronics/stm32h5xx-hal-driver/blob/v1.0.0/Inc/stm32h5xx_hal_eth.h#L122-L163), newer versions name it [ETH_TxPacketConfigTypeDef](https://github.com/STMicroelectronics/stm32h5xx-hal-driver/blob/v1.5.0/Inc/stm32h5xx_hal_eth.h#L121-L162)).

The reason for keeping the ST HAL drivers along the network interfaces in the FreeRTOS+TCP repo was primarily due to the inconsistencies across the STM32 HAL, so that anyone who likes to change the driver version could use them as a reference. [_[refer](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/804#issuecomment-2095695778)_]

> My gut says the legacy drivers should not be changed

I agree with that; the legacy network interfaces of STM32 platforms are supposed to be used with the driver provided along with the network interfaces (it has slight fixes/modifications to work with those legacy interfaces). It's better to keep it as it is.

tony-josi-aws (approved):

aggarg (approved):

tony-josi-aws:
Yes, this will be included in an upcoming release. While I cannot provide specific timing details at this moment, I recommend using the main branch to proceed with your project. You can then transition to the official release version once it becomes available.

PR #1253, opened 2025-05-23
STM32 network interface bug fixes

ldehenni:
STM32 network interface bug fixes

Description
-----------
All modifications are isolated to the STM32 networkinterface.c file.

A documented bug in the STM32 HAL drivers has since been fixed for the F7 port (https://github.com/STMicroelectronics/stm32f7xx-hal-driver/commit/2770385dd04c97d49cf285e9a5c9d1a52edbdf25).

I removed warnings and asserts related to said target (F7), but kept enforcing them for F4.

In the network driver  traceISR_ENTER() is called but isn't defined anywhere, not in TCP or in FreeRTOS.
I have removed it, but defining an empty macro in the appropriate file would also work.

As it wasn't respecting the `iptraceXXX` I sepcifically didn't add it to the ip default macros. 

Test Steps
-----------
Compiling for STM32F7 

Checklist:
----------
- [ X] I have tested my changes. No regression in existing tests.
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1279, opened 2025-07-30
STM32: Fetch Latest Drivers

HTRamsey:
Description
-----------

- Fetch latest Eth drivers rather than providing a static set.
- Allows easy updating in the future, and avoids spelling/formatting CI checks.
- Also provides some minor fixes/updates due to updating the Eth drivers.
- Replaces #1253

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


HTRamsey:
Either way. I used FetchContent because this repo was already using it elsewhere to get the kernel

HTRamsey:
@tony-josi-aws I would highly recommend at least doing the fix for [arp](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1279/files#diff-2f786125eb9c288911c4d929cf252b861bd611c789c6a8580df88d2c1a1e85c4R1722), I can make a separate PR that doesn't mess with the FetchContent stuff to get that fix in if you'd prefer

HTRamsey:
@htibosch it happens during cmake configure time when FREERTOS_PLUS_TCP_NETWORK_IF_LIST is STM32

PR #1283, opened 2025-08-06
Zynq synchronize transmission

htibosch:
Zynq: wait for transmission-ready before sending a new packet
-----------

In his post [Long delay between two splitted frames](https://forums.freertos.org/t/freertos-tcp-long-delay-between-two-splitted-frames/23637), [Tam](https://forums.freertos.org/u/trantam_cdt92) wrote that he sees many retransmissions when sending from a Zynq7000 to a local device.

I found that in the driver, after a packet is sent, a variable `int txBusy` is set and when sending a new packet, the variable is cleared again. The variable should be use to synchronise the two events: TX-start and TX-end.

Now when the system is fast, the GMAC gets overrun and setting the start-bit `STARTTX` in the register `NWCTRL` is ignored. As soon as any new packet is sent, the earlier one is flushed as well. This cause many annoying re-transmissions and timeouts on the remote side.

Solution:
~~~c
SemaphoreHandle_t txMutex;
~~~

This mutex will be given to by the `emacps_send_handler()`. It is called from an ISR so it does;
~~~d
    /* Take the mutex without blocking. */
    xSemaphoreGiveFromISR( xemacpsif->txMutex, &xHigherPriorityTaskWoken );
~~~

Tam, thank you for your valuable input, and your help while we were testing.

mentioned post in the FreeRTOS forum explains it all.

Checklist:
----------

Related Issue
-----------

There is no issue ID, the case was presented on the FreeRTOS forum.

### DIFF
--- a/source/portable/NetworkInterface/Zynq/x_emacpsif.h
+++ b/source/portable/NetworkInterface/Zynq/x_emacpsif.h
@@ -60,7 +60,7 @@
 
     extern XEmacPs_Config mac_configs[ XPAR_XEMACPS_NUM_INSTANCES ];
 
-
+    *
     void xemacpsif_setmac( uint32_t index,
                            uint8_t * addr );
     uint8_t * xemacpsif_getmac( uint32_t index );
@@ -116,7 +116,7 @@
         volatile int txHead, txTail;
 
 /*      txMutex is replacing the earlier int variable 'txBusy'. */
-		SemaphoreHandle_t txMutex;  /* Replacing the earlier variable 'volatile int txBusy'. */
+        SemaphoreHandle_t txMutex; /* Replacing the earlier variable 'volatile int txBusy'. */
 
         volatile uint32_t isr_events;
 
@@ -171,7 +171,7 @@
     void vInitialiseOnIndex( BaseType_t xIndex );
 
     #ifdef __cplusplus
-}     /* extern "C" */
+    } /* extern "C" */
     #endif
 
 #endif /* __NETIF_XAXIEMACIF_H__ */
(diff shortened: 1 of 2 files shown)


source/portable/NetworkInterface/Zynq/x_emacpsif.h
```diff
@@ -60,7 +60,7 @@
 
     extern XEmacPs_Config mac_configs[ XPAR_XEMACPS_NUM_INSTANCES ];
 
-
+    *
```
moninom1, line 63: This seems to be syntax error.

moninom1, line 63:
```suggestion
```

htibosch: Corrected, thanks

htibosch:
@moninom1, thanks for approving, but we're not ready yet.
The patch works in one situation, not in another case.
I will report back here when [the issue](https://forums.freertos.org/t/freertos-tcp-long-delay-between-two-splitted-frames/23637) is solved,

htibosch:
I continued testing the setup as described [here](forums.freertos.org/t/freertos-tcp-long-delay-between-two-splitted-frames/23637).
My conclusion is that the Zynq network interface needs more protection against racing conditions. And also I suspect that the Python script was showing some weird characteristics.
In order to avoid confusions, I will create a new PR to solve this subject.

htibosch:
I created a new [PR 1286](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1286), which replaces this PR. It has the title "Zynq driver: avoid race conditions".

htibosch:
Please refer to [PR 1286](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1286) which will replace this PR.

PR #1286, opened 2025-09-01
Zynq driver: avoid race conditions

htibosch:
Description
-----------
A visitor on the FreeRTOS forum, [trantam](https://forums.freertos.org/u/trantam_cdt92) found that sometimes a TCP connection has non-active moments of 60 ms or longer. See this [long conversation](https://forums.freertos.org/t/freertos-tcp-long-delay-between-two-splitted-frames/23637/29).

I did an earlier attempt in [PR 1283](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1283), but it didn't solve all.

Changes how to handle notification value: 
--

In the current driver, the EMAC task is woken up using a generic task notify:

`vTaskNotifyGiveFromISR( xEMACTaskHandles[ xEMACIndex ], &xTaskWoken );`

and the notification value is stored in a separate variable.

In the new version, the notification value is stored in the task:

`xTaskNotifyFromISR( xEMACTaskHandles[ xEMACIndex ], EMAC_IF_RX_EVENT, eSetBits, &( xTaskWoken ) );`

This guarantees 'atomic' updates of the notification value, both when setting or clearing.

Introduced a new mutex:
--

For received packets, there is only a single path: a RX interrupt setting `EMAC_IF_RX_EVENT`, the EMAC task calls `emacps_check_rx()`, which sends the messages to a queue, owned by the IP-task.

For transmissions, there are two paths, handled by these functions:

- `emacps_send_message()` which sends packets.
- `emacps_check_tx()` which cleared used tx descriptors.

Test Steps
-----------
Please refer to [the forum post](https://forums.freertos.org/t/freertos-tcp-long-delay-between-two-splitted-frames/23637)

Checklist:
----------

Related Issue
-----------
Be careful when using Python socket for testing purposes, the behaviour may be unexpected. See forum post as well.

### DIFF
--- a/source/portable/NetworkInterface/Zynq/NetworkInterface.c
+++ b/source/portable/NetworkInterface/Zynq/NetworkInterface.c
@@ -439,7 +439,7 @@ static BaseType_t xZynqNetworkInterfaceOutput( NetworkInterface_t * pxInterface,
         {
             emacps_send_message( &( xEMACpsifs[ xEMACIndex ] ), pxBuffer, bReleaseAfterSend );
             xSemaphoreGive( xEMACpsifs[ xEMACIndex ].tx_mutex );
-            *The function emacps_send_message() has released the packet. * /
+            /* The function emacps_send_message() has released the packet. */
             bReleaseAfterSend = pdFALSE;
         }
     }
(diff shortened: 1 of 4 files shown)


kstribrnAmzn:
@htibosch - I've started the checks!

htibosch:
Thanks, the checks are now done. Is there currently anyone who could review/test this PR for Zynq-7000?
Thanks

PR #1292, opened 2025-09-28
mDNS: when a reply do not repeat the questions, set authoritative flag

htibosch:
Description
-----------

Both mDNS and LLMNR work fine, except whan it receives a request from a Linux machine: we found that an mDNS server **must** set the authoritative flag, and also that it shall not repeat the Questions from the requester.
Other OSs like Windows are more forgiving.

1. Set a new variable `xSet.uxSkipCount` as soon as the full name in Questions has been parsed. Why: because the Questions section will become the `Answers` section, containing:

   Name: 'full_name.local', Type A, Class IN, TTL, Addr length, Address.
   Or the name followed by 16 or 28 bytes.

1. When replying to an mDNS request, do not include the Questions section: 'xSet.usQuestions' will become zero.

1. With thanks to @ActoryOu : mDNS sets the authoritative flag when replying. See PR #1223  

1. Do not handle type A when IPv4 is not enabled
   Do not handle type AAAA when IPv6 is not enabled

1. Make sure that 'pxNewBuffer' will be released just before exiting the function `DNS_ParseDNSReply()`

Test Steps
-----------
I tested while enabling either `ipconfigUSE_MDNS` or `ipconfigUSE_LLMNR`. The former replies with an Answers section only, the latter repeats the questions and then sends the Answers.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/include/FreeRTOS_DNS_Globals.h
+++ b/source/include/FreeRTOS_DNS_Globals.h
@@ -110,13 +110,13 @@
     #endif /* ipconfigBYTE_ORDER == pdFREERTOS_BIG_ENDIAN */
 
 /* MDNS constants. */
-    #define ipMDNS_TIME_TO_LIVE    255U
+    #define ipMDNS_TIME_TO_LIVE          255U
     #define dnsMDNS_FLAGS_IS_RESPONSE    0x8400U /**< MDNS flag value for response. */
 
-    #define ipLLMNR_PORT           5355U  /* Standard LLMNR port. */
-    #define ipDNS_PORT             53U    /* Standard DNS port. */
-    #define ipNBNS_PORT            137U   /* NetBIOS Name Service. */
-    #define ipNBDGM_PORT           138U   /* Datagram Service, not included. */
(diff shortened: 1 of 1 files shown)


moninom1:
@htibosch Seems like UTs are also failing, can you please help in checking on that. Thanks

htibosch:
Please find attached some code that I use for `xApplicationDNSQueryHook_Multi()`:

[dns_query_hook.zip](https://github.com/user-attachments/files/22921578/dns_query_hook.zip)

It will check the IP-type that is desired by the requester. Maybe that will solve the problem of the empty IPv6 address.

tony-josi-aws (approved):

htibosch:
Hi, is there anyone who can look at the failing [CI checks](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/actions/runs/18520137186/job/52894721551?pr=1292) please?
I only made changes to the UT to get 100% coverage. Could that be related?

AniruddhaKanhere (approved):

PR #1300, opened 2025-11-10
Add support for EUI64 IPv6 addresses

### LINKED ISSUE #1150: [Feature Request] Support EUI-64 or similar for global unicast IPv6 addresses derived from RA
**Is your feature request related to a problem? Please describe.**
When router advertisements are used as a means of IPv6 address configuration, the IP stack generates a random host id, appends it to the prefix, and sends a neighbor solicitation to rule out a collision: https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/FreeRTOS_RA.c#L579 This works fine if you assume that the device using the stack will be a client. However, when the device needs to be a server, a stable address is needed. Normally, EUI-64 or a similar method could be used for generation of such address.

**Describe the solution you'd like**
Please add a way to derive stable addresses from router advertisements.

**Describe alternatives you've considered**
I am forced to assign my addresses statically.

ljanyst:
Add support for EUI64 IPv6 addresses

Description
-----------
Needed to run as a server.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------
Fixes #1150 

### DIFF
--- a/source/FreeRTOS_DNS_Parser.c
+++ b/source/FreeRTOS_DNS_Parser.c
@@ -263,6 +263,7 @@
         BaseType_t xReturn = pdTRUE;
         uint32_t ulIPAddress = 0U;
         BaseType_t xDNSHookReturn = 0U;
+        NetworkBufferDescriptor_t * pxNewBuffer = NULL;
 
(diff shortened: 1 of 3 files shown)


ljanyst:
Any actions left on my side?

PR #1309, opened 2026-02-08
stm32 network interface overhaul

HTRamsey:
Few fixes, update HAL drivers, implement rest of HAL


HTRamsey:
@htibosch Should I make a separate PR with critical fixes first and then later come back at some point in another PR and fill in missing HAL features?

PR #1310, opened 2026-02-08
Add Generic CMSIS Network Interface Template

HTRamsey:
Provides a Network Interface for every platform with a CMSIS implementation for the ETH peripheral


HTRamsey:
@htibosch This should be really convenient for getting most platforms to work at least to a minimal functional state

Issue #1284, opened 2025-08-06
[Feature Request] mDNS Service Discovery


george-palmsens:
I want to be able to do automatic service discovery using DNS-SD over mDNS.

---

I would like to be able to use a more advanced hook than `xApplicationDNSQueryHook` to be able to describe a more comprehensive set of DNS queries to provide responses to. Then DNS-SD could be supported on top of that.

---

For example:
```c
/* Records necessary for DNS-SD */
const DNSRecordTable_t records[] = {
    {dnsTYPE_PTR, .name ="service._udp.local", .domain="myname.service._udp.local"},
    {dnsTYPE_SRV, .name ="myname.service._udp.local", .port = 5353, target="myname.local"},
    {dnsTYPE_HOST_A, .name = "myname.local"}
};

BaseType_t xApplicationDNSQueryHook(DNSRecordTable_t const ** out_records, size_t* out_records_len) {
    *out_records = records
    *out_records_len = LEN(records);
    return pdTRUE;
}
```

I am also happy to help deliver this feature.

This would serve LLMNR through the same mechanism.

moninom1:
Thank you for raising this feature request. I’ll review it with the team and follow up soon.

george-palmsens:
> [@george-palmsens](https://github.com/george-palmsens) We have a multicast support PR that is in progress - [#1019](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1019) Since it's already in progress, I'm wondering if this feature can be implemented better with multicast support.

That would certainly work - My proposition was mainly to keep the delta as small as possible, for the sake of implementation speed.

george-palmsens:
> [@george-palmsens](https://github.com/george-palmsens), you proposal sounds good!
> 
> [@evpopov](https://github.com/evpopov), what is your feeling about the above proposals? Any comments, recommendations?
> 
>   // Current signature
>   BaseType_t xApplicationDNSQueryHook_Multi( NetworkEndPoint_t *pxEndPoint, const char *pcName )
>   // Desired signature
>   BaseType_t xApplicationDNSQueryHook( const DNSRecord_t ** outRecords, size_t * outLen )
> 
> We used to have a application hook called `xApplicationDNSQueryHook()`, which is now renamed to `xApplicationDNSQueryHook_Multi()`, provided that `ipconfigIPv4_BACKWARD_COMPATIBLE` is not enabled.
> 
[...]

PR #1326, opened 2026-03-11
Overhaul mDNS to allow arbitrary records (supports DNS Service Discovery)

george-palmsens:
This changes the DNS hook to return a list of records to serve over mDNS. By setting an appropriate set of records, this satisfies DNS-SD (Avahi, Bonjour, etc).

An "appropriate set of records" looks like:

```c
static DNSRecord_t s_records[] = {
    {.usRecordType = dnsTYPE_PTR,
     .pcName       = "_services._dns-sd._udp.local",
     .xData.pcPtrRecord  = "_myservice._udp.local"},

    {.usRecordType = dnsTYPE_PTR,
     .pcName       = "_myservice._udp.local",
     .xData.pcPtrRecord  = "myhostname._myservice._udp.local"},

    {.usRecordType = dnsTYPE_SRV,
     .pcName       = "myhostname._myservice._udp.local",
     .xData.xSrvRecord   = {.pcTarget = "myhostname.local", .usPort = SERVICE_PORT}},

    {.usRecordType = dnsTYPE_TXT,
     .pcName = "myhostname._myservice._udp.local",
     .xData.pcTxtRecord = "Arbitrary service text"},
     
    {.usRecordType = dnsTYPE_A_HOST,
     .pcName = "myhostname.local"},
};
```

Other things that have been changed to accomodate this:

- DNS responses now check all questions, not just the first one
- DNS responses may contain many answers

Checklist:
----------

I have tested my changes locally, and verified the device is discoverable through Avahi. I have not run an application level test suite.

I have added a shim that allows the old hooks to still work via a back-compatibility flag, I hope that is a fine approach.

I have added the necessary unit tests that reach 100% coverage. A few isolated cases had to be ignored from the coverage report since they are defensive checks that are already caught earlier in the logic.

Requests for assistance:
----------
- How do I document this feature?

Related Issue
-----------
See https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/1284

### DIFF
--- a/source/FreeRTOS_DNS_Parser.c
+++ b/source/FreeRTOS_DNS_Parser.c
@@ -457,8 +457,85 @@
                 }
             }
         }
-    #endif /* if ( ( ipconfigUSE_LLMNR != 0 ) || ( ipconfigUSE_MDNS != 0 ) ) */
 
+/**
+ * @brief Get the network buffer associated with the given parse set.
+ * Will calculate it if necessary, otherwise just return the previously calculated value.
+ *
+ * @param[in,out] xSet A set of variables that are shared among the helper functions.
+ * @return A pointer to the network buffer descriptor.
+ */
+        static NetworkBufferDescriptor_t * DNS_GetNetworkBuffer( ParseSet_t * xSet )
+        {
+            if( xSet->pxNetworkBuffer == NULL )
+            {
+                xSet->pxNetworkBuffer = pxUDPPayloadBuffer_to_NetworkBuffer( xSet->pucUDPPayloadBuffer );
+            }
+
+            return xSet->pxNetworkBuffer;
+        }
+
+/**
+ * @brief Get the endpoint (by copy) associated with the given parse set.
+ *
+ * @param[in,out] xSet A set of variables that are shared among the helper functions.
+ * @param[out] pxEndPointOut A pointer to the endpoint structure where the result will be stored.
+ * @return pdTRUE if the endpoint was successfully retrieved, pdFALSE otherwise.
+ */
+        static BaseType_t DNS_GetEndpoint( ParseSet_t * xSet,
+                                           NetworkEndPoint_t * pxEndPointOut )
+        {
+            NetworkBufferDescriptor_t * pxNetworkBuffer = DNS_GetNetworkBuffer( xSet );
+
+            if( ( pxNetworkBuffer == NULL ) || ( pxNetworkBuffer->pxEndPoint == NULL ) )
+            {
+                return pdFALSE;
+            }
+
+            memcpy( pxEndPointOut, pxNetworkBuffer->pxEndPoint, sizeof( NetworkEndPoint_t ) );
+            return pdTRUE;
+        }
+
+/**
+ * @brief Get the DNS records to be used for answering the questions in xSet.
+ * Will call the application hook if necessary, or return a cached answer if not.
+ * @param[in,out] xSet A set of variables that are shared among the helper functions.
+ * @return pdTRUE if the records were successfully retrieved, pdFALSE otherwise.
+ */
+        static BaseType_t DNS_GetRecords( ParseSet_t * xSet )
+        {
+            UBaseType_t x;
+
+            #if ( ipconfigDNSQuery_BACKWARD_COMPATIBLE == 1 )
+
+                /* No need to do anything. We aren't *really* fetching records in the back-compatible
+                 * path, we're instead creating them on the way*/
+            #else
+                NetworkEndPoint_t xEndPoint;
+
+                if( xSet->pxDNSRecords == NULL )
+                {
+                    #if ( ( ipconfigIPv4_BACKWARD_COMPATIBLE == 1 ) )
+                        ( void ) xEndPoint;
+                        xSet->pxDNSRecords = xApplicationDNSRecordQueryHook( &xSet->uxDNSRecordCount );
+                    #else
+                        if( DNS_GetEndpoint( xSet, &xEndPoint ) == pdFALSE )
+                        {
+                            return pdFALSE;
+                        }
+                        xSet->pxDNSRecords = xApplicationDNSRecordQueryHook_Multi( &xEndPoint, &xSet->uxDNSRecordCount );
+                    #endif
+
+                    for( x = 0U; x < xSet->uxDNSRecordCount; x++ )
+                    {
+                        xSet->pxDNSRecords[ x ].uxIncludeInAnswer = pdFALSE;
+                    }
+                }
+            #endif /* if ( ipconfigDNSQuery_BACKWARD_COMPATIBLE == 1 ) */
+            return pdTRUE;
+        }
+
+    #endif /* if ( ( ipconfigUSE_LLMNR != 0 ) || ( ipconfigUSE_MDNS != 0 ) ) */
 
 
 /**
@@ -471,47 +548,23 @@
  *          - pxDNSRecords and uxDNSRecordCount
  *
  * @param[in,out] xSet a set of variables that are shared among the helper functions.
- * @param[in] pxEndPoint The endpoint on which the DNS message was received.
- *                       Necessary when LLMNR/MDNS are enabled
- *                       Otherwise may be NULL.
  * @return pdTRUE if everything went okay
  */
-    static BaseType_t parseDNSQuestions( ParseSet_t * xSet,
-                                         NetworkEndPoint_t const * pxEndPoint )
+    static BaseType_t parseDNSQuestions( ParseSet_t * xSet )
     {
         UBaseType_t x;
         size_t uxResult;
 
         #if ( ( ipconfigUSE_LLMNR == 1 ) || ( ipconfigUSE_MDNS == 1 ) )
             NetworkEndPoint_t xEndPoint;
-            configASSERT( pxEndPoint != NULL );
-
-            /* Make a copy of the end-point because xApplicationDNSQueryHook() is allowed
-             * to write into it. */
-            ( void ) memcpy( &( xEndPoint ), pxEndPoint, sizeof( xEndPoint ) );
-
-            #if ( ipconfigDNSQuery_BACKWARD_COMPATIBLE == 1 )
-                xSet->uxDNSRecordCount = 0;
-            #else
-                #if ( ( ipconfigIPv4_BACKWARD_COMPATIBLE == 1 ) )
-                    xSet->pxDNSRecords = xApplicationDNSRecordQueryHook( &xSet->uxDNSRecordCount );
-                #else
-                    xSet->pxDNSRecords = xApplicationDNSRecordQueryHook_Multi( &xEndPoint, &xSet->uxDNSRecordCount );
-                #endif
-
-                for( x = 0U; x < xSet->uxDNSRecordCount; x++ )
-                {
-                    xSet->pxDNSRecords[ x ].uxIncludeInAnswer = pdFALSE;
-                }
-            #endif /* if ( ipconfigDNSQuery_BACKWARD_COMPATIBLE == 1 ) */
-        #endif /* if ( ( ipconfigUSE_LLMNR == 1 ) || ( ipconfigUSE_MDNS == 1 ) ) */
-
-        ( void ) pxEndPoint;
+            const uint8_t * pucThisNameField;
+            BaseType_t xEndPointValid = pdFALSE;
+        #endif
 
         for( x = 0U; x < xSet->usQuestions; x++ )
         {
             #if ( ( ipconfigUSE_LLMNR == 1 ) || ( ipconfigUSE_MDNS == 1 ) )
-                const uint8_t * const pucThisNameField = xSet->pucByte;
+                pucThisNameField = xSet->pucByte;
 
                 if( x == 0U )
                 {
@@ -555,8 +608,10 @@
                     xSet->usType = usChar2u16( xSet->pucByte );
                     xSet->usClass = usChar2u16( &( xSet->pucByte[ 2 ] ) );
 
-                    /* This should have been set by now. */
-                    configASSERT( xSet->pxDNSRecords != NULL );
+                    if( DNS_GetRecords( xSet ) == pdFALSE )
+                    {
+                        return pdFALSE;
+                    }
 
                     #if ( ( ipconfigDNSQuery_BACKWARD_COMPATIBLE == 1 ) )
                     {
@@ -564,10 +619,22 @@
 
                         if( x == 0U )
                         {
+                            BaseType_t xIsMatched;
+
+                            if( xEndPointValid == pdFALSE )
+                            {
+                                if( DNS_GetEndpoint( xSet, &xEndPoint ) == pdFALSE )
+                                {
+                                    return pdFALSE;
+                                }
+
+                                xEndPointValid = pdTRUE;
+                            }
+
                             #if ( ( ipconfigIPv4_BACKWARD_COMPATIBLE == 1 ) )
-                                BaseType_t const xIsMatched = xApplicationDNSQueryHook( xSet->pcName );
+                                xIsMatched = xApplicationDNSQueryHook( xSet->pcName );
                             #else
-                                BaseType_t const xIsMatched = xApplicationDNSQueryHook_Multi( &xEndPoint, xSet->pcName );
+                                xIsMatched = xApplicationDNSQueryHook_Multi( &xEndPoint, xSet->pcName );
                             #endif
 
                             if( xIsMatched == pdTRUE )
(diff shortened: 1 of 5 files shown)


source/FreeRTOS_DNS_Parser.c
```diff
@@ -235,8 +235,518 @@
...
 
+
+    #if ( ( ipconfigUSE_LLMNR != 0 ) || ( ipconfigUSE_MDNS != 0 ) )
+
+/**
+ * @brief Check two strings for equality, up to the given length, with
+ * ascii case-insensitivity.
+ */
+        static BaseType_t prvCompareStringsCaseInsensitive( const char * s1,
+                                                            const char * s2,
```
htibosch: The library already uses [`strcasecmp()`](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/af6b379e3a580b86ac99f493918c6d18ffcdf264/source/FreeRTOS_DNS_Callback.c#L85)
So I think you can leave out the definition of `prvCompareStringsCaseInsensitive()`?

Sorry, `prvCompareStringsCaseInsensitive()` does the same as `strncasecmp()`. But still I'd prefer to use that standard function.

george-palmsens:
Added support for "Additional RRs" in the generated response, with a corresponding hook.

This is a `SHOULD` in the DNS-SD spec, and clients aren't allowed to rely on it. But they still do :shrug: and so this is necessary in order to get everything working reliably with a variety of client libraries.

htibosch:
Hi @george-palmsens , you're doing a great job. Please allow me time to review and test the changes, most importantly: keeping thing downward compatible.
Thanks so far!

david-sawatzke:
@george-palmsens Thank you for the implementation, just tested it and found it to work great!

For documentation: Some example code with the hooks & usage of `SERVE_ADDITIONAL` in the MatchedHook as well documenting expected "reasonable" config options, such as increasing the `ipconfigDNS_CACHE_NAME_LENGTH` to the mDNS-SD query length would've made it easier for me. But I'm also unsure how this is done for freertos+tcp.

Regarding the TXT records: Currently only one printable key/value pair per TXT record is supported. 

The current approach works fine for me, but the specification generally expects one TXT records for most cases (6.8.) and putting multiple key/value pairs inside the one (and recommends adding `txtvers` as the first key/value pair). Furthermore the value in the key value pairs can be arbitrary binary data (including \0), although this seems to be very uncommon in practice. As far as I can tell, Avahi is limited to one record. 

I've thought about this for a bit, but I don't know how a nice API for this would look. Perhaps something like this for the "full" approach?

```
struct xDNSTXTRecordEntry {
  char *pcKey;
  void *pvValue;    // No value if nullptr (i.e. no '=' inserted), empty value if value_length == 0. nullptr takes precendence
  uint8_t ucValueLength;  // Performs strlen on value if e.g. 0xFF, an otherwise invalid value?
}

#define dnsTXTRECORDENTRY_EMPTYVALUE 0
#define dnsTXTRECORDENTRY_PRINTABLEVALUE 0xFF

#define xDNSTXTRecordPrintableEntry(KEY, VALUE) {KEY, VALUE, dnsTXTRECORDENTRY_PRINTABLEVALUE}

// ...

// DNS Record Union

union
            {
                char * pcPtrRecord;
                struct
                {
                    const char * pcTarget;
                    uint16_t usPort;
                } xSrvRecord;
                struct 
                {
                    xDNSTXTRecordEntry *pxEntries;
                    size_t uxEntryCount;
                }
            } xData;
```

Alternatively when ignoring binary data, I see two other options:
- Keep with one string and use a magic value prefixed for each key/value pair: "a=1\001b=2\001c=3" as [laid out here](https://www.zeroconf.org/rendezvous/txtrecords.html). Would make it less convenient to update values, but builds on the trivial case currently supported.
- Uses a `{char *key, char *value}` struct array, which saves a byte & the "magic" value, as done by the [ESP-IDF](https://docs.espressif.com/projects/esp-protocols/mdns/docs/latest/en/index.html#mdns-services), misses empty value/no value differentiation.

george-palmsens:
@david-sawatzke Thanks for test-driving the code, great to hear it works!

Where should documentation be added? Just in the relevant headers, or is there an example/website docs location I should be putting it?

You're definitely right about TXT records. I'll need to get my head back into the code and look at your suggestions and I'll get back to you :+1:

george-palmsens:
@david-sawatzke So the key/value pair thing is DNS-SD specific - this PR doesn't specifically care about DNS-SD, so I'd feel uncomfortable adding them into the actual structure of the code.

So I would aim for making the text record simply contain an arbitrary number of string entries. And if you want to use those for `key=value` then you can.

As such I think I would just alter the struct to take a pointer to an array of strings, rather than a single entry. So it would look like:

```c
typedef struct xDNSRecord
{
    uint16_t usRecordType;

    /* How to serve this record.
     * See `dnsRECORD_SERVE_NO`, `dnsRECORD_SERVE_ADDITIONAL` and `dnsRECORD_SERVE_ANSWER`. */
    BaseType_t uxServeRecord;
    const char * pcName;
    union
    {
        char * pcPtrRecord;
        struct
        {
            const char * pcTarget;
            uint16_t usPort;
        } xSrvRecord;
        struct
        {
            const char * const * pcStrings;
            UBaseType_t uxNumStrings;
        } xTxtRecord;
    } xData;
} DNSRecord_t;
```

This would still stop us from serving strings with null bytes, but it sounds like we can tolerate that...?

david-sawatzke:
> Where should documentation be added? Just in the relevant headers, or is there an example/website docs location I should be putting it?

I don't know, but I'd guess it should be in the headers anyways so there first?

> This would still stop us from serving strings with null bytes, but it sounds like we can tolerate that...?

At a quick look I haven't found a mDNS-SD service type that requires it, so I think it's fine.

> So I would aim for making the text record simply contain an arbitrary number of string entries. And if you want to use those for key=value then you can.

This approach sounds good to me, a bit more annoying in the mDNS-SD usecase (as the string needs to be mutated to change the value if the value is already stored as a string), but also supports encodings outside RFC1464

george-palmsens:
> This approach sounds good to me, a bit more annoying in the mDNS-SD usecase (as the string needs to be mutated to change the value if the value is already stored as a string), but also supports encodings outside RFC1464

I haven't established how any of this works if you want to change the records being served dynamically.

I don't think the current API supports it - because at any given moment the IP task could be iterating over the records. You would need to do some sort of deep copy in the hook, which seems very impractical.

EDIT:

We could add an additional hook for when the DNS code is finished using the records. That would give an acquire/release pair that the user could work with to update the DNS records from another thread.

That would compose nicely with what we have now.

david-sawatzke:
> I don't think the current API supports it - because at any given moment the IP task could be iterating over the records. You would need to do some sort of deep copy in the hook, which seems very impractical.

Ah, you're right, I've misunderstood. I thought `xApplicationDNSRecordQueryHook` would be called every time. But I don't think a deep copy is required if it's only mutated in the callback, as that's all in the IP task afaict?

 As a hacky approach, one could also update the contents of the records in `xApplicationDNSRecordsMatchedHook`, which mutatest it anyways.

Then I've probably also misunderstood `xApplicationDNSRecordsMatchedHook` or there's a bug, as `uxServeRecord` only gets reset if `xApplicationDNSRecordQueryHook` is called, so `xApplicationDNSRecordQueryHook` needs to reset `uxServeRecord` as well if they should only be sent out for a specific answer? 

I think the best option could be to call `xApplicationDNSRecordQueryHook` on every request? in the trivial case, this is almost free, in the non-trivial case this is probably the best place to do it (otherwise the information might be quite out of date).

EDIT: as `xSet` is always cleared on every `ParseDNSReply`, so `xApplicationsDNSRecordQueryHook` already gets called on every request and I works how I thought it should work before?

george-palmsens:
> Ah, you're right, I've misunderstood. I thought `xApplicationDNSRecordQueryHook` would be called every time. But I don't think a deep copy is required if it's only mutated in the callback, as that's all in the IP task afaict?

Yeah it is called every request - but then the IP task iterates through the returned records.

And yeah if you do in-place mutation in the callback then you will be fine :+1: . That is a workable solution.

source/FreeRTOS_DNS_Parser.c
```diff
@@ -174,69 +174,729 @@
...
+            UBaseType_t x;
+
+            #if ( ipconfigDNSQuery_BACKWARD_COMPATIBLE == 1 )
+
+                /* No need to do anything. We aren't *really* fetching records in the back-compatible
+                 * path, we're instead creating them on the way*/
+            #else
+                NetworkEndPoint_t xEndPoint;
+
+                if( xSet->pxDNSRecords == NULL )
```
david-sawatzke, line 465: Will `xSet->pxDNSRecords` in `DNS_GetRecords` ever not be NULL? Even if yes, should it just be overwritten anyways?

george-palmsens: It is not NULL in the case where there are multiple questions. `GetRecords` is called in the question loop.

george-palmsens:
@david-sawatzke Multiple strings are now present, and I've also generally tidied and improved the implementation, and added example code.

```
❯ avahi-browse _myservice._udp -rt                                                                                                                    FreeRTOS-Plus-TCP
+ enp34s0u2u4 IPv4 instance                                      _myservice._udp      local
= enp34s0u2u4 IPv4 instance                                      _myservice._udp      local
   hostname = [myhostname.local]
   address = [192.168.0.137]
   port = [83]
   txt = ["name=FreeRTOS" "example=DNS-SD"]
```

Note to reviewers, coverage is <100% in the DNS code but this is also true on main. It is not caused by this PR.

david-sawatzke:
Tested it, also works great for me. Thank you @george-palmsens for the changes & cleanup, and the example in the header!

source/FreeRTOS_DNS_Parser.c
htibosch: @george-palmsens, thanks for your PR. If I test your PR with Windows as a host, what kind of software should I install for testing the mDNS records? Or is it easier to test it using Linux?

george-palmsens: Avahi on Linux does make life a lot easier...

That said, on Windows you should have the `dns-sd` command if you install Bonjour. That should do the job too!

george-palmsens: I have rebased and used strncasecmp

george-palmsens: Just tested it on a colleague's machine - `dns-sd` works fine on Windows :)

george-palmsens:
Is there anything I can do to help push this over the line?

htibosch:
@george-palmsens wrote:

> Is there anything I can do to help push this over the line?

Yes please. Maybe your write down shortly what steps I need to take for testing.
If you want, write an email to
hein [at] htibosch [dot] net
so we can communicate quicker.
I have boards, Windows 11, and Linux (mint)

george-palmsens:
Service discovery (DNS-SD) is a system by which members of an IP network can discover the location of other devices implementing particular services.

mDNS is a system by which clients on a local network can resolve DNS requests in a peer-to-peer way, even in the absence of a DNS server.

Combined, you get mDNS-SD. With this, local network members can find services like printers, displays, audio devices and so forth without any global configuration.

---

The PR implements mDNS, allowing the device to serve a broad range of DNS records.

Strictly, nothing in the PR actually cares about DNS-SD specifically. However as a motivating example, a lot of the discussion is centred on that application.

---

Here is some example code to add to a normal project:

```c

static char const* const TxtRecordStrings[] = {
    "name=FreeRTOS",
    "example=DNS-SD",
};

static DNSRecord_t DnsRecords[] = {
    {.usRecordType      = dnsTYPE_PTR,
     .pcName            = "_services._dns-sd._udp.local",
     .xData.pcPtrRecord = "_myservice._udp.local"},

    {.usRecordType      = dnsTYPE_PTR,
     .pcName            = "_myservice._udp.local",
     .xData.pcPtrRecord = "instance._myservice._udp.local"},

    {.usRecordType     = dnsTYPE_SRV,
     .pcName           = "instance._myservice._udp.local",
     .xData.xSrvRecord = {.pcTarget = "myhostname.local", .usPort = 123}},

    {.usRecordType = dnsTYPE_TXT,
     .pcName       = "instance._myservice._udp.local",
     .xData.xTxtRecord =
         {
             .ppcStrings    = TxtRecordStrings,
             .uxStringCount = 2,
         }},

    {.usRecordType = dnsTYPE_A_HOST, .pcName = "myhostname.local"},

};

DNSRecord_t* xApplicationDNSRecordQueryHook(UBaseType_t* outLen) {
    *outLen = sizeof(DnsRecords) / sizeof(DnsRecords[0]);
    return DnsRecords;
}

extern void xApplicationDNSRecordsMatchedHook(void) {
    // Serve additional records according to RFC6763, section 12
    // Instance Enumeration implies all other records
    if (DnsRecords[1].uxServeRecord == dnsRECORD_SERVE_ANSWER) {
        DnsRecords[2].uxServeRecord = dnsRECORD_SERVE_ADDITIONAL;
        DnsRecords[3].uxServeRecord = dnsRECORD_SERVE_ADDITIONAL;
        DnsRecords[4].uxServeRecord = dnsRECORD_SERVE_ADDITIONAL;
    }
    // SRV record implies A record
    if (DnsRecords[2].uxServeRecord == dnsRECORD_SERVE_ANSWER) {
        DnsRecords[4].uxServeRecord = dnsRECORD_SERVE_ADDITIONAL;
    }
}
```

This is the IPv4 back compatible version  - the hooks are slightly different in IPv6.

To explain what these do:

- xApplicationDNSRecordQueryHook
  - Tells the stack where the records to serve are
- DnsRecords[]
  - Record 0:
    - Question: What services are available on this network?
    - Answer: `_myservice._udp.local` is available
  - Record 1:
    - Question: What instances are there of `_myservice._udp.local`?
    - Answer: There is an instance available at `instance._myservice._udp.local`
  - Record 2:
    - Question: Where can I find `instance._myservice._udp.local`?
    - Answer: You can find it at `myhostname.local:123`
  - Record 3:
    - Question: What additional info is there about `instance._myservice._udp.local`?
    - Answer: Here are some strings with additional info...
  - Record 4:
    - Question: Where can I find `myhostname.local`?
    - Answer: Here's my IP
- xApplicationDNSRecordsMatchedHook
  - This callback allows the application to customise which records are served when
  - DNS-SD recommends that Records 2, 3 and 4 should be served whenever you serve Record 1, for example


---

Once your project has this code running, you can test it!

`ping myhostname.local` should find your IPv4 address.

On Linux:
- `avahi-browse -rt _myservice._udp` will show you everything the device is serving

On Windows:
  - `dns-sd -B _myservice._udp` fill find that there is an instance named `instance`
  - `dns-sd -L instance _myservice._udp` will resolve the hostname, port and additional text data

htibosch:
Hello @george-palmsens, thank you for sharing your work on DNS-SD with the community. 

I had posted an earlier reaction here above:

> most importantly: keeping thing downward compatible

Your FreeRTOS_DNS_Parser.c has 102 difference compared to the current version.

I am missing a macro that allows the new code to become active.
~~~
    /* By default, disable the DNS Service Discovery. */
    #define ipconfigUSE_DNS_SD    ipconfigDISABLE
~~~
When a FreeRTOS user updates FreeRTOS+TCP, we don't want new features to be included automatically, like my android handphone always does.

As it is now, it is hard to review. The existing code (DNS-client/mDNS/LLMNR) is old and reliable, it has been tested and reviewed many times. We must be very careful to keep the code safe and correct.

Can't you (temporarily) isolate the new code in a new source+header file?

One minor thing: every function should have 1 point of exit. If that is difficult, you can use a do {} while ( 0 ) loop so you can use a break.
~~~
BaseType_t foo()
{
    BaseType_t rc = 0;
    do
    {
        rc = bind();
        if( rc < 0 )
        {
            break;
        }
    } while( 0 )
    return rc;
}
~~~
If possible, I would like the see the previous code as it was and see how it connects to the SD code.

george-palmsens:
Hi @htibosch .

Okay. I wrote it this way because the new system is powerful enough to do everything the old system used to do.

But I understand your hesitation to ship a large chunk of new code here.

I will refactor the PR to better match your expectations.

---

To be clear about my intentions so I don't waste work, I will make it so that:

- There is a new configuration flag `ipconfigUSE_DNS_RECORDS` which enables all the new functionality
  - I don't want to call it `USE_DNS_SD`, because this can be used for more than `DNS_SD`
- If that flag is _not_ set, then all the existing code remains the same.
- If the flag is set, it enables new application callbacks, and uses the new code.
  - New hooks are:
    - `xApplicationNBNSQueryHook` (replaces `xApplicationDNSQueryHook`)
    - `xApplicationDNSRecordQueryHook`
    - And the `_Multi` variants when in IPv6
- New code will be (as far as possible) contained in `FreeRTOS_DNS_Records.c/h` or something like that

PR #1328, opened 2026-03-17
STM32H legacy driver fix: request usable RX size in legacy zero-copy

liteStefanoUgolini:
Description
-----------
The STM32H legacy zero-copy RX path still requested ETH_RX_BUF_SIZE for replacement RX buffers even though BufferAllocation_1 now limits allocations to the interface-reported usable size.

For this driver the usable size is ETH_RX_BUF_SIZE - ipBUFFER_PADDING. Requesting the full hardware buffer size caused every replacement allocation to fail, left pxReceivedBuffer as NULL, and dropped all received frames before they reached the IP task.

Request ETH_RX_BUF_SIZE - ipBUFFER_PADDING in the RX fast path so the receive path matches the enforced allocation limit.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/portable/NetworkInterface/STM32/Legacy/STM32Hxx/NetworkInterface.c
+++ b/source/portable/NetworkInterface/STM32/Legacy/STM32Hxx/NetworkInterface.c
@@ -780,8 +780,9 @@ static BaseType_t prvNetworkInterfaceInput( void )
 
         #if ( ipconfigZERO_COPY_RX_DRIVER != 0 )
         {
-            /* Reserve the maximum length for the next reception. */
-            uxLength = ETH_RX_BUF_SIZE;
+            /* Request the usable Ethernet payload area, excluding the descriptor
+             * metadata stored in front of pucEthernetBuffer. */
+            uxLength = ETH_RX_BUF_SIZE - ipBUFFER_PADDING;
 
             if( data_buffer.buffer != NULL )
             {


source/portable/NetworkInterface/STM32/Legacy/STM32Hxx/NetworkInterface.c
```diff
@@ -780,8 +780,9 @@ static BaseType_t prvNetworkInterfaceInput( void )
 
         #if ( ipconfigZERO_COPY_RX_DRIVER != 0 )
         {
-            /* Reserve the maximum length for the next reception. */
-            uxLength = ETH_RX_BUF_SIZE;
+            /* Request the usable Ethernet payload area, excluding the descriptor
+             * metadata stored in front of pucEthernetBuffer. */
+            uxLength = ETH_RX_BUF_SIZE - ipBUFFER_PADDING;
```
htibosch, line 785: That sounds like a good change, thank you.

The sizeof of the RX buffer is usually set at :

    #define ETH_RX_BUF_SIZE     1536U

because it has a great alignment ( `0x600` ).

@AniruddhaKanhere @kstribrnAmzn Can you please check this PR?

htibosch (approved):
Thank you @liteStefanoUgolini for this PR. I approve it.

cookpate (approved):

kstribrnAmzn (approved):

source/portable/NetworkInterface/STM32/Legacy/STM32Hxx/NetworkInterface.c
kstribrnAmzn: Of course!

htibosch:
We have two approvals, I like the PR, let's go for it.
Hein

liteStefanoUgolini:
@htibosch Merge is blocked by the build checks. Is there something I should do?

htibosch:
@liteStefanoUgolini As you are using STM32H, have you tried the latest version of the network interface? i do and I am very satisfied with the driver on my STM32H755

See also [this comment](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1253#issuecomment-5043339165).

htibosch:
The PR is OK, expect for this piece of code:

~~~diff
     if( ( ulRxDesc & ETH_IP_PAYLOAD_MASK ) == ETH_IP_PAYLOAD_UNKNOWN )
     {
-        iptraceETHERNET_RX_EVENT_LOST();
-        break;
     }
~~~
otherwise receiving ARP will not work.

liteStefanoUgolini:
> The PR is OK, expect for this piece of code:
> 
> ```diff
>      if( ( ulRxDesc & ETH_IP_PAYLOAD_MASK ) == ETH_IP_PAYLOAD_UNKNOWN )
>      {
> -        iptraceETHERNET_RX_EVENT_LOST();
> -        break;
>      }
> ```
> 
> otherwise receiving ARP will not work.

@htibosch it is not clear to me why this should be added in my PR, since the change will affect NetworkInterface.c
Maybe it is better you made a commit, and then I can cherry pick it if not merged already.

liteStefanoUgolini:
> @liteStefanoUgolini As you are using STM32H, have you tried the latest version of the network interface? i do and I am very satisfied with the driver on my STM32H755
> 
> See also [this comment](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1253#issuecomment-5043339165).

Yes I briefly tried it some months ago, but because we would need to update also the STM HAL library to use it, we haven't use it in our project yet. Eventually we would move to the unified STM driver.

htibosch:
@liteStefanoUgolini we're still waiting for approval for you PR.

Does the legacy driver still have advantages that the new STM32 drivers don't have?

I recommend using the latest release of the [STM32 driver](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/STM32/NetworkInterface.c).

Moving from "legacy" to the current driver is quite easy and straight forward.

The only thing is that by default it drops lots of packets. So I started by bypassing these filters and then activate them one by one.

**EDIT** : please also have a look at [PR #1366](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1366).

PR #1329, opened 2026-03-27
Fix: Cap DNS answer loop iterations to prevent DoS

felicityzhao9:
Description
-----------
In the current implementation, the loop iterates up to pxSet->usAnswers, which is derived directly from the packet header. The early-exit condition depends on usNumARecordsStored reaching ipconfigDNS_CACHE_ADDRESSES_PER_ENTRY, but this counter is only incremented when pxSet->xDoStore != pdFALSE. As a result, for unexpected replies, the counter never increases and the loop may continue iterating over all advertised answers and potentially perform excessive work.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


htibosch:
@felicityzhao9 , thank you for opening this PR.
I'm also working on it. I will test the proposed changes and report tomorrow, Friday.

htibosch:
@felicityzhao9 , it looks like the issue is already solved in this [commit](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/commit/2a165257ae5b559e0ca307c9e9b466e47896c1dd).
I think that this PR can be closed.

PR #1340, opened 2026-05-08
Add a hook function called on buffer release

fxrb:
Description
-----------
This patch adds a hook function to FreeRTOS-Plus-TCP to notify the network driver whenever a buffer is released.
One could use iptraceNETWORK_BUFFER_RELEASED (what I did at first) but this is not what it was designed for.
A network driver allocates a buffer to store incoming frames from the network. Once a frame has been received the buffer is passed to FreeRTOS-Plus-TCP which will consume it and then probably release it or not. For a zero-copy network driver using BufferAllocation_1.c it is helpful or even necessary to get a notification once FreeRTOS-Plus-TCP releases a buffer. The notification can be used for memory management, for instance to re-allocate a new buffer to receive further frames.

Checklist
----------

### DIFF
--- a/source/portable/BufferManagement/BufferAllocation_1.c
+++ b/source/portable/BufferManagement/BufferAllocation_1.c
@@ -416,6 +416,10 @@ void vReleaseNetworkBufferAndDescriptor( NetworkBufferDescriptor_t * const pxNet
             prvShowWarnings();
         }
 
+        #if ( ipconfigBUFFER_AND_DESCRIPTOR_RELEASE_HOOK == 1 )
+            vReleaseNetworkBufferAndDescriptorHook( pxNetworkBuffer );
+        #endif /* if ( ipconfigBUFFER_AND_DESCRIPTOR_RELEASE_HOOK ) */
+
         iptraceNETWORK_BUFFER_RELEASED( pxNetworkBuffer );
     }
 }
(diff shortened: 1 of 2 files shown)


source/portable/BufferManagement/BufferAllocation_1.c
```diff
@@ -416,6 +416,10 @@ void vReleaseNetworkBufferAndDescriptor( NetworkBufferDescriptor_t * const pxNet
             prvShowWarnings();
         }
 
+        #if ( ipconfigBUFFER_AND_DESCRIPTOR_RELEASE_HOOK == 1 )
+            vReleaseNetworkBufferAndDescriptorHook( pxNetworkBuffer );
```
htibosch, line 420: It would be desirable to make the same change in BufferAllocation_2.c as well to keep the interfaces equal.

htibosch (commented):
Hi @fxrb, thanks for this proposal.
Can you explain why the existing macro (`ipconfigBUFFER_AND_DESCRIPTOR_RELEASE_HOOK`) can not be used your purpose? It doesn't have any particular purpose.
If we add an application hook, I like to see a short warning: the hook will run in the IP-task context and shall not call user APIs or use blocking functions.
I like to hear more about your plans to manage the network buffers, interesting.

fxrb:
> Can you explain why the existing macro (`ipconfigBUFFER_AND_DESCRIPTOR_RELEASE_HOOK`) can not be used your purpose?

There is no such  _existing_  macro. My patch introduces it.

PR #1350, opened 2026-07-01
Fix PIC32 MAC initialization call

### LINKED ISSUE #1252: [BUG] Build breaks due to incorrect StartInitMac Usage in xPIC32_Eth_NetworkInterfaceInitialise
**Describe the bug**
In the [xPIC32_Eth_NetworkInterfaceInitialise](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/pic32mzef/NetworkInterface_eth.c#L262) function, [StartInitMac](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/pic32mzef/NetworkInterface_eth.c#L269) is used without any arguments, but it requires a [NetworkInterface_t *](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/main/source/portable/NetworkInterface/pic32mzef/NetworkInterface_eth.c#L372) parameter. The absence of this argument causes the build to fail.

**Target**
- PIC32MZ
- Instruction Set Architecture: MIPS
- IDE and version: MPLAB X
- Toolchain and version: XC32 v2.4

Old-Ding:
Description
-----------
Closes #1252.

Pass the current `NetworkInterface_t *` into `StartInitMac()` from the
PIC32MZ network interface initialisation path. The helper is declared and
defined with that parameter, and it uses the interface when iterating the
configured endpoints.

Test Steps
-----------
- `git diff --check upstream/main...HEAD`
- Verified the changed file remains UTF-8 without BOM and CRLF-only.
- Verified `StartInitMac` has one call site and it now passes `pxInterface`.

Unit tests were not run locally because this Windows environment does not
have the repo test dependencies installed (`cmake`, `ninja`, `ruby`, `make`),
and this PIC32 target file depends on the target/vendor build environment.

Checklist:
----------

Related Issue
-----------
#1252

### DIFF
--- a/source/portable/NetworkInterface/pic32mzef/NetworkInterface_eth.c
+++ b/source/portable/NetworkInterface/pic32mzef/NetworkInterface_eth.c
@@ -266,7 +266,7 @@
         if( xMacInitStatus == eMACInit )
         {
             /* This is the first time this function is called. */
-            if( StartInitMac() != false )
+            if( StartInitMac( pxInterface ) != false )
             {
                 /* Indicate that the MAC initialisation succeeded. */
                 xMacInitStatus = eMACPass;


Old-Ding:
Follow-up verification:

- This is intentionally limited to the missing `NetworkInterface_t *` argument reported in #1252.
- `StartInitMac()` has a single call site in this driver, and the callee uses the interface while iterating endpoints.
- `git diff --check upstream/main...HEAD` passes on the branch.

I do not have the PIC32MZ/MPLAB XC32 target environment locally, so this remains a source-level build-fix patch rather than a target-tested driver validation.

AniruddhaKanhere (approved):
Thank you for that change @Old-Ding.

As you mentioned, this only fixes the build but the pic32 is still NOT tested with the multi-interface stuff.

That being said, the change is good. Approved.

rawalexe (approved):

PR #1302, opened 2025-11-21
Fix incorrect state transition

Fr3ya:
When a TCP connection in eCONNECT_SYN state times out after exceeding the maximum retry count, the socket was incorrectly transitioning to eCLOSE_WAIT state instead of eCLOSED.

According to the standard TCP state machine (RFC 793):
- eCONNECT_SYN is the state where a client has sent a SYN and is waiting for a SYN-ACK
- When connection establishment fails (timeout or RST), the socket should transition to eCLOSED state

This fix:
1. Changes the state transition in prvTCPSendPacket() from `eCLOSE_WAIT` to `eCLOSED` when SYN retries are exhausted
2. Updates vTCPStateChange() to handle the `eCONNECT_SYN -> eCLOSED` transition, ensuring connection state change callbacks are properly triggered.

This also aligns the timeout behavior with the existing RST handling code, which correctly uses `eCLOSED` state (see FreeRTOS_TCP_IP.c:859).

Files changed:
- source/FreeRTOS_TCP_Transmission.c: Fix state transition
- source/FreeRTOS_TCP_IP.c: Update vTCPStateChange to handle eCLOSED transition from connecting states

Description
-----------

Test Steps
-----------

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [ ] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1355, opened 2026-07-10
Fix eCONNECT_SYN timeout incorrectly transitioning to eCLOSE_WAIT (#1301)

### LINKED ISSUE #1301: [BUG] State Transition from eCONNECT_SYN to eCLOSE_WAIT?
**Describe the bug**
The TCP state machine incorrectly transitions from `eCONNECT_SYN` (SYN-SENT) to `eCLOSE_WAIT` in error scenarios. Per RFC 793, connections that fail during establishment should transition to `eCLOSED`, not `eCLOSE_WAIT`.

https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/blob/22c095e9c3bff7773085dd9b061a0013200eadda/source/FreeRTOS_TCP_Transmission.c#L115-L139

**RFC 793 Compliance Issue**

RFC 793 specifies that SYN-SENT state errors/timeouts should transition to **CLOSED**, not CLOSE-WAIT.

94xhn:
Fixes #1301.

## Problem

Per RFC 793, a connection that fails during establishment (SYN-SENT state) must transition to `CLOSED`, not `CLOSE-WAIT`. `CLOSE-WAIT` is only valid for a connection that was previously `ESTABLISHED` and is now waiting for the local application to close it after the remote peer initiated a close.

`prvTCPSendPacket()` in `FreeRTOS_TCP_Transmission.c` violated this: when a socket in `eCONNECT_SYN` gives up after 3 SYN retries with no response, it called `vTCPStateChange(pxSocket, eCLOSE_WAIT)` instead of `eCLOSED`.

This isn't just a naming/spec-compliance issue - it has a concrete functional consequence. `vTCPStateChange()`'s "connecting phase failed" branch in `FreeRTOS_TCP_IP.c` only matched `eCLOSE_WAIT` as the target state:

```c
if( ( ( xPreviousState == eCONNECT_SYN ) || ... ) &&
    ( eTCPState == eCLOSE_WAIT ) )
{
    /* forces bBefore = pdTRUE so the state-change-notification logic below runs */
}
```

Since the SYN-retry-exhausted path already used `eCLOSE_WAIT` (matching this branch), the immediate symptom is masked for that one call site. However, this same "connecting phase failed" branch is what must also fire for the *other* legitimate way a connecting socket gets closed - e.g. receiving an RST while in `eCONNECT_SYN`, which correctly calls `vTCPStateChange(pxSocket, eCLOSED)` elsewhere (see `xProcessReceivedTCPPacket()`'s RST handling). Because `eCLOSED` was never part of this branch's condition, that path does not set `xEventBits |= eSOCKET_CLOSED` (the `bAfter == pdFALSE` branch a few lines below).

`FreeRTOS_connect()` (`FreeRTOS_Sockets.c`) waits specifically on `eSOCKET_CONNECT | eSOCKET_CLOSED` via `xEventGroupWaitBits()` to detect a failed connection attempt promptly:

```c
uxEvents = xEventGroupWaitBits( pxSocket->xEventGroup,
                                eSOCKET_CONNECT | eSOCKET_CLOSED,
                                ... , xRemainingTime );
if( ( uxEvents & eSOCKET_CLOSED ) != 0U )
{
    xResult = -pdFREERTOS_ERRNO_ENOTCONN;
    ...
}
```

Without `eSOCKET_CLOSED` ever being set for this transition, `FreeRTOS_connect()` cannot detect the failure early and instead just sits until its own `xRemainingTime` timeout expires, returning `ETIMEDOUT` instead of promptly returning `ENOTCONN`.

## Fix

Two changes, matching exactly what maintainer @htibosch proposed and tested on real hardware in the issue thread, in a prior PR #1302 that was independently approved by two maintainers but never merged (its CI was never triggered for the fork PR, and the original submitter closed it after ~2.5 months of inactivity - not due to any code problem):

1. `FreeRTOS_TCP_Transmission.c`: change the SYN-retry-exhausted call from `vTCPStateChange(pxSocket, eCLOSE_WAIT)` to `vTCPStateChange(pxSocket, eCLOSED)`, matching RFC 793.
2. `FreeRTOS_TCP_IP.c`: extend the "connecting phase failed" branch's condition to also match `eCLOSED` (not just `eCLOSE_WAIT`), so that *any* connecting-phase failure - whether via SYN-retry timeout or RST - consistently sets `eSOCKET_CLOSED` and lets `FreeRTOS_connect()` return promptly.

## Verification

I don't have a Ruby/Ceedling environment fully set up in this checkout to run the complete unit-test suite (got as far as installing Ruby/CMock, but the test build also requires the `unifdefall` tool which isn't readily available on this machine), so I verified in two ways:

1. Diffed my change against PR #1302's source changes (already reviewed and approved by two maintainers) - identical.
2. Wrote a standalone C reproduction of the exact `bBefore`/`bAfter`/`eSOCKET_CLOSED` logic from `vTCPStateChange()`, confirming:
   - `eCONNECT_SYN -> eCLOSED`: `eSOCKET_CLOSED` is NOT set before the fix, IS set after.
   - `eCONNECT_SYN -> eCLOSE_WAIT` (the pre-existing, already-working path used by e.g. keep-alive timeout from `eESTABLISHED`): behavior unchanged before/after, confirming no regression.
   - `eSYN_FIRST`/`eSYN_RECEIVED -> eCLOSED`: same fix applies symmetrically (these share the same condition).

Added `test_vTCPStateChange_ClosedState_PrvStateSyn` to `FreeRTOS_TCP_IP_utest.c`, following the existing `test_vTCPStateChange_ClosedWaitState_PrvStateSyn` as a template, to cover the newly-fixed `eCONNECT_SYN -> eCLOSED` path and assert `eSOCKET_CLOSED` is correctly set in `xEventBits`.

## Disclosure

Generative AI (Claude) was used to help investigate this issue, implement, and verify this fix. All changes were reviewed by me before submission.

### DIFF
--- a/source/FreeRTOS_TCP_IP.c
+++ b/source/FreeRTOS_TCP_IP.c
@@ -324,7 +324,8 @@
         if( ( ( xPreviousState == eCONNECT_SYN ) ||
               ( xPreviousState == eSYN_FIRST ) ||
               ( xPreviousState == eSYN_RECEIVED ) ) &&
-            ( eTCPState == eCLOSE_WAIT ) )
+            ( ( eTCPState == eCLOSE_WAIT ) ||
+              ( eTCPState == eCLOSED ) ) )
         {
             /* A socket was in the connecting phase but something
              * went wrong and it should be closed. */
--- a/source/FreeRTOS_TCP_Transmission.c
+++ b/source/FreeRTOS_TCP_Transmission.c
@@ -132,11 +132,11 @@
             {
                 /* The connection is in the SYN status. The packet will be repeated
                  * to most 3 times.  When there is no response, the socket get the
-                 * status 'eCLOSE_WAIT'. */
+                 * status 'eCLOSED'. */
                 FreeRTOS_debug_printf( ( "Connect: giving up %xip:%u\n",
                                          ( unsigned ) pxSocket->u.xTCP.xRemoteIP.ulIP_IPv4, /* IP address of remote machine. */
                                          pxSocket->u.xTCP.usRemotePort ) );                 /* Port on remote machine. */
-                vTCPStateChange( pxSocket, eCLOSE_WAIT );
+                vTCPStateChange( pxSocket, eCLOSED );
             }
             else if( prvTCPMakeSurePrepared( pxSocket ) == pdTRUE )
             {
--- a/test/unit-test/FreeRTOS_TCP_IP/FreeRTOS_TCP_IP_utest.c
+++ b/test/unit-test/FreeRTOS_TCP_IP/FreeRTOS_TCP_IP_utest.c
@@ -828,6 +828,52 @@ void test_vTCPStateChange_ClosedWaitState_PrvStateSyn( void )
 }
 
 /**
+ * @brief Test functionality when the state to be reached is eCLOSED (e.g.
+ *        SYN retries exhausted, or an RST received while connecting) and
+ *        the current state is equal to connect syn. Per RFC 793, a failed
+ *        connection attempt must transition to eCLOSED, and this must be
+ *        recognised as a "connecting phase failure" the same way the
+ *        existing eCLOSE_WAIT case is, so that eSOCKET_CLOSED gets set and
+ *        FreeRTOS_connect() can return promptly instead of waiting for its
+ *        own timeout. See https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/issues/1301
+ */
+void test_vTCPStateChange_ClosedState_PrvStateSyn( void )
+{
+    FreeRTOS_Socket_t xSocket = { 0 };
+    enum eTCP_STATE eTCPState;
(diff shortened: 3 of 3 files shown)


htibosch:
Thank you @94xhn for reminding of this issue.
At this moment I can not reaach[1302](https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1302). But I do have a copy of my proposed changes.
Mentioned PR page gives me an error. I will try again tomorrow.
Hein

94xhn:
Thanks for checking, Hein. No problem - please take your time. I?ll leave #1355 open while you verify the reference change; the relevant changes are self-contained in this PR, and #1302 is only included for comparison.

nding0405:
Hi, I independently encountered the same bug while developing the CHERIoT network stack. The thread waiting on socket connection remains blocked until timing out if the socket receives a valid RST packet from its peer. I think this is worth fixing, and I would really appreciate seeing this PR merged!

PR #1363, opened 2026-08-31
Add STM32N6 and STM32H7RS Ethernet driver support

HTRamsey:
Description
-----------

Add STM32N6 and STM32H7RS support to the common STM32 network interface and refresh the existing STM32 Ethernet HAL driver copies.

This change:

- vendors the STM32N6 and STM32H7RS Ethernet HAL and HAL extension drivers;
- adds `N6` and `H7RS` CMake driver selectors;
- supports the STM32N6 ETH1 multi-channel descriptor, interrupt, clock, and register layout;
- integrates STM32H7RS through the existing single-channel H7/H5 path;
- updates the F4, F7, H5, and H7 Ethernet HAL drivers to their current upstream versions;
- uses HAL helpers such as `HAL_ETH_GetTxBuffersNumber()` where available; and
- hardens RX/TX buffer ownership, malformed-frame handling, MAC filter reset/update behavior, descriptor alignment checks, and error recovery.

Family selection remains device-header driven; this does not add STM32 family compile definitions from CMake.

Test Steps
-----------

- Ran the repository CI formatting configuration with Uncrustify 0.69.0 on the modified C and header files.
- Ran syntax-only compilation of the common interface against the F4, F7, H5, H7, and N6 HAL/CMSIS headers.
- Ran Cortex-M7 syntax-only compilation of the common interface and both H7RS HAL sources against STM32Cube FW H7RS V1.3.0.
- Verified that the H7RS CMake selector resolves the expected HAL sources and include directory.
- Ran `git diff --check`.

Checklist:
----------

The changes are specific to the STM32 HAL integration and were validated with the vendor headers and compilers; no hardware-independent unit-test harness currently covers this interface.

Related Issue
-----------

None.


htibosch:
@HTRamsey , it looks like you've mad many people happy. Many more STM platforms can run the FreeRTOS+TCP stack.

I don't have the time or capacity to check every build setup, or every type of hardware. I'm now testing it on my STM32H755, and I will report back here.

Thank you Holden.

HTRamsey:
I used an H755 to test as well. Unfortunately I don't have an H7RS or N6 to test with but I figure adding untested support is fine as long as it doesn't affect the other existing families

HTRamsey:
working on ways to clean this up. If you could take a look at https://github.com/FreeRTOS/FreeRTOS-Plus-TCP/pull/1366 first, that would be a start to help simplify this

